PR #1, opened 2016-11-11
Initial structure

2m:
Package name is anticipating future additions, such as stream management to a separate subproject.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: scala
+
+scala:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -0,0 +1,25 @@
+language: scala
+
+sudo: required
```
2m: Yes. But thanks for catching it. Services are copy pasta from Alpakka here...

PR #7, opened 2016-12-05
Do no publish root and docs projects

2m:
Also generate 2.12 style links to scaladoc (2.11 supports 2.12 links as well, surprisingly).


2m:
So the PR validation failed here, because I have tagged the commit of this PR so I could try the release process. And while doing the PR validation travis also tried to publish which failed because of the already existing artifact in bintray. In other words, validation validated the required parts successfullly.

PR #20, opened 2017-04-13
Feature/cluster leader

agadek:
Ive added cluster leader info to members resources as I see it quite useful. I case of any objection we can substitute it with extra resource path.


raboof:
LGTM, seems useful indeed!

agadek:
I'd reale like to use this feature. Is their option to get snapshot jar from some maven jar ?

2m:
@agadek looks like you pushed some additional commits to this branch. Did you mean to do so?

agadek:
accident reverted ;-)

2m:
Cool. Thanks. :)

PR #22, opened 2017-05-09
Expose Oldest Member in GET /members

vikasuy:
See https://github.com/akka/akka-cluster-management/issues/21

From the issue
> At eero, we've found that deploying the oldest member last makes a huge difference on stability of the akka cluster during deploys. This is because oldest hosts shard coordinators and cluster singletons.

We currently expose `oldest` in our custom healthchecks, but adding this to the `akka-cluster-management` tool will help us DRY this out (and potentially help other akka cluster users)

/cc @wbertelsen @jlynn

### DIFF
--- a/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
@@ -23,7 +23,8 @@ final case class ClusterMember(node: String, nodeUid: String, status: String, ro
 final case class ClusterMembers(selfNode: String,
                                 members: Set[ClusterMember],
                                 unreachable: Seq[ClusterUnreachableMember],
-                                leader: Option[String])
+                                leader: Option[String],
+                                oldest: Option[String])
 final case class ClusterHttpManagementMessage(message: String)
 
 private[akka] sealed trait ClusterHttpManagementOperation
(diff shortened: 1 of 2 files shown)


ktoso:
Looks good, WDYT @patriknw ?

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -61,8 +62,9 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
         }
 
         val leader = cluster.readView.leader.map(_.toString)
+        val oldest = cluster.state.members.toSeq.sorted(Member.ageOrdering).headOption.map(_.address.toString)
```
patriknw: I think this should also filter to only include members with status `Up`
```
.filter(_..status == MemberStatus.Up)
```

That is what the `ClusterSingletonProxy` is doing.

ktoso: Thanks for checking, will apply that change

ktoso:
Meh, was not simple enough to just github-ui add that, I'll add the change and fix the test, then merge, then release.

ktoso:
Messed up amending this PR, sending in cleaned up one, preserves author attribute ofc.

Issue #19, opened 2017-04-02
Support for Akka-2.5.0 Cluster


mushtaq:
I tried using it with Akka-2.5.0-RC2 but started getting errors like "java.lang.NoClassDefFoundError: akka/stream/Fusing$" maybe because of binary compatibility thing. Is it intended to work with 2.5 at all?

patriknw:
Thanks for reporting. I think that can be because of the akka-http version, which used some internal APIs previously. Could you try to override the akka-http dependency to 10.0.5?

We will update the dependency in this library, of course.

mushtaq:
Overriding akka-http to 10.0.5 did the trick. Thank you!

patriknw:
Good, let's leave this ticket open until we have fixed the akka-http dependency.
Perhaps you can submit a pull request, @mushtaq ?

mushtaq:
Sure, will do this over the weekend.

stuartsaltzman:
Any update on this?

ktoso:
Just override the dependency as https://github.com/akka/akka-cluster-management/issues/19#issuecomment-291499547 mentions and it'll work. This shouldn't be a stopper for anyone.

patriknw:
I have updated the dependencies in PR #24

PR #24, opened 2017-05-10
update akka, akka-http, scala dependencies, #19

patriknw:
Refs #19

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,8 +2,8 @@ import sbt._, Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.4.14"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -2,8 +2,8 @@ import sbt._, Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.4.14"
-  val AkkaHttpVersion = "10.0.0"
+  val AkkaVersion = "2.4.18"
+  val AkkaHttpVersion = "10.0.6"
```
ktoso: hm note to self that we should update this one right away when we have important releases of http,
will add a note to our releasing doc there

Issue #14, opened 2017-03-03
Consider exposing shard details


ktoso:
Information like where which shards are allocated etc could be pretty interesting to expose here.

ktoso:
FYI @patriknw

sebastianharko:
I'm willing to implement this if it's considered useful.

patriknw:
that would be great, @sebastianharko 
Thanks

sebastianharko:
@patriknw @ktoso So I'm going to get these shard details via ShardRegion.GetClusterShardingStats  and ShardRegion.ClusterShardingStats  etc... We do not want to have akka-cluster-management depend on the sharding JAR because akka-cluster-management should be lightweight. 

So, I'm going to do this using reflection. If the akka.cluster.sharding is not on the class path, we can return something like Bad Request via the HTTP API.  

Let me know if this approach is ok or if there's another approach that does not require reflection.

patriknw:
Not sure it has to be lightweight. Doing all of it with reflection doesn't sound attractive. Perhaps you can have the dependency as `provided` and and detect at runtime if you can "load this module". Perhaps that is what you described?

sebastianharko:
@patriknw yes you're right - I totally forgot about "% provided", that's definitely the right approach.

vikasuy:
This would be really helpful. 
@patriknw what's the best way to find the shard regions actors -- will they need to be passed in explicitly to `ClusterHttpManagement`?

patriknw:
I suggest the http GET request should contain the sharding `typeName` as part of the path. That `typeName` can then be used in `ClusterSharding(system).shardRegion(typeName)` to get the region `ActorRef`. That method will throw `IllegalArgumentException` if there is no such region, so catch that and treat as 404.

vikasuy:
Thanks @patriknw. 
I've been looking into this. Building the route inside `ClusterHttpManagementRoutes` is not difficult, but testing the route is tricky. Do you have a suggestion for mocking out `ClusterSharding`. 

At a high level, i imagine this as code inside `ClusterHttpManagementRoutes` as follows:

```
private def routeGetShardInfo(cluster: Cluster, shardRegionName: String) = {
  val clusterSharding = ClusterSharding(cluster.system)
  // get shard region actor from clusterSharding and return a response
}
```

[...]

PR #29, opened 2017-05-18
Expose Shard Details

vikasuy:
Fixes https://github.com/akka/akka-management/issues/14

This PR exposes the `GET /shards/{name}` path, which returns the following data:
```
{
  regions: [
    {
      "shardId": // shard id [string]
      "numEntities": // number of entities in the shard [int]
    },
    ...
  ]
]
```

At eero, we use this information in the following ways
* monitoring: check the health of an akka cluster instance (example: did it lose shards all of a sudden?)
* deploys: after deploying new code to an akka cluster instance, we wait for it to attain shards before deploying the next instance

One change to note is that commit e7ff9c7 changes how the `prefix` works. Before, `prefix` would override the `members` path. Now, if `prefix` is set, it will be a prefix to `members` and `shard_regions`

### DIFF
--- a/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
@@ -136,7 +136,7 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
     get {
       extractExecutionContext { implicit executor =>
         complete {
-          implicit val timeout = Timeout(1.second)
+          implicit val timeout = Timeout(5.seconds)
           try {
             ClusterSharding(cluster.system)
               .shardRegion(shardRegionName)
@@ -173,7 +173,7 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
         } ~
         routesMember(cluster)
       } ~
-      pathPrefix("shard_regions" / Remaining) { shardRegionName =>
+      pathPrefix("shards" / Remaining) { shardRegionName =>
         pathEnd {
           routeGetShardInfo(cluster, shardRegionName)
         }
--- a/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
@@ -129,7 +129,7 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
    * [[akka.cluster.Cluster]] instance. This version does not provide Basic Authentication. It uses
    * the default path "members".
    */
-  def apply(cluster: Cluster): Route = apply(cluster, "members")
+  def apply(cluster: Cluster): Route = apply(cluster, "")
 
   /**
    * Creates an instance of [[akka.cluster.http.management.ClusterHttpManagementRoutes]] to manage the specified
@@ -137,11 +137,13 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
    * the specified path `pathPrefixName`.
    */
   def apply(cluster: Cluster, pathPrefixName: String): Route =
-    pathPrefix(pathPrefixName) {
-      pathEndOrSingleSlash {
-        routeGetMembers(cluster) ~ routePostMembers(cluster)
-      } ~
-      routesMember(cluster)
+    rawPathPrefix(pathPrefixName) {
+      pathPrefix("members") {
+        pathEndOrSingleSlash {
+          routeGetMembers(cluster) ~ routePostMembers(cluster)
+        } ~
+        routesMember(cluster)
+      }
     }
 
   /**
--- a/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
@@ -7,16 +7,20 @@ import java.util.concurrent.atomic.AtomicReference
 
 import akka.Done
 import akka.actor.AddressFromURIString
+import akka.cluster.sharding.{ ClusterSharding, ShardRegion }
 import akka.cluster.{ Cluster, Member, MemberStatus }
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.{ Route, RouteResult }
 import akka.http.scaladsl.{ ConnectionContext, Http }
+import akka.pattern.ask
 import akka.stream.ActorMaterializer
+import akka.util.Timeout
 import spray.json.DefaultJsonProtocol
 
 import scala.concurrent.{ Future, Promise }
+import scala.concurrent.duration._
 
 final case class ClusterUnreachableMember(node: String, observedBy: Seq[String])
 final case class ClusterMember(node: String, nodeUid: String, status: String, roles: Set[String])
@@ -26,6 +30,8 @@ final case class ClusterMembers(selfNode: String,
                                 leader: Option[String],
                                 oldest: Option[String])
 final case class ClusterHttpManagementMessage(message: String)
+final case class ShardRegionInfo(shardId: String, numEntities: Int)
+final case class ShardDetails(regions: Seq[ShardRegionInfo])
 
 private[akka] sealed trait ClusterHttpManagementOperation
 private[akka] case object Down extends ClusterHttpManagementOperation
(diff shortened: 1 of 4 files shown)


cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -124,24 +132,52 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
...
-        routeGetMembers(cluster) ~ routePostMembers(cluster)
+    rawPathPrefix(pathPrefixName) {
+      pathPrefix("members") {
+        pathEndOrSingleSlash {
+          routeGetMembers(cluster) ~ routePostMembers(cluster)
+        } ~
+        routesMember(cluster)
       } ~
-      routesMember(cluster)
+      pathPrefix("shard_regions" / Remaining) { shardRegionName =>
```
patriknw: perhaps just `"shards"` ?

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -124,24 +132,52 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
       }
     }
 
+  private def routeGetShardInfo(cluster: Cluster, shardRegionName: String) =
+    get {
+      extractExecutionContext { implicit executor =>
+        complete {
+          implicit val timeout = Timeout(1.second)
```
patriknw: 1 second might be too short, and I don't see a reason for having it too short. `5.seconds`? or is there some general http request timeout that we can use?

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -275,5 +286,91 @@ class ClusterHttpManagementRoutesSpec
...
+            |}
+            |akka.actor.provider = "cluster"
+            |akka.remote.log-remote-lifecycle-events = off
+            |akka.remote.netty.tcp.port = 0
+          """.stripMargin
+        )
+        val configClusterHttpManager = ConfigFactory.parseString(
+          """
+            |akka.cluster.http.management.hostname = "127.0.0.1"
+            |akka.cluster.http.management.port = 20100
```
patriknw: Instead of hardcoding the port you can use `akka.stream.testkit.TestUtils.temporaryServerAddress`. That avoids the risk of port conflicts if running several instances of the test at the same time (e.g. CI environment).

project/Dependencies.scala
```diff
@@ -14,13 +14,15 @@ object Dependencies {
...
     libraryDependencies ++= Seq(
-      "com.typesafe.akka" %% "akka-cluster"                      % AkkaVersion,
-      "com.typesafe.akka" %% "akka-http"                         % AkkaHttpVersion,
-      "com.typesafe.akka" %% "akka-http-spray-json"              % AkkaHttpVersion,
-      "io.spray"          %% "spray-json"                        % "1.3.2",                  // ApacheV2
-      "com.typesafe.akka" %% "akka-http-testkit"                 % AkkaHttpVersion % "test",
-      "junit"             % "junit"                              % junitVersion    % "test",
-      "org.mockito"       % "mockito-all"                        % "1.10.19"       % "test"  // Common Public License 1.0
+      "com.typesafe.akka" %% "akka-cluster"                       % AkkaVersion,
+      "com.typesafe.akka" %% "akka-cluster-sharding"              % AkkaVersion    % "provided",
```
patriknw: remove "provided", as implemented now the dependency is required and I think we agreed that would be fine

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
vikasuy: I couldn't find an http timeout in akka timeout, so i used 5 seconds

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
vikasuy: @patriknw i couldn't get the dependencies to work. 

First I added  this to library dependencies

`"com.typesafe.akka" %% "akka-stream-testkit"                % AkkaVersion     % "test"`

But that didn't make `TestUtils` accessible since its in the `test` configuration of `akka-stream-testkit` [source](https://github.com/akka/akka/blob/9d2bec7f232b628cc087231af75e457072823e61/akka-stream-testkit/src/test/scala/akka/stream/testkit/TestUtils.scala)

So then I tried

`"com.typesafe.akka" %% "akka-stream-testkit"                % AkkaVersion     % "test->test"`

And got this error:
```
[error] (cluster-http/*:update) sbt.ResolveException: unresolved dependency: com.typesafe.akka#akka-stream-testkit_2.11;2.4.18: configuration not public in com.typesafe.akka#akka-stream-testkit_2.11;2.4.18: 'test'. It was required from com.lightbend.akka#akka-management-cluster-http_2.11;0.2+11-7994a5b9+20170531-1122 test
```

let me know if you have any other ideas on how to do this. 
Note that the `ClusterHttpManagementSpec` also hardcodes in ports (which I agree is a bad pattern)

vikasuy:
@patriknw updated per your feedback

vikasuy:
@patriknw or @ktoso  -- how is this looking?

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -124,24 +132,52 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
...
    * Creates an instance of [[akka.cluster.http.management.ClusterHttpManagementRoutes]] to manage the specified
    * [[akka.cluster.Cluster]] instance. This version does not provide Basic Authentication. It uses
    * the specified path `pathPrefixName`.
    */
   def apply(cluster: Cluster, pathPrefixName: String): Route =
-    pathPrefix(pathPrefixName) {
-      pathEndOrSingleSlash {
-        routeGetMembers(cluster) ~ routePostMembers(cluster)
+    rawPathPrefix(pathPrefixName) {
+      pathPrefix("members") {
```
raboof: Hmm, this would break the API of any library user that specified a custom `pathPrefixName`, right?

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
raboof: I don't like `akka.stream.testkit.TestUtils.temporaryServerAddress` much anyway, it allows for a race condition still... ideally we'd bind using port `0` and then somehow inspect what port was automatically chosen to connect to it. But I suppose this is OK for now.

patriknw: I think the OS will typically not reuse the same port immediately

raboof: Ah yes, and that works because `Net.serverSocket` will set `SO_REUSEADDR` when binding for real. Not sure I expected that, but turns out nicely here :). (https://eklitzke.org/binding-on-port-zero)

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
vikasuy: it would. 
The old behavior was that the prefix would replace `members`. 
The behavior in this pr would change prefix to be a prefix of `members`. This is a more sustainable as we add more base REST resources to this library (in this case, `shards`). 

So here are our options:
1. make a breaking change to make `pathPrefix` a prefix of `/members, `/shards`, etc. (what this PR does as it stands). Perhaps, bump the version of this library to `0.4` to alert the breaking change. 

2. preserve old behavior: `pathPrefix` replaces `members` only, and create a new prefix parameter to prefix both `members` and `shards`, which we will use going forward. 

I like 1 because it's more flexible as we extend this library more and more. Since this library is pre 1.0 release, I feel like we have more flexibility to make these changes. That said @raboof let me know what you think

raboof: I guess it's OK since it's indeed early days for this library :+1:

patriknw: agreed

raboof:
I think we should add this feature to the documentation (https://github.com/akka/akka-management/blob/master/docs/src/main/paradox/cluster-http-management.md) and then this PR is good to go!

vikasuy:
@raboof good idea, added docs in cdc27e1b3c00ca06443b3a75d4cfff460ef5f863

PR #32, opened 2017-06-12
Whitesource

raboof:


project/Whitesource.scala
```diff
@@ -13,14 +13,13 @@ object Whitesource extends AutoPlugin {
...
-   )
-
-   def majorMinor(version: String): Option[String] =
-     """\d+\.\d+""".r.findFirstIn(version)
+      val projectName = (moduleName in LocalRootProject).value.replace("-root", "")
+      projectName + "-" + (
+        if (isSnapshot.value)
+          if (gitCurrentBranch.value == "master") "master"
+          else "adhoc"
+        else "stable"
```
2m: It is a bit unfortunate this this logic needs to be in every project that uses sbt-whitesource. Would it be possible to move that to the sbt plugin itself?

raboof: I agree it's unfortunate, though looking at the current git branch might not be something we'd like to promote universally...

patriknw:
@raboof ready to be merged?

raboof:
Jup!

PR #34, opened 2017-06-26
Update to akka-http 10.0.9

raboof:


project/Dependencies.scala
```diff
@@ -3,7 +3,7 @@ import sbt._, Keys._
 object Dependencies {
 
   val AkkaVersion = "2.4.18"
```
patriknw: could you update to 2.4.19 at the same time

Issue #28, opened 2017-05-12
Merging slashes for security reasons on URI's will break the REST interface


avalon12:
The REST interface of akka-management expects endpoints to be addressed in the following format:

`PUT /member/akka.tcp://mycluster@example.org:1234`

NGINX' default behaviour is to rewrite double slashes in the path of a URI to single slashes for security reasons (e.g. bypassing URL rewrites and deny, allow rules etc.)

A solution I would like to suggest is to add support for any encoding like base64 or URI encoding in the API to bypass double slash rewriting.

patriknw:
I agree we should support encoding, isn't `:` a also a problem?
As convenience I think we should also support address with only the `hostname:port` part, the prefix can be derived on the server side with `Cluster(system).selfAddress` because that is the same for all nodes in the cluster.

Would you be able to create a pull request @avalon12 ?

avalon12:
Base64 is not safe for urls since it can contain slashes (also multiple slashes after each other in rare cases)
URL encoding is also escaped by NGINX. 
So unless we find another encoding that suitable, we would have to find another way to avoid slashes.

Another option would be to use a pattern like PUT /member/<host>/<port>?cluster=<cluster>

patriknw:
as I tried to say, host and port is the any thing that is needed.
I like `host/port`
I think we can change the URIs to that. Any objections @hveiga ?

hveiga:
I think it makes sense. This sounds like a reasonable solution: `PUT /member/<host>/<port>?cluster=<cluster>`

patriknw:
but `?cluster=<cluster>` is not needed, you still want that as some kind of protection against human errors?

hveiga:
oh yes, you are right. In that case I would vote to remove it since it is not needed.
Then `PUT /member/<host>/<port>` should be sufficient.

raboof:
Should be fixed by #40

PR #40, opened 2017-09-13
Making the akka.tcp:// part in URI optional

sjoerdmulder:
This PR should resolve issue #28 by eliminating the requirement to have `akka.tcp://` in the URI. Instead you could just use `application@host:1234` to identify a member.

### DIFF
--- a/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
+++ b/cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
@@ -122,11 +122,9 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
       }
     }
 
-  private def findMember(cluster: Cluster, memberAddress: String): Option[Member] = {
-    cluster.readView.members.find(m ⇒
-      s"${m.uniqueAddress.address}" == memberAddress || m.uniqueAddress.address.hostPort == memberAddress
-    )
-  }
+  private def findMember(cluster: Cluster, memberAddress: String): Option[Member] =
+    cluster.readView.members.find(
+        m ⇒ s"${m.uniqueAddress.address}" == memberAddress || m.uniqueAddress.address.hostPort == memberAddress)
 
   private def routesMember(cluster: Cluster) =
     path(Remaining) { memberAddress ⇒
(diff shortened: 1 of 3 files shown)


raboof:
Hi, thanks for taking the time to put this together! It'd be really nice to be able to avoid encoding issues here. Did you have a reason not to go for the `PUT /member/<host>/<port>` pattern discussed in #28? That seems a bit safer still (though a slightly larger change)

raboof:
(the failing build is unrelated to your commit, I'll fix that)

sjoerdmulder:
Yes, the problem with the `<host>/<port>` is that it requires more extensive changes. And according to the [RFC(https://tools.ietf.org/html/rfc3986#section-3.3) for URI]  `@` or`:` are allowed in a path as long the path is not relative and prefixed with the authority. Since this has the lowest impact imho this is a better solution :see_no_evil:

avalon12:
From a RESTfull perspective, both solutions are equally correct. In this specific domain, the URI part `application@host:1234` can be seen as a single entity. Personally think both solutions are equally fine....

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -71,8 +71,8 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
 
         val leader = cluster.readView.leader.map(_.toString)
         val oldest = cluster.state.members.toSeq
-          .sorted(Member.ageOrdering)
           .filter(_.status == MemberStatus.Up)
+          .sorted(Member.ageOrdering)
```
ktoso: good

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -122,9 +122,13 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementHelper {
       }
     }
 
+  private def findMember(cluster: Cluster, memberAddress: String): Option[Member] =
+    cluster.readView.members.find(
+        m ⇒ s"${m.uniqueAddress.address}" == memberAddress || m.uniqueAddress.address.hostPort == memberAddress)
```
ktoso: good. Could have crazy edge cases when people restart into using artery suddenly, but that's crazy and let's not care here...

PR #42, opened 2017-09-22
Some changes for building under jdk9

raboof:


raboof:
A different error on travis than locally: `error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror', required by /home/travis/.sbt/boot/scala-2.10.6/lib/scala-library.jar(scala/package.class)`

project/NoPublish.scala
```diff
@@ -0,0 +1,14 @@
+import sbt._, Keys._
+
+/**
+ * For projects that are not to be published.
+ */
+object NoPublish extends AutoPlugin {
```
raboof: we might want to move this back to its original place once we get this to work

PR #50, opened 2017-10-09
Test multi-dc in a separate project

chbatey:
* Test multi-dc works for akak 2.5.6 keeping main dep at 2.4
* Upgrade to sbt 0.13.16
* Upgrade to akka 2.4.20

0.4 was broken as it compared nodes from different DCs, this test will
fail if that is re-introduced

### DIFF
--- a/cluster-http-tests/src/test/scala/akka/cluster/http/management/MultiDcSpec.scala
+++ b/cluster-http-tests/src/test/scala/akka/cluster/http/management/MultiDcSpec.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.cluster.http.management
+
+import akka.actor.ActorSystem
+import akka.cluster.Cluster
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.{ HttpRequest, StatusCodes }
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.ActorMaterializer
+import akka.testkit.SocketUtil
+import com.typesafe.config.ConfigFactory
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.{ Millis, Seconds, Span }
+import org.scalatest.{ Matchers, WordSpec }
+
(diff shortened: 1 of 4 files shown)


cluster-http-tests/src/test/scala/akka/cluster/http/management/MultiDcSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      val dcBSystem = ActorSystem("MultiDcSystem", config.withFallback(dcB))
+      implicit val materialiser = ActorMaterializer()
+
+      try {
+        val cluster = Cluster(dcASystem)
+        ClusterHttpManagement(cluster).start().futureValue
+        val response = Http().singleRequest(HttpRequest(uri = s"http://127.0.0.1:$httpPortA/members")).futureValue
+        response.status should equal(StatusCodes.OK)
+        val members = Unmarshal(response.entity).to[ClusterMembers].futureValue
+        members.members.size should equal(2)
```
chbatey: I could test more, before this it blew up entirely

patriknw: so it was already fixed in master?

chbatey: Yep, by accident by the looks of it: https://github.com/akka/akka-management/commit/e4fc9d5849342590daba3495597f68bd68887f9c

raboof: If that's the fix are we sure it's sufficient? It looks like it'd still blow up when members from different dc's are in the status `MemberStatus.Up` (https://github.com/akka/akka-management/commit/e4fc9d5849342590daba3495597f68bd68887f9c#diff-c6868e05e6924399e82f52ce16d1a10cR73)

raboof: Ah sorry already fixed further in https://github.com/akka/akka-management/pull/52/files#diff-c6868e05e6924399e82f52ce16d1a10cL74

PR #55, opened 2017-11-22
SVC DNS and Contact Point HTTP lookup based joining procedure 

ktoso:
Work in progress of a proposal to simplify cluster joining when deployed in envs with managed DNS for services, such as kubernetes. Note that this does not require the use of stateful sets, which we're aware of but avoiding for now since they've been reported to be not yet fully reliable. We're keeping an eye on them and provide such implementation as well once the time is right.

This is a Work In Progress utilizing kubernetes's SVC ("service") DNS records, when a service is deployed "headless".


joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -0,0 +1,87 @@
...
+          Future.successful(Done)
+        case _ ⇒ Future.failed(new Exception("NOPE"))
+      }
+  }
+}
+
+object ClusterBootstrap extends ExtensionId[ClusterBootstrap] with ExtensionIdProvider {
+  override def lookup: ClusterBootstrap.type = ClusterBootstrap
+
+  override def get(system: ActorSystem) = super.get(system)
```
patriknw: explicit return type

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,230 @@
...
+    // TODO as it is right now, this just means that one of the contact points was stable and did not return any seed nodes
+    // TODO this is likely fine, but we could add additional guards here, like:
+    // TODO  - like the dns IPs we get back also being stable for the stableMargin
+    // TODO  - or all children we spawned having reported that their deadlines passed
+    // this moment by definition "risky" though, so a matter of how much we think that would help
+  }
+
+  /** Awaiting joining to complete before shutting down the bootstrap actors */
+  def joining(seedNodes: immutable.Seq[Address]): Receive = {
+    case initialState: CurrentClusterState ⇒
```
patriknw: This will be sent immediately after subscribe, with current state, which might be before it has joined. Not sure what the purpose of this subscription is?
You could subscribe to `MemberJoined` but I don't see much point in waiting for that. After `cluster.joinSeedNodes` it's done.

joining/src/main/scala/akka/cluster/bootstrap/http/ClusterBootstrapRoutes.scala
```diff
@@ -0,0 +1,55 @@
...
+  import akka.http.scaladsl.server.Directives._
+
+  private def routeGetSeedNodes: Route = extractActorSystem { implicit system ⇒
+    import system.dispatcher
+    val cluster = Cluster(system)
+
+    def memberToClusterMember(m: Member): ClusterMember =
+      ClusterMember(m.uniqueAddress.address, m.uniqueAddress.longUid, m.status.toString, m.roles)
+
+    val readView = cluster.readView
```
patriknw: that's internal, I think you can get same from `cluster.state`

ktoso:
Last commit makes through all the steps of discovering the nodes, and probing them via http:

```
[DEBUG] [11/22/2017 18:10:07.997] [DemoApp-akka.actor.default-dispatcher-2] [DemoApp/Pool(shared->http://172.17.0.3:8558)] [0] </172.17.0.11:55728->172.17.0.3:8558> Received response: GET /bootstrap/seed-nodes Empty -> 200 OK Strict(64 bytes)
[DEBUG] [11/22/2017 18:10:08.003] [DemoApp-akka.actor.default-dispatcher-17] [DemoApp/Pool(shared->http://172.17.0.3:8558)] [0] </172.17.0.11:55728->172.17.0.3:8558> Finished reading response entity for GET /bootstrap/seed-nodes Empty -> 200 OK Strict(64 bytes)
[DEBUG] [11/22/2017 18:10:08.008] [DemoApp-akka.actor.default-dispatcher-5] [DemoApp/Pool(shared->http://172.17.0.13:8558)] [0] </172.17.0.11:49734->172.17.0.13:8558> Received response: GET /bootstrap/seed-nodes Empty -> 200 OK Strict(65 bytes)
[DEBUG] [11/22/2017 18:10:08.012] [DemoApp-akka.actor.default-dispatcher-17] [DemoApp/Pool(shared->http://172.17.0.13:8558)] [0] </172.17.0.11:49734->172.17.0.13:8558> Finished reading response entity for GET /bootstrap/seed-nodes Empty -> 200 OK Strict(65 bytes)
[DEBUG] [11/22/2017 18:10:08.016] [DemoApp-akka.actor.default-dispatcher-3] [DemoApp/Pool(shared->http://172.17.0.12:8558)] [0] </172.17.0.11:52732->172.17.0.12:8558> Received response: GET /bootstrap/seed-nodes Empty -> 200 OK Strict(65 bytes)
```

It seems the actual "yep, time to join myself!" has some minor bug yet. Will fix tomorrow and seems we have our basic version running.

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+// TODO security for the contact point requests?
+// FIXME FIXME !! redesign the settings !! half of those are kubernetes and DNS specific
+/**
+ *
+ * @param serviceName if None the actor system name will be used
+ */
+final case class ClusterBootstrapSettings(
+    bootstrapMethod: String = "dns-svc",
+    serviceName: Option[String] = Some("appka-service"), // k8s specific ??
+    namespaceName: Option[String] = Some("default"), // k8s specific ??
```
fsat: I think the `default` namespace is Kubernetes specific - Mesos has a different naming convention for its SRV entries - it's documented in `SRV Records` in this link:

https://mesosphere.github.io/mesos-dns/docs/naming.html

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  private def onSufficientContactPointsDiscovered(serviceName: String,
+                                                  observation: DnsServiceContactsObservation): Unit = {
+    log.info("Initiating contact-point probing, sufficient contact points: {}",
+      observation.observedContactPoints.mkString(", "))
+
+    observation.observedContactPoints.foreach { contactPoint ⇒
+      // TODO would be nice if we can discover this via DNS as well? and leave option to override here
+      val targetPort = settings.httpContactPointPort
```
fsat: @ktoso - we can discover the target port when we received the `SrvResolved` message from `akka-dns`:

```
object AsyncDnsResolver {
  ...
  case class SrvResolved(name: String, srv: immutable.Seq[SRVRecord])
}

case class SRVRecord(override val name: String, override val ttl: Int,
                     priority: Int, weight: Int, port: Int, target: String) extends ResourceRecord(name, ttl, RecordType.SRV.id.toShort, RecordClass.IN.id.toShort) {
...
}
```

I can see these options in relation to obtaining the value of the port comes from:

1. Only accept SRV record as this would have the actual port number

This means the normal DNS A-record (non-SRV) is ignored/not accepted as it only has address without the port information. This option also means the Akka cluster won't be formed unless there are SRV record present for the seed nodes.

2. Use the port from SRV record if exists, otherwise falling back using A-Record with `settings.httpContactPoint`

This means the DNS A-record (non-SRV) is also accepted with the port value coming from `settings.httpContactPoint`. Maybe have config like `resolve-from-dns-a-record=true`.

As Kubernetes and DCOS both have DNS SRV support, they only need Option 1, and not Option 2.

I was initially thinking about non orchestration deployment, but then AWS Route 53 and GCP Cloud DNS also support SRV records, so perhaps Option 2 is not immediately required.

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+  private def scheduleNextResolve(serviceName: String, resolveTimeout: FiniteDuration): Unit =
+    timers.startSingleTimer(TimerKeyResolveDNS, Internal.AttemptResolve(serviceName, resolveTimeout),
+      settings.dnsResolveInterval)
+
+  private def resolve(name: String, resolveTimeout: FiniteDuration): Future[DnsResolvedTo] = {
+    def cleanIpString(ipString: String) =
+      ipString.replaceAll("/", "")
+
+    dns.ask(Dns.Resolve(name))(resolveTimeout) map {
+      case srv: SrvResolved =>
```
fsat: You can extract the ports from `srv`, i.e. `srv.srv.map(v => cleanIpString(v.target) -> v.port)` gives you address and port pair.

ktoso: Related problem, at least in minikube is that I'm getting back A records actually, so I have to figure out why that is, see:

```
[akka.tcp://DemoApp@172.17.0.8:2552/system/IO-DNS/async-dns/$a/$a] Decoded: Message(2,<AN,QUERY,RD,RA,SUCCESS>,Vector(Question(appka-service.default.svc.cluster.local,A,IN)),Vector(ARecord(appka-service.default.svc.cluster.local,30,/172.17.0.6), ARecord(appka-service.default.svc.cluster.local,30,/172.17.0.9), ARecord(appka-service.default.svc.cluster.local,30,/172.17.0.8), ARecord(appka-service.default.svc.cluster.local,30,/172.17.0.7)),Vector(),Vector())
```

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    // TODO make those DNS lookup settings
+    dnsSuffix: String = ".svc.cluster.local", // k8s specific ??
```
fsat: I think this is Kubernetes specific too.

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
fsat: Perhaps we can try `akka.io.dns.resolve-srv=true` setting? This will enable SRV record resolve, and this setting is set to `false` by default:

https://github.com/ilya-epifanov/akka-dns#advanced-configuration

I'm asking because I can't seem to find this setting being set in the branch.

ktoso: lol, great catch -- I'm blind, missed that setting :)

fsat: 😄 all good 👍

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
ktoso: Thanks for the info! I guess we should pre-package good defaults for the clouds we want to support initially - can you help me find the names/defaults in the other systems?

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Locale
+
+import akka.actor.ActorSystem
+import com.typesafe.config.Config
+
+import scala.concurrent.duration._
+import scala.concurrent.duration.FiniteDuration
+
+// FIXME make easy to binary evolve (kaze class)
```
patriknw: yes, and make it possible to create from Config

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    // TODO make those http contact point settings
+    httpContactPointPort: Int = 8558, // TODO I wonder if we could discover those as well, but likely not needed to be honest?
+    contactPointNoSeedsStableMargin: FiniteDuration = 10.seconds,
+    httpProbeInterval: FiniteDuration = 2.second,
+    httpProbeIntervalJitter: Double = 0.2, // +/- 20%
+    httpMaxSeedNodesToExpose: Int = 5
+) {
+
+  require(requiredContactPointsNr > 2, "Number of contact points is strongly recommended to be greater than 2")
```
patriknw: 2 should still be possible

joining/src/main/scala/akka/cluster/bootstrap/DemoApp.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import akka.actor.{ Actor, ActorLogging, ActorSystem, Props }
+import akka.cluster.ClusterEvent.ClusterDomainEvent
+import akka.cluster.{ Cluster, ClusterEvent }
+import akka.http.scaladsl.Http
+import akka.io.{ Dns, IO }
+import akka.stream.ActorMaterializer
+import com.typesafe.config.ConfigFactory
+
+object DemoApp extends App {
```
patriknw: move to test src

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+    /** Contact point with the "lowest" address, it is expected to join itself if no other cluster is found in the deployment. */
+    def lowestAddressContactPoint: Option[String] = observedContactPoints.sorted.headOption
+
+    def willBeStableAt(settings: ClusterBootstrapSettings): Long =
+      observedAt + settings.dnsStableMargin.toMillis
+
+    def isPastStableMargin(settings: ClusterBootstrapSettings, timeNow: Long): Boolean =
+      willBeStableAt(settings) < timeNow
+
+    def durationSinceObservation(timeNowMillis: Long): Duration = {
```
patriknw: FiniteDuration?

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+      val millisSince = timeNowMillis - observedAt
+      math.max(0, millisSince).millis
+    }
+
+    def membersChanged(other: DnsServiceContactsObservation): Boolean = {
```
patriknw: s/membersChanged/contactPointsChanged/

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+    if (contactPoints.size < settings.requiredContactPointsNr)
+      onInsufficientContactPointsDiscovered(serviceName, lastContactsObservation)
+    else
+      onSufficientContactPointsDiscovered(serviceName, lastContactsObservation)
+  }
+
+  private def onInsufficientContactPointsDiscovered(serviceName: String,
+                                                    observation: DnsServiceContactsObservation): Unit = {
+    log.info("Discovered ({}) observation, which is less than the required ({}), retrying (interval: {})",
+      observation.observedContactPoints.size, settings.requiredContactPointsNr, settings.dnsResolveTimeout)
```
patriknw: pretty duration?

ktoso:
TODO: If contacted to contact point, from IP that we don't know from DNS, consider that the DNS view is not stable yet?

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+
+    case ex: Failure ⇒
+      log.warning("Resolve attempt failed! Cause: {}", ex.cause)
+      scheduleNextResolve(serviceName, settings.dnsResolveTimeout)
+
+    case ObtainedHttpSeedNodesObservation(infoFromAddress, observedSeedNodes) ⇒
+      log.info("Contact point [{}] returned [{}] seed-nodes [{}], initiating cluster joining...", infoFromAddress,
+        observedSeedNodes.size, observedSeedNodes)
+
+      replyTo ! "JOINING SEED NODES!" // FIXME nicer typed message
```
patriknw: yes

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+
+      val seedNodesList = observedSeedNodes.toList
+      cluster.joinSeedNodes(seedNodesList)
```
patriknw: This is used for joining existing cluster. Remove the selfAddress from the seedNodesList so that there is no chance that it's joining itself.

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -0,0 +1,87 @@
...
+      implicit val initTimeout: Timeout = Timeout(1.day) // configure? or what should it do really? try until forever meh?
+
+      val initCompleted = Promise[Done]()
+
+      // TODO load the right one depending on config / env?
+      // TODO should get it's own config then too
+      val bootstrapManager =
+        system.systemActorOf(HeadlessServiceDnsBootstrap.props(settings), "headlessServiceDnsBootstrap")
+      val reply = (bootstrapManager ? HeadlessServiceDnsBootstrap.Protocol.InitiateBootstraping).mapTo[
+          HeadlessServiceDnsBootstrap.Protocol.BootstrapingCompleted]
```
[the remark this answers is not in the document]
patriknw: if we decide to keep this as separate module I think it should still include the routes from ordinary cluster management

joining/src/main/scala/akka/cluster/bootstrap/http/ClusterBootstrapRoutes.scala
```diff
@@ -0,0 +1,71 @@
...
+      def memberToClusterMember(m: Member): ClusterMember =
+        ClusterMember(m.uniqueAddress.address, m.uniqueAddress.longUid, m.status.toString, m.roles)
+
+      val state = cluster.state
+
+      // TODO shuffle the members so in a big deployment nodes start joining different ones and not all the same?
+      val members = state.members.take(settings.httpMaxSeedNodesToExpose).map(memberToClusterMember)
+
+      // TODO add a method to find oldest to cluster state?
+      val oldest = state.members.toSeq
```
patriknw: not used?

joining/src/main/scala/akka/cluster/bootstrap/http/ClusterBootstrapRoutes.scala
```diff
@@ -0,0 +1,71 @@
...
+        .filter(node => node.status == MemberStatus.Up && node.dataCenter == cluster.selfDataCenter)
+        .sorted(Member.ageOrdering)
+        .headOption // we are only interested in the oldest one that is still Up
+        .map(_.uniqueAddress.address)
+
+      val info = SeedNodes(cluster.selfMember.uniqueAddress.address, members, oldest)
+      log.info("Bootstrap request from {}: Contact Point returning {} seed-nodes ([{}])", clientIp, members.size,
+        members)
```
patriknw: members.mkString...

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+    case DnsResolvedTo(name, contactPoints) ⇒
+      onContactPointsResolved(name, contactPoints)
+
+    case ex: Failure ⇒
+      log.warning("Resolve attempt failed! Cause: {}", ex.cause)
+      scheduleNextResolve(serviceName, settings.dnsResolveTimeout)
+
+    case ObtainedHttpSeedNodesObservation(infoFromAddress, observedSeedNodes) ⇒
+      log.info("Contact point [{}] returned [{}] seed-nodes [{}], initiating cluster joining...", infoFromAddress,
+        observedSeedNodes.size, observedSeedNodes)
```
patriknw: observedSeedNodes.map...mkString

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+
+    if (contactPoints.size < settings.requiredContactPointsNr)
+      onInsufficientContactPointsDiscovered(serviceName, lastContactsObservation)
+    else
+      onSufficientContactPointsDiscovered(serviceName, lastContactsObservation)
+  }
+
+  private def onInsufficientContactPointsDiscovered(serviceName: String,
+                                                    observation: DnsServiceContactsObservation): Unit = {
+    log.info("Discovered ({}) observation, which is less than the required ({}), retrying (interval: {})",
```
patriknw: we normally use `[ ]` instead of `( )` around such values

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+          contactPoint) // TODO debug
+
+      onNoSeedNodesObtainedWithinStableDeadline(contactPoint)
+  }
+
+  private def onContactPointsResolved(serviceName: String, contactPoints: immutable.Seq[String]): Unit = {
+    val newObservation = DnsServiceContactsObservation(timeNow(), contactPoints.toList)
+    lastContactsObservation = lastContactsObservation.sameOrChanged(newObservation)
+
+    if (contactPoints.size < settings.requiredContactPointsNr)
```
patriknw: as discussed in Slack we should perhaps rename it to `initial-nr-of-contact-points` and it must be exactly that number of contact points before self-joining.

I'm thinking about if it's needed to wait here before starting probing? It should be enough with one other contact point and if that returns some seed-nodes we can join that?

For example you might want to set `initial-nr-of-contact-points` to 5, becuase you know that you are starting 5 nodes initially, but for later restarts it should be possible to join with less than 5 nodes. E.g. 3 of 5 nodes crash and 2 nodes are started again it's better that those can join than having to wait for a 3rd.

joining/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,278 @@
...
+        val props = HttpContactPointBootstrap.props(settings, self, baseUri)
+        context.actorOf(props, childActorName)
+    }
+  }
+
+  private def scheduleNextResolve(serviceName: String, resolveTimeout: FiniteDuration): Unit =
+    timers.startSingleTimer(TimerKeyResolveDNS, Internal.AttemptResolve(serviceName, resolveTimeout),
+      settings.dnsResolveInterval)
+
+  private def resolve(name: String, resolveTimeout: FiniteDuration): Future[DnsResolvedTo] = {
```
longshorej: Would it be possible to make this modular? The real concept here is just service discovery, right? It would be nice if we could provide our own module that actually ties this to DNS SRV or whatever else we'd like to.  As an example, I mean something like:

```
trait ServiceDiscovery {
  def lookup(name: String, resolveTimeout: Future[ResolvedTo] 
}

// then perhaps you provide a default impl with the library

class DnsSrvServiceDiscovery {
  def lookup(name: String, resolveTimeout: Future[ResolvedTo] =
    dns.ask(Dns.Resolve(name))(resolveTimeout) map {
      ...
    }
}

// and in reference.conf

akka.management.service-discovery-class = "akka.cluster.bootstrap.dns.DnsSrvServiceDiscovery"

```

Felix and I could then provide our own implementation in `reactive-lib` that inspects its runtime environment to decide if discovery is done via DNS SRV, HTTP, or whatever else.

patriknw: I think that is a good idea.

Impl can optionally take ActorSystem as param, as we do in many other places.

patriknw: Btw, that will probably also simplify testing, using a stub.

ktoso: Yeah I’ll do that :)

joining/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
longshorej: Should 1 be possible as well i.e. dev environments?

cluster-http/src/test/scala/akka/cluster/bootstrap/contactpoint/HttpContactPointRoutesSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+
+    "include seed nodes when part of a cluster" in {
+      val cluster = Cluster(system)
+      cluster.join(cluster.selfAddress)
+
+      val p = TestProbe()
+      cluster.subscribe(p.ref, ClusterEvent.InitialStateAsEvents, classOf[ClusterDomainEvent])
+      val up = p.expectMsgType[ClusterEvent.MemberUp]
```
patriknw: when subscribing to all ClusterDomainEvent I don't think it's guaranteed that MemberUp will be the first, change the `subscribe` to only `MemberUp`

cluster-http/src/main/scala/akka/discovery/DnsSrvServiceDiscovery.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  val config = new DnsSrvServiceDiscoverySettings(effectiveConfig)
+
+  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[ServiceDiscovery.Resolved] = {
+    def cleanIpString(ipString: String): String =
+      if (ipString.startsWith("/")) ipString.tail else ipString
+
+    dns.ask(Dns.Resolve(name))(resolveTimeout) map {
+      case srv: SrvResolved =>
+        log.debug("Resolved Srv.Resolved: " + srv)
```
patriknw: use `{}` placeholders

cluster-http/src/main/scala/akka/discovery/DnsSrvServiceDiscovery.scala
```diff
@@ -0,0 +1,52 @@
...
+
+      case resolved: Dns.Resolved =>
+        log.debug("Resolved Dns.Resolved: " + resolved)
+        val addresses = resolved.ipv4.map { entry ⇒
+          ServiceDiscovery.ResolvedTarget(cleanIpString(entry.getHostAddress), None)
+        }
+        ServiceDiscovery.Resolved(name, addresses)
+
+      case resolved ⇒
+        log.warning("Resolved UNEXPECTED (resolving to Nil): " + resolved)
```
patriknw: toString of unknown things is scary, use class name instead

project/Dependencies.scala
```diff
@@ -17,11 +19,15 @@ object Dependencies {
       "com.typesafe.akka" %% "akka-cluster-sharding"              % AkkaVersion,
       "com.typesafe.akka" %% "akka-http"                          % AkkaHttpVersion,
       "com.typesafe.akka" %% "akka-http-spray-json"               % AkkaHttpVersion,
-      "io.spray"          %% "spray-json"                         % "1.3.3",                  // ApacheV2
+      "io.spray"          %% "spray-json"                         % SprayJsonVersion,                  // ApacheV2
+
+      "ru.smslv.akka"     %% "akka-dns"                           % akkaDns,
```
patriknw: I get unresolved dependency: ru.smslv.akka#akka-dns_2.11;2.4.2: not found

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -384,21 +387,41 @@ class ClusterHttpManagement(
...
+      // Basically: "give me the SRV host/port for the port called `akka-bootstrap`"
+      // discovery.lookup("_akka-bootstrap" + ".effective-name.default").find(myaddress)
+      // ----
+
+      // FIXME -- think about the style of how we want to make these available
+      // I was rather thinking that management extensions should be able to be registered somehow
+      // and then be included in here
+      val bootstrapSettings = ClusterBootstrapSettings(
+          ConfigFactory
+            .parseString("akka.cluster.bootstrap.contact-point.port-fallback = ")
```
patriknw: that looks strange, no value for the port-fallback?

cluster-http/src/main/scala/akka/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -0,0 +1,81 @@
...
+    val discoveryClass: String = discoveryConfig.getString("discovery-method")
+
+    val stableMargin: FiniteDuration =
+      effectiveDiscoveryConfig.getDuration("stable-margin", TimeUnit.MILLISECONDS).millis
+
+    val interval: FiniteDuration =
+      effectiveDiscoveryConfig.getDuration("interval", TimeUnit.MILLISECONDS).millis
+
+    val requiredContactPointsNr = discoveryConfig.getInt("required-contact-point-nr")
+    require(requiredContactPointsNr >= 2, "Number of contact points ")
```
patriknw: complete the error message

cluster-http/src/main/resources/reference.conf
```diff
@@ -16,3 +16,65 @@ akka.cluster.http.management {
...
+######################################################
+
+
+akka cluster.bootstrap {
+
+  contact-point-discovery {
+
+    # define this name to be looked up in service discovery for "neighboring" nodes
+    # if undefined, the name will be extracted from the ActorSystem name
+    service-name = null
```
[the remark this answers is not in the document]
ktoso: Ok, can do "" and treat it as empty.

Yeah the null allows documeting the key here, but if you check `hasPath` it will return false for the given key. So it's like an option. But you're right, let's just use ""

cluster-http/src/main/scala/akka/cluster/bootstrap/BootstrapMode.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+package akka.cluster.bootstrap
+
+/**
+ * Dictates mode in which cluster bootstraping will be initiated.
+ *
+ * Generally this will depend on the cloud platform or,
+ * available external service discovery services you want to integrate with.
+ */
+sealed trait BootstrapMode
```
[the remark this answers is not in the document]
ktoso: Yeap removed

cluster-http/src/main/scala/akka/discovery/DnsSrvServiceDiscovery.scala
ktoso: thanks

cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
ktoso: yeah that's from my "let me push before weekend" now cleanedup

cluster-http/src/main/scala/akka/discovery/DnsSrvServiceDiscovery.scala
ktoso: Ok will do

project/Dependencies.scala
ktoso: Whoa weird, where did I get the artifact from hm...

cluster-http/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,272 @@
...
+    case ServiceDiscovery.Resolved(name, contactPoints) ⇒
+      onContactPointsResolved(name, contactPoints)
+
+    case ex: Failure ⇒
+      log.warning("Resolve attempt failed! Cause: {}", ex.cause)
+      scheduleNextResolve(serviceName, settings.contactPointDiscovery.interval)
+
+    case ObtainedHttpSeedNodesObservation(infoFromAddress, observedSeedNodes) ⇒
+      log.info("Contact point [{}] returned [{}] seed-nodes [{}], initiating cluster joining...", infoFromAddress,
+        observedSeedNodes.size, observedSeedNodes)
```
patriknw: `observedSeedNodes.mkString....`

cluster-http/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,272 @@
...
+      log.info("Locating service members, via DNS lookup: {}", serviceName)
+      discovery.lookup(serviceName, settings.contactPointDiscovery.resolveTimeout).pipeTo(self)
+
+      context become bootstraping(serviceName, sender())
+  }
+
+  /** In process of searching for seed-nodes */
+  def bootstraping(serviceName: String, replyTo: ActorRef): Receive = {
+    case Internal.AttemptResolve(name) ⇒
+      discovery.lookup(name, settings.contactPointDiscovery.resolveTimeout).pipeTo(self)
```
patriknw: can that Future fail, and `Status.Failure` should be handled?

cluster-http/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -0,0 +1,272 @@
...
+        // nothing to do anymore, the probing will continue until the lowest addressed node decides to join itself.
+        // note, that due to DNS changes this may still become this node! We'll then await until the dns stableMargin
+        // is exceeded and would decide to try joining self again (same code-path), that time successfully though.
+      }
+
+    } else {
+      // TODO throttle this logging? It may be caused by any of the probing actors
+      log.debug(
+          "DNS observation has changed more recently than the dnsStableMargin({}) allows (at: {}), not considering to join myself. " +
+          "This process will be retried.", settings.contactPointDiscovery.stableMargin,
```
patriknw: pretty duration

build.sbt
```diff
@@ -12,7 +12,8 @@ lazy val `cluster-http` = project
   .enablePlugins(AutomateHeaderPlugin)
   .settings(
     name := "akka-management-cluster-http",
-    Dependencies.ClusterHttp
+    Dependencies.ClusterHttp,
+    resolvers += Resolver.bintrayRepo("hajile", "maven")
```
ktoso: Oh i see... so we may have to publish it ourselfs?

ktoso: Does that mean we have to make sure it gets to maven central?

ktoso: (fork it or ask the maintainer)

ktoso:
Thanks for merging, I'll follow up with tests and addressing the additional cases (same IP address) tomorrow

cluster-http/src/main/scala/akka/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
ktoso: Good point, could fail with a timeout; will handle

PR #58, opened 2017-11-28
Bug fixes

longshorej:
Thanks again for getting this done, we really appreciate it. I've begun integration with `reactive-lib` and was able to get a working implementation with the changes in this PR.

1) Fix bug `selfAddressIfAbleToJoinItself` would always return `None` since it was comparing unrelated types
2) Ensure that probe is retried when HTTP request fails

Also, just to note, we're not using the supplied `DnsSrvServiceDiscovery` but instead are providing our own implementation, but I did notice that `DnsSrvServiceDiscovery` cannot be instantiated because it takes an `effectiveConfig` and the reflection code is only passing it a single argument (`system`)


lightbend-cla-validator:
Hi @longshorej,

patriknw:
Superseded by https://github.com/akka/akka-management/pull/60

PR #60, opened 2017-11-28
nodes now can figure out their address to join, even if random

ktoso:
Resolves #55  and adds tests for joining as well as the routes.
Also some cleanups.

Please review (merge at will), will continue with more docs and may find some cleanups along the way tomorrow

### DIFF
--- a/cluster-http/src/main/scala/akka/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-http/src/main/scala/akka/cluster/bootstrap/ClusterBootstrap.scala
@@ -1,24 +1,29 @@
 /*
- * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
  */
 package akka.cluster.bootstrap
 
 import java.util.concurrent.atomic.AtomicReference
 
 import akka.actor.{ ActorSystem, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider }
+import akka.annotation.InternalApi
 import akka.cluster.bootstrap.dns.HeadlessServiceDnsBootstrap
 import akka.discovery.ServiceDiscovery
 import akka.event.Logging
+import akka.http.scaladsl.model.Uri
 import akka.pattern.ask
 import akka.stream.ActorMaterializer
 import akka.util.Timeout
 
+import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success }
 
 final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Extension {
+
   import ClusterBootstrap._
   import system.dispatcher
+
   private implicit val mat = ActorMaterializer()(system)
 
   private val log = Logging(system, classOf[ClusterBootstrap])
@@ -32,6 +37,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     val clazz = settings.contactPointDiscovery.discoveryClass
     system.dynamicAccess.createInstanceFor[ServiceDiscovery](clazz, List(classOf[ActorSystem] → system)).get
   }
+
+  private[this] val _selfContactPointUri: Promise[Uri] = Promise()
 
   def start(): Unit =
     if (bootstrapStep.compareAndSet(NotRunning, Initializing)) {
@@ -50,8 +57,25 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
         case Success(_) ⇒ // ignore, all's fine
         case Failure(_) ⇒ log.warning("Failed to complete bootstrap within {}!", bootTimeout)
       }
+    } else log.warning("Bootstrap already initiated, yet start() method was called again. Ignoring.")
 
-    }
+  /**
+   * INTERNAL API
+   *
+   * Must be invoked by whoever starts the HTTP server with the `HttpClusterBootstrapRoutes`.
+   * This allows us to "reverse lookup" from a lowest-address sorted contact point list,
+   * that we discover via discovery, if a given contact point corresponds to our remoting address,
+   * and if so, we may opt to join ourselves using the address.
+   *
+   * @return true if successfully set, false otherwise (i.e. was set already)
+   */
+  @InternalApi
+  def setSelfContactPoint(baseUri: Uri): Boolean =
+    _selfContactPointUri.trySuccess(baseUri)
+
+  /** INTERNAL API */
+  private[akka] def selfContactPoint: Future[Uri] =
+    _selfContactPointUri.future
 
 }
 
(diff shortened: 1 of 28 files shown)


cluster-http/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -407,39 +218,32 @@ class ClusterHttpManagement(
 
       val bootstrapSettings = ClusterBootstrapSettings(bootstrapConfig)
       val bootstrapContactPointRoutes = new HttpClusterBootstrapRoutes(bootstrapSettings)
+      ClusterBootstrap(system).setSelfContactPoint(Uri(s"http://$hostname:$port"))
```
raboof: This won't work when the port is 'ephemeral' (0/auto-assigned by the OS), right? and in case of NAT?

project/Common.scala
```diff
@@ -9,7 +9,7 @@ object Common extends AutoPlugin {
 
   val FileHeader = (HeaderPattern.cStyleBlockComment,
     """|/*
-       | * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+       | * Copyright (C) 2017 LightbendInc. <http://www.lightbend.com>
```
raboof: LightbendInc?

cluster-http/src/main/scala/akka/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -33,6 +37,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     system.dynamicAccess.createInstanceFor[ServiceDiscovery](clazz, List(classOf[ActorSystem] → system)).get
   }
 
+  @volatile private[this] var _selfContactPointUri: Uri = _
```
raboof: would it 'ripple through' too much to make this a promise?

ktoso: Hmm, that could work :)

project/Common.scala
ktoso: lol, thanks :)

ktoso:
Back to this and documenting finally

raboof:
Not introduced in this PR, but I'd like to understand better how this bootstrapping approach ties in with k8s readiness/healthiness checks: it'd seem that since we discover other nodes via DNS, they have to be marked ready/healthy before they have successfully joined the cluster. How do we prevent nodes from receiving traffic before joining the cluster?

cluster-http/src/main/scala/akka/discovery/DnsSrvServiceDiscovery.scala
```diff
@@ -1,9 +1,9 @@
 /*
- * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
  */
 package akka.discovery
 
-import akka.actor.ActorSystem
+import akka.actor.{ ActorSystem, ExtendedActorSystem }
 import akka.event.Logging
 import akka.io.AsyncDnsResolver.SrvResolved
```
patriknw: This class is from ru akka-dns, so I'm still confused about this dependency.

ktoso: Right, please look at, my statement in the standup was incorrect -- i assumed I don't use it because I forgot it has packages in the "wrong package". and grepped the codebase so did not notice any `ru.` code.

Indeed this is the russian one.

Please look at the latest commit here https://github.com/akka/akka-management/pull/60/commits/5e3d773c92a14b39e85ce45437e2e4f0757ea007 which makes it really work without the dependency, but with limitations that we can't discover the management port from DNS records

ktoso:
Please see https://github.com/akka/akka-management/pull/60/commits/5e3d773c92a14b39e85ce45437e2e4f0757ea007 for the latest and "without russian dns library"

patriknw:
@longshorej Konrad has removed the dependency to ru akka-dns.
It had support for SRV records which we can get the port from, but I guess that can be added later.
Does that resolve your dependency concern?

ktoso:
I pushed initial docs: https://github.com/akka/akka-management/pull/60/commits/1639baf8d435d1844cc0c6da7ba585e60672a52d

They're nowhere as detailed as I want them to be, to really explain that this works and is fine, but it's a start. I think that this deserves some pictures so people really understand it etc.

longshorej:
Thank you - It fixes the biggest issue for now and we can move forward with some validation logic in our SBT plugin to make sure versions of the various akka modules are the same.

I don't think it is perfect, however. I have two thoughts. I don't think we need to do anything right now but it'll be useful to discuss it at least.

First, there's a dependency on `akka-cluster-sharding` which I don't think is strictly required for cluster formation. Changing this would require refactoring the project structure though.

Second, suppose you have a Lagom project using persistence: It automatically includes `akka-cluster` and `akka-persistence` as a dependency in your project. Most users don't explicitly declare a dependency on these modules so the version used is the one declared by Lagom.

Then, you add `akka-management` (or in our case, our SBT plugin which has a transitive dependency on `akka-management`). Let's then assume the version of `akka-cluster` that `akka-management` depends on is a different patch version than the one Lagom is declaring. The project would then have dependencies for one version of `akka-cluster` but a different version of `akka-persistence`, which causes breakage as all modules need to be on the same version.

A solution to this problem is to compile `akka-management` with `akka-cluster` but leave it out of the dependency metadata. `sbt-assembly` supports this via using a `provided` scope for such dependencies. I think this means the dependencies would look something like this:

```
    libraryDependencies ++= Seq(
      "com.typesafe.akka" %% "akka-cluster"                     % AkkaVersion         % "provided",
      "com.typesafe.akka" %% "akka-http"                          % AkkaHttpVersion,
      "com.typesafe.akka" %% "akka-http-spray-json"         % AkkaHttpVersion,
      "io.spray"                  %% "spray-json"                         % SprayJsonVersion
    )
```

As mentioned above, I think what we'll do for now is validate known modules to make sure versions line up. If they don't, we'll throw an exception telling the user to explicitly declare dependencies on the necessary libraries.

Does that make sense? Is there something I'm not thinking about?

patriknw:
I agree that we should probably make things more modular, especially make it easier to compose different Routes to add more things. I have added issue https://github.com/akka/akka-management/issues/61 for that.

Lagom projects will anyway use akka-cluster-sharding if they use this. akka-cluster-sharding has dependency to akka-persistence so I'm not sure that your argument there is correct, as long as we have sharding included here.

longshorej:
> Not introduced in this PR, but I'd like to understand better how this bootstrapping approach ties in with k8s readiness/healthiness checks: it'd seem that since we discover other nodes via DNS, they have to be marked ready/healthy before they have successfully joined the cluster. How do we prevent nodes from receiving traffic before joining the cluster?

That's a good thought. In our reactive-lib project, it currently uses the service locator under the hood, which on K8s is DNS-based and thus is prone to this issue.

What we're planning to do to solve this problem is, for our implementation of `ServiceDiscovery` trait, use the Kubernetes (and DC/OS API later on) to find the other pods / tasks.

patriknw:
I have squashed this and incorporated the "Ensure probe is retried when HTTP request fails" fix by @longshorej 

I assume that the other fix in https://github.com/akka/akka-management/pull/58 for "fix lowest node comparison" was already fixed here because that code had been changed.

fixed some minor formatting thing in the docs also

johanandren:
After the fact LGTM

PR #67, opened 2017-12-06
Document updating the whitesource project name when releasing

raboof:


RELEASING.md
```diff
@@ -1,9 +1,10 @@
 ## Releasing
 
+1. Update the version number in the `akka-management-xx-stable` project name in the [whitesource web UI](https://saas.whitesourcesoftware.com)
```
patriknw: thanks, can you add same thing in the other repos

raboof: I think I covered them all now 👍

Issue #61, opened 2017-11-29
Modularize (esp. Boostrap)


patriknw:
Make things more modular, especially make it easier to compose different Routes to add more things.

Consider to break out the Cluster Sharding part to not force that dependency.

Perhaps also consider making (some) dependencies provided, as suggested by @longshorej in https://github.com/akka/akka-management/pull/60#issuecomment-347931293

ktoso:
Doing this today, along with more docs

PR #69, opened 2017-12-07
Modularize akka-management and bootstrap and discovery

ktoso:
Resolves https://github.com/akka/akka-management/issues/61


bootstrap-joining-demo/src/main/scala/akka/cluster/bootstrap/DemoApp.scala
```diff
@@ -31,7 +29,7 @@ object DemoApp extends App {
 
   log.info(s"Started [$system], cluster.selfAddress = ${cluster.selfAddress}")
 
-  ClusterHttpManagement(cluster).start()
+  AkkaManagement(system).start()
   ClusterBootstrap(system).start()
```
ktoso: the first line actually knows about the routes from bootstrap, and the 2nd line starts the probing and DNSing around here

ktoso: This is done so we only have one "management http server"

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -33,13 +37,28 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
+        log.info("Bootstrap using default `akka.discovery` mechanism: {}", Logging.simpleName(discovery))
+        discovery
+
+      case otherDiscoveryMechanism ⇒
+        val implClazz = system.settings.config.getString(otherDiscoveryMechanism + ".impl")
+        log.info("Bootstrap using [{}] discovery mechanism, instantiating [{}]", otherDiscoveryMechanism, implClazz)
+        system.dynamicAccess
+          .createInstanceFor[SimpleServiceDiscovery](implClazz, List(classOf[ActorSystem] → system))
+          .get
+    }
```
ktoso: this allows for two things.
since discovery is "just a tool", but we're using it here, if someone does not configure anything in bootstrap, but they have picked some discovery impl, THAT discovery impl will be used by default here.
if they want to use DNS here, but something else for "discovery" everywhere else they can configure that too

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -33,13 +37,28 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
 
   private[this] val _selfContactPointUri: Promise[Uri] = Promise()
 
+  override def routes(routeProviderSettings: ManagementRouteProviderSettings): Route = {
+    this.setSelfContactPoint(routeProviderSettings.selfBaseUri)
```
ktoso: This is the reason we do the `ManagementRouteProviderSettings` dance.
We need this in order to know if "we are the lowest address" or not, since we do this decision on the management addresses which are from discovery

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
  */
-package akka.cluster.bootstrap
+package akka.management.cluster.bootstrap
```
ktoso: all tools here are now in management, except the discovery -- that's `akka.discovery`

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpClusterBootstrapRoutes.scala
```diff
@@ -39,17 +39,10 @@ final class HttpClusterBootstrapRoutes(settings: ClusterBootstrapSettings) exten
     }
   }
 
-  // TODO ip whitelist feature?
-  val routes = {
-    // TODO basePath, same as akka-management
-    // val basePath = if (pathPrefixName.isEmpty) rawPathPrefix(pathPrefixName) else pathPrefix(pathPrefixName)
```
ktoso: the base path things are now only in the akka-management, since it's "basepath" for all the APIs, we can add more if needed though, in each of the route providers

project/Common.scala
```diff
@@ -18,7 +18,7 @@ object Common extends AutoPlugin {
 
   override lazy val projectSettings = reformatOnCompileSettings ++
     Dependencies.Common ++ Seq(
-    organization := "com.lightbend.akka",
+    organization := "com.lightbend.akka.management", // FIXME proposing this change, then we can have "com.lightbend.akka.management|discovery" etc; we'd do management.commercial for example later
```
ktoso: WDYT?
Though akka discovery maybe should be not under management?

management/src/main/scala/akka/management/AkkaManagementSettings.scala
```diff
@@ -0,0 +1,38 @@
...
+      require(p >= 0, s"akka.management.http.port MUST be > 0 (was ${p})")
+      p
+    }
+
+    def isHttps = false // FIXME
+
+    val BasePath: Option[String] =
+      Option(cc.getString("base-path")).flatMap(it ⇒ if (it.trim == "") None else Some(it))
+
+    val RouteProviders: List[String] = cc.getStringList("route-providers").asScala.toList
```
ktoso: This mechanism is the same as the "autostarting extensions" in Akka, each lib contributes to this list.
I really think it makes most sense like this since that's just "include lib and got more things in endpoint", and the active things, like the Bootstrap's `start()` are still required to be called by the user.

Same with binding the routes actually -- users have to call that because they may want to configure HTTPS etc

longshorej: I like this -- we can use it to provide, for instance, sane readiness checks on k8s and dcos based on cluster status.

bootstrap-joining-demo/build.sbt
```diff
@@ -4,14 +4,29 @@ enablePlugins(JavaServerAppPackaging)
...
 //  "-Dakka.io.dns.async-dns.resolv-conf=on"
 )
 
+/*
+libraryDependencies += Seq(
+  // 1) we want to use akka-management
+  "akka-management",
+
+  // 2) we want to use the bootstrap
+  cluster-bootstrap
```
patriknw: not prefixed with `akka-`?

bootstrap-joining-demo/src/main/resources/application.conf
```diff
@@ -0,0 +1,40 @@
+akka {
+  actor {
+    provider = "cluster"
+  }
+  remote {
+    enabled-transports = ["akka.remote.netty.tcp"]
```
patriknw: not needed, afaik

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -301,92 +292,72 @@ class ClusterHttpManagementRoutesSpec
...
+//          HttpRequest(uri = s"http://127.0.0.1:20100/shards/ThisShardRegionDoesNotExist")
+//        )
+//        val responseInvalidGetShardDetails = Await.result(responseInvalidGetShardDetailsFuture, 1.second)
+//        responseInvalidGetShardDetails.entity.getContentType shouldEqual ContentTypes.`application/json`
+//        responseInvalidGetShardDetails.status shouldEqual StatusCodes.NotFound
+//
+//        val bindingFuture = clusterHttpManagement.stop()
+//        Await.ready(bindingFuture, 5.seconds)
+//        system.terminate()
+//      }
```
patriknw: fix before merge?

discovery-dns/src/test/resources/reference.conf
```diff
@@ -0,0 +1,11 @@
+akka {
+
+  discovery {
+    impl = akka.discovery.mock-dns
+
+    # this can be refered to in tests to use the mock discovery implementation
+    mock-dns.class = "akka.discovery.MockDiscovery"
+  }
```
patriknw: probably best to inline this into the tests instead, we have had problems with that reference.conf in test/resources are not always first in classpath

discovery/src/test/resources/reference.conf
```diff
@@ -0,0 +1,11 @@
+akka {
+
+  remote {
+    enabled-transports = ["akka.remote.netty.tcp"]
+    netty.tcp {
+      hostname = "127.0.0.1"
+      port = 2552
+    }
+  }
+
```
patriknw: and same here

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
```diff
@@ -67,7 +63,7 @@ private[dns] class HttpContactPointBootstrap(
         s"Uri: $baseUri, this node's remoting address: ${cluster.selfAddress}")
   }
 
-  private implicit val mat = ActorMaterializer()
+  private implicit val mat = ActorMaterializer()(context.system)
```
patriknw: btw, do we check somewhere if `akka.cluster.seed-nodes` config is non-empty? If non-empty I think we should not start joining in the bootstrap extension.

management/src/main/scala/akka/management/AkkaManagementSettings.scala
patriknw: immutable.Seq instead of List

project/Dependencies.scala
```diff
@@ -1,34 +1,89 @@
...
+      DependencyGroups.AkkaSharding ++
+      DependencyGroups.AkkaHttpCore ++
+      DependencyGroups.AkkaTesting ++ Seq(
+      "com.typesafe.akka" %% "akka-distributed-data"              % AkkaVersion     % "test"
+    )
+  )
+
+  val ClusterBootstrap = Seq(
+    libraryDependencies ++=
+      DependencyGroups.AkkaSharding ++
```
longshorej: Can we get away without `AkkaSharding` for `ClusterBootstrap`? Don't we just need `akka-cluster`?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
ktoso: Oh good point, thanks! Will add that and a test

project/Dependencies.scala
ktoso: Good catch, thanks! Yeah only cluster is needed

bootstrap-joining-demo/build.sbt
ktoso: yeah, prefixed `akka-management-cluster-bootstrap` even, this was copied from build.sbt so was project name, fixed here now

ktoso:
Now even with fancy images 😉

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
ktoso: This is done

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
ktoso: Right, done :)

ktoso:
I think this is ready now. Docs can still be improved and more things can be added, but this is a good start.

Please review and merge at will

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpClusterBootstrapRoutes.scala
```diff
@@ -39,17 +36,10 @@ final class HttpClusterBootstrapRoutes(settings: ClusterBootstrapSettings) exten
...
-
+  val routes =
     toStrictEntity(1.second) { // always drain everything
```
johanandren: Isn't this dangerous, given that we will get concatenated with other routes, what if this route goes before others?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -1,31 +1,31 @@
...
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
+import scala.util.Try
 
 /** INTERNAL API */
 @InternalApi
 private[bootstrap] object HeadlessServiceDnsBootstrap {
 
-  def props(discovery: ServiceDiscovery, settings: ClusterBootstrapSettings): Props =
+  def props(discovery: SimpleServiceDiscovery, settings: ClusterBootstrapSettings): Props =
```
johanandren: So much simpler now!

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
```diff
@@ -88,11 +84,12 @@ private[dns] class HttpContactPointBootstrap(
...
   override def receive = {
     case Internal.ProbeNow() ⇒
-      log.debug("Probing {} for seed nodes...", probeRequest.uri)
+      val req = ClusterBootstrapRequests.bootstrapSeedNodes(baseUri)
+      log.debug("Probing {} for seed nodes...", req.uri)
 
       http
         .singleRequest(probeRequest)
-        .flatMap(_.entity.toStrict(settings.contactPoint.probeTimeout))
+        .flatMap(_.entity.toStrict(1.second))
```
johanandren: 1 second sounds tight, and not configurable, is it so little data that it will never be slower?

cluster-http/src/main/resources/reference.conf
```diff
@@ -1,106 +1,18 @@
...
-    # Interval at which contact points should be polled
-    # the effective interval used is this value plus the same value multiplied by the jitter value
-    probe-interval = 1 second
-
-    # Max amount of jitter to be added on retries
-    probe-interval-jitter = 0.2
-
-    # Probe will be failed if it doesn't return in this amount of time
-    probe-timeout = 1 second
+  cluster.http {
```
johanandren: Any point in keeping this block when it's anyways empty?

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -90,22 +89,23 @@ class ClusterHttpManagementRoutesSpec
     }
 
     "join a member" when {
-      "calling POST /members with form field 'memberAddress'" in {
+      "callin" +
+      "g POST /cluster/members with form field 'memberAddress'" in {
```
johanandren: creative line break there!

bootstrap-joining-demo/src/main/resources/application.conf
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+//  # <<<< ru.smslv.akka-dns library >>>>
+//  # need to force that library to be used:
+//  akka.io.dns.resolver = async-dns
+//  # we do want to resovle SRV records if available
+//  akka.io.dns.async-dns.resolve-srv = on
+//  # do use resolv conf -- no reason not to do so really
+//  akka.io.dns.async-dns.resolv-conf = on
+//  # <<<< end of smslv.akka-dns library >>>>
```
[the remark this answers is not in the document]
ktoso: ok will remove for now :)

bootstrap-joining-demo/src/main/resources/application.conf
```diff
@@ -0,0 +1,37 @@
...
+
+}
+
+
+akka.management.http {
+  port = 19999
+}
+
+akka.management.cluster.bootstrap {
```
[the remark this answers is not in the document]
ktoso: ok

build.sbt
```diff
@@ -1,30 +1,84 @@
...
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(
+    name := "akka-management-cluster-bootstrap",
+    Dependencies.ClusterBootstrap
   )
+  .dependsOn(`akka-management`, `akka-discovery`)
 
-lazy val `joining-demo` = project
-  .in(file("joining-demo"))
+// TODO cluster-bootstrap-dns which would just pull together things
```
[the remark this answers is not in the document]
ktoso: I will clarify in there, it's about introducing a module called `akka-management-cluster-bootstrap-dns` which does not do anything, except pull together the 2 modules of `cluster bootstrap` and `akka discovery dns` so it's only 1 dependency you need to pick. I was thinking it would be nice to have "hello world bootstrap in the smallest number of steps" so that would reduce 2 deps into 1.

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -0,0 +1,78 @@
...
+
+  cluster.bootstrap {
+
+    # Configuration for the first phase of bootstraping, during which contact points are discovered
+    # using the configured service discovery mechanism (e.g. DNS records).
+    contact-point-discovery {
+
+      # Define this name to be looked up in service discovery for "neighboring" nodes
+      # If undefined, the name will be extracted from the ActorSystem name
+      service-name = null
```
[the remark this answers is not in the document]
ktoso: I could not find the <extract> you refer to? What would it do?
The setting here is the way of documenting but not setting the value. 
The setting it to null does not actually "set it to null", since `.hasPath` on `service-name` still returns false here. So it's only a way of documenting a value that is not set

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -0,0 +1,78 @@
...
+
+      # Added as suffix to the service-name to build the effective-service name used in the contact-point service lookups
+      # If undefined, nothing will be appended to the service-name.
+      #
+      # Examples, set this to:
+      # "default.svc.cluster.local" or "my-namespace.svc.cluster.local" for kubernetes clusters.
+      service-namespace = null
```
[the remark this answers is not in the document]
ktoso: Yeah empty string here sounds ok

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
ktoso: Here i'll set it to "" though the effective name I do believe should remain `= null`

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
ktoso: changed here, but effective-name should remain with the  `= null style`

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -0,0 +1,78 @@
...
+
+      # The effective service name is the exact string that will be used to perform service discovery
+      # usually, this means the service-name being suffixed with an additional namespace (e.g. "name.default"
```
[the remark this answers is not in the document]
ktoso: Clarified

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -0,0 +1,78 @@
...
+      #
+      # By setting this to `akka.discovery` we ride on the configuration mechanisms that akka-discovery has,
+      # and reuse what is configured for it. You can set it explicitly to something else here, if you want to
+      # use a different discovery mechanism for the bootstrap than for the rest of the application.
+      discovery-method = akka.discovery
+
+      # Amount of time for which a discovery observation must remain "stable"
+      # (i.e. not change list of discovered contact-points) before a join decision can be made.
+      # This is done to decrease the likelyhood of performing decisions on fluctuating observations.
+      stable-margin = 3 seconds
```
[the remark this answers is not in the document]
ktoso: Yah perhaps I'll go with 5 here? In reality it'd seem that nodes are being added "one by one" or at least in batches, so some "changes" should be detected within 5 seconds?

tuning welcome from real world feedback

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -0,0 +1,78 @@
...
+      # Timeout for getting a reply from the service-discovery subsystem
+      resolve-timeout = 3 seconds
+
+    }
+
+    # Configured how we communicate with the contact point once it is discovered
+    contact-point {
+
+      # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
+      fallback-port = 8558 # port pun, it "complements" 2552 which is often used for Akka remoting
```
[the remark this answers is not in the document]
ktoso: 👍

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
ktoso: People can use `ServiceDiscovery().locate("thingy")` just in apps to locate anything they want.
So maybe they do want to do bootstrap using DNS, but other discovery they want ZK or whatever, this is to enable that usecase

docs/src/main/paradox/bootstrap.md
```diff
@@ -36,27 +79,47 @@ of them should perform this join. We make this decision by sorting the known add
...
+simply adding new nodes to it wanting them to discover and join that cluster.
+
+First we explain the Initial Bootstrap. Both cases need to discover neighbouring nodes, which they do via Akka Discovery,
+which we recommend (and implement) with DNS, though you may use alternative lookup methods if you want. The bootstrap 
+process will then start probing those nodes Contact Points (which are HTTP endpoints, exposed via Akka Management by the
+Bootstrap Management Extension) for known seeds to join. Since no cluster exists yet, the seed node replies will be 
+empty from all contact points. Once at least `akka.management.cluster.bootstrap.required-contact-point-nr` nodes are
+discovered and `akka.management.cluster.bootstrap.stable-margin` time has passed, the nodes will decide that it is time 
+to confidently (enough) form a new cluster. The node with the "lowest" contact point address will decide to join itself,
+and other nodes are expecting this awaiting a seed-node address to appear in the contact point probing responses.
```
johanandren: I found the previous description in some list form easier to follow than this long block of text.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpClusterBootstrapRoutes.scala
ktoso: Will increase a bit

docs/src/main/paradox/bootstrap.md
```diff
@@ -66,104 +129,123 @@ store would not help with these as well. The race exists regardless of how we ob
...
+## Joining mechanism precedence
+
+As Akka Cluster allows nodes to join to a cluster using a few methods it is the precedence of each method
+is strictly defined and is as follows:
+
+- If configured `akka.cluster.seed-nodes` (in your `application.conf`) are non-empty, those nodes will be joined, and bootstrap will NOT execute even if `start()` is called (a warning will be logged though).
+- If an explicit `cluster.join` or `cluster.joinSeedNodes` is invokes before the bootstrap completes that
+ joining would take precedence over the bootstrap. *This is however **not** recommended to mix multiple
+ joining methods like this. Stick to one joining mechanism to avoid surprising behaviour during cluster
+ formation.*
```
johanandren: Warning duplicated below, maybe stick with just the warning box

docs/src/main/paradox/discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Implementations
+
+TODO: Discussion about DNS vs other key-value stores.
+
+## Akka DNS Discovery
+
+The simplest, and also most natural form of service discovery is to use DNS as the source of truth regarding available 
+services. In the simplest version, we query for a service name -- which each cluster manager, such as Kubernetes, Mesos 
+or others define using their own naming schemes, and expect to get back a list of IPs that are related to this service.
```
johanandren: Two sentences in a row starting with "... simplest". Maybe the second doesn't really need it?

docs/src/main/paradox/discovery.md
```diff
@@ -0,0 +1,128 @@
...
+```
+akka.discovery {
+  impl = akka.discovery.akka-dns
+}
+```
+
+From there on, you can use the generic API that hides the fact which discovery method is being used by calling::
+
+Scala
+```
```
johanandren: A third style of code blocks here

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
ktoso: Yeah I'm open to a better name. The name `ServiceDiscovery` sadly is taken by the akka extension hmm

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
ktoso: Ok will increase

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -0,0 +1,186 @@
...
+
+  private val routeProviders: immutable.Seq[ManagementRouteProvider] = loadRouteProviders()
+
+  /**
+   * Set async authenticator to be used for management routes.
+   *
+   * Must be called BEFORE `start()` to take effect.
+   * Do not call concurrently.
+   */
+  def setAsyncAuthenticator(auth: AsyncAuthenticator[String]): Unit =
```
johanandren: Maybe we should even throw for the case where it's called late, so users won't be running unauthenticated without knowing it?

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -0,0 +1,186 @@
...
+    _connectionContext = context.asInstanceOf[akka.http.scaladsl.ConnectionContext]
+  private[this] var _connectionContext: akka.http.scaladsl.ConnectionContext = Http().defaultServerHttpContext
+
+  private val bindingFuture = new AtomicReference[Future[Http.ServerBinding]]()
+  private val selfUriPromise = Promise[Uri]() // TODO has to keep config as well as the Uri, so we can reject 2nd calls with diff uri
+
+  /**
+   * Start the HTTP management endpoint
+   */
+  // FIXME make it accept config object that would have all the `withHttps`
```
johanandren: Ah, that would indeed be much nicer.

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -0,0 +1,186 @@
...
+          serverBindingPromise.completeWith(serverFutureBinding).future.flatMap { _ =>
+            log.info("Bound Akka Management (HTTP) endpoint to: {}:{}", hostname, port)
+            selfUriPromise.success(selfBaseUri).future
+          }
+
+        case Failure(ex) ⇒
+          log.warning(ex.getMessage)
+          Future.failed(new IllegalArgumentException("Failed to start Akka Management HTTP endpoint.", ex))
+      }
+    } else selfUriPromise.future
```
johanandren: I'd consider throwing if `start()` is called multiple times perfectly reasonable

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -0,0 +1,186 @@
...
+            "No routes configured for akka management! " +
+            "Double check your `akka.management.http.route-providers` config.")
+    }
+  }
+
+  def stop(): Future[Done] =
+    if (bindingFuture.get() == null) {
+      Future.successful(Done)
+    } else {
+      val stopFuture = bindingFuture.get().flatMap(_.unbind()).map(_ => Done)
```
johanandren: Shouldn't this be a cas as well - extra weird that start() can be called multiple times, but stop() could throw an NPE here if called multiple times

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -0,0 +1,186 @@
...
+      bindingFuture.set(null)
+      stopFuture
+    }
+
+  private def loadRouteProviders(): immutable.Seq[ManagementRouteProvider] = {
+    val dynamicAccess = system.dynamicAccess
+
+    // since often the providers are akka extensions, we initialize them here as the ActorSystem would otherwise
+    settings.Http.RouteProviders map { fqcn ⇒
+      log.warning(s"CREATING ${fqcn}")
```
johanandren: Should that really be a warning, sounds relatively positive to me ;)

management/src/main/scala/akka/management/AkkaManagementSettings.scala
```diff
@@ -0,0 +1,38 @@
...
+
+final class AkkaManagementSettings(val config: Config) {
+  private val managementConfig = config.getConfig("akka.management")
+
+  object Http {
+    private val cc = managementConfig.getConfig("http")
+
+    val Hostname: String = {
+      val hostname = cc.getString("hostname")
+      if (hostname == "") InetAddress.getLocalHost.getHostAddress
```
johanandren: Use the style we do in artery instead with `<getHostAddress>` or something like that to trigger this?

project/Common.scala
johanandren: Isn't this going to be problematic with releasing, we'll need permissions for an extra group name etc.?
It's not like it will conflict with another akka-management or akka-discovery modules?

project/Dependencies.scala
johanandren: Still have sharding here

discovery/src/main/resources/reference.conf
```diff
@@ -0,0 +1,10 @@
+######################################################
+# Akka Service Discovery Config                      #
+######################################################
+
+akka.discovery {
+
+  # impl libraries should set this value
+  # impl = ...
```
patriknw: didn't we agree on that we shouldn't do this classpath magic?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
johanandren: I'm ok with it, just found it funny.

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
johanandren: gut feeling is more comfortable with 5, but that is based on no facts

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
johanandren: This is the one I was thinking about: https://github.com/akka/akka/blob/master/akka-remote/src/main/resources/reference.conf#L737

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
patriknw: +1 unscientific vote for 5s

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
ktoso: Ok i see, will do

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
ktoso: changed to the <>

cluster-bootstrap/src/main/resources/reference.conf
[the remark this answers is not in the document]
ktoso: done

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpClusterBootstrapRoutes.scala
ktoso: Ah right, I moved it into the matched route

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
ktoso: I guess it's simple because just one method of string to resolved thingy... Open to better names though, I'm not happy with the Simple prefix hm

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapExistingSeedNodesSpec.scala
```diff
@@ -0,0 +1,154 @@
...
+
+              contact-point {
+                no-seeds-stable-margin = 4 seconds
+              }
+            }
+          }
+        }
+        """.stripMargin).withFallback(ConfigFactory.load())
+
+      ActorSystem("System", config)
```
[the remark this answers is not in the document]
ktoso: yes, sir

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapExistingSeedNodesSpec.scala
```diff
@@ -0,0 +1,154 @@
...
+
+    "since the nodes have seed nodes configured in config, they should join that, rather than perform bootstrap" in {
+      bootstrapA.discovery.getClass should ===(classOf[MockDiscovery])
+
+      bootstrapA.start()
+      bootstrapB.start()
+      bootstrapNein.start()
+
+      val pB = TestProbe()(systemA)
+      pB.awaitAssert {
```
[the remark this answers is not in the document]
ktoso: Thanks! Should be system B indeed

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapExistingSeedNodesSpec.scala
```diff
@@ -0,0 +1,154 @@
...
+        val stateNein = pNein.expectMsgType[CurrentClusterState]
+        stateNein.members.size should ===(1)
+        stateNein.members.map(_.uniqueAddress.address.port) should ===(Set(clusterNein.selfAddress.port))
+      }
+    }
+
+    "terminate all systems" in {
+      try TestKit.shutdownActorSystem(systemA, 3.seconds)
+      finally try TestKit.shutdownActorSystem(systemB, 3.seconds)
+    }
```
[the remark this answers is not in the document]
ktoso: Thanks!

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapIntegrationSpec.scala
```diff
@@ -80,9 +88,21 @@ class ClusterBootstrapDnsHttpIntegrationSpec extends WordSpecLike with Matchers
...
-      ClusterHttpManagement(Cluster(systemC)).start()
+      def start(system: ActorSystem, contactPointPort: Int) = {
+        import system.dispatcher
+        implicit val sys = system
+        implicit val mat = ActorMaterializer()(system)
+
+        val bootstrap = ClusterBootstrap(system)
+        val routes = new HttpClusterBootstrapRoutes(bootstrap.settings).routes
+        bootstrap
+          .setSelfContactPoint(s"http://127.0.0.1:$contactPointPort") // FIXME how can we set this in akka-management
```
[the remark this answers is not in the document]
ktoso: true

cluster-http/src/main/resources/reference.conf
ktoso: Yeah I wanted to keep it since this is the place to configure this module. In case someone contributes etc

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHttpManagementRoutesSpec.scala
ktoso: :D

discovery-dns/src/main/resources/reference.conf
```diff
@@ -0,0 +1,16 @@
...
+akka.discovery {
+
+  impl = akka.discovery.dns.DnsSimpleServiceDiscovery
+
+  akka-dns {
+    class = akka.discovery.AkkaDnsServiceDiscovery
+
+    resolve-interval = 1 second
+
+    resolve-timeout = 2 second
```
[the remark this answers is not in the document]
ktoso: these were not used actually

discovery-dns/src/test/scala/akka/cluster/bootstrap/discovery/DnsDiscoverySpec.scala
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.cluster.bootstrap.discovery
+
+import org.scalatest.{ Matchers, WordSpecLike }
+
+class DnsDiscoverySpec extends WordSpecLike with Matchers {
+
+  "Dns Discovery" should {
+    "work" in pending
```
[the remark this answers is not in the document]
ktoso: Yeah would have to start actual DNS server... :\

discovery/src/main/resources/reference.conf
ktoso: Right, removing this.

discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
```diff
@@ -0,0 +1,59 @@
...
+  private lazy val _simpleImpl = {
+    val i = system.dynamicAccess
+      .createInstanceFor[SimpleServiceDiscovery](_simpleImplClassConf, (classOf[ExtendedActorSystem] → system) :: Nil)
+      .recoverWith {
+        case _ ⇒
+          system.dynamicAccess.createInstanceFor[SimpleServiceDiscovery](_simpleImplClassConf,
+            (classOf[ActorSystem] → system) :: Nil)
+      }
+      .recoverWith {
+        case _ ⇒ system.dynamicAccess.createInstanceFor[SimpleServiceDiscovery](_simpleImplClassConf, Nil)
```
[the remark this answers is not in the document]
ktoso: Changed it a bit

discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
```diff
@@ -0,0 +1,59 @@
...
+          s"Illegal `akka.discovery.impl` value (${_simpleImplClassConf}) or incompatible class! " +
+          "The implementation class MUST extend akka.discovery.SimpleServiceDiscovery and take an ExtendedActorSystem as constructor argument.",
+          i.failed.get)
+    )
+  }
+
+  /**
+   * Expose [[SimpleServiceDiscovery]] as configured in `akka.discovery.impl`
+   *
+   * Could throw an [[IllegalArgumentException]] at first usage if the configured implementation class is illegal.
```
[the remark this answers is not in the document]
ktoso: Ah true

docs/src/main/paradox/akka-management.md
```diff
@@ -0,0 +1,248 @@
...
+    @@@
+
+And in addition to that, include all of the dependencies for the features you'd like to use,
+like `akka-management-bootstrap` etc. Refer to each extensions documentation page to learn about how
+to configure and use it.
+
+# Configuring the management endpoint
+
+The management endpoint has a number of settings which can be set before invoking 
+`AkkaManagement(system).start()`.
```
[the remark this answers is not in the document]
ktoso: Yeah ok

docs/src/main/paradox/akka-management.md
```diff
@@ -0,0 +1,248 @@
...
+@@@ note
+
+HTTPS is not enabled by default as additional configuration from the developer is required This module does not provide security by default. It's the developer's choice to add security to this API.
+
+@@@
+
+The non-secured usage of the module is as follows:
+
+Scala
+:   ```
```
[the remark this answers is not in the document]
ktoso: I think ```

docs/src/main/paradox/bootstrap.md
```diff
@@ -11,9 +11,52 @@ particularity well in environments like Kubernetes or Mesos where DNS records ar
...
+
+For example, you might depend on the DNS discovery and bootstrap extensions:
+
+sbt
+:   @@@vars
+    ```scala
+    libraryDependencies += "com.lightbend.akka" %% "akka-management-cluster-bootstrap" % "$version$"
+    libraryDependencies += "com.lightbend.akka" %% "akka-discovery-dns"                % "$version$"
+    ```
+    @@@
```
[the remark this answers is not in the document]
ktoso: Yeah but it does not handle multiple deps I thought?

docs/src/main/paradox/bootstrap.md
```diff
@@ -11,9 +11,52 @@ particularity well in environments like Kubernetes or Mesos where DNS records ar
...
+      compile group: "com.lightbend.akka", name: "akka-management-cluster-bootstrap_$scala.binaryVersion$", version: "$version$"
+      compile group: "com.lightbend.akka", name: "akka-discovery-dns_$scala.binaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
+
+
+## Akka Cluster Bootstrap Process explained
+
+The Akka Cluster Bootstrap process is composed of two phases. First, a minimum number of Contact Points (by default at least
```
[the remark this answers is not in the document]
ktoso: Now it's 1 even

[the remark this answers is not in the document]
ktoso: Or not, strike that

docs/src/main/paradox/bootstrap.md
ktoso: Yeah I'll add a small version then

docs/src/main/paradox/bootstrap.md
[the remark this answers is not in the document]
johanandren: I see. That explains.

management/src/main/scala/akka/management/AkkaManagement.scala
ktoso: Good point!

management/src/main/scala/akka/management/AkkaManagement.scala
ktoso: good call

management/src/main/scala/akka/management/AkkaManagement.scala
ktoso: Hah right, was my debug logging, removed. I <3 CAPS

management/src/main/scala/akka/management/AkkaManagementSettings.scala
ktoso: ok

project/Common.scala
ktoso: We have permissions to com.lightbend.akka so we can release under the com.lightbend.akka.management as well

ktoso:
b458f71e-36df-4068-9cd1-8c5d703b5342

Issue #74, opened 2017-12-18
Change package of Akka Discovery


ktoso:
I initially intended it to be `com.lightbend.akka % akka-discovery`

yet it's now under management

ktoso:
Notice that docs already assumed that package `    libraryDependencies += "com.lightbend.akka" %% "akka-discovery-dns" % "$version$"
`

PR #75, opened 2017-12-18
move akka-discovery one package up in groupid  #74

ktoso:
Resolves https://github.com/akka/akka-management/issues/74

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,7 @@ lazy val `akka-discovery` = project
   .settings(unidocSettings)
   .settings(
     name := "akka-discovery",
+    organization := "com.lightbend.akka",
     Dependencies.Discovery
   )
 
@@ -31,6 +32,7 @@ lazy val `akka-discovery-dns` = project
   .settings(unidocSettings)
   .settings(
     name := "akka-discovery-dns",
+    organization := "com.lightbend.akka",
     Dependencies.DiscoveryDns
   )
   .dependsOn(`akka-discovery`)


build.sbt
```diff
@@ -21,6 +21,7 @@ lazy val `akka-discovery` = project
   .settings(unidocSettings)
   .settings(
     name := "akka-discovery",
+    organization := "com.lightbend.akka",
```
johanandren: Why only on these two submodules?

raboof: changing it for artifacts that we have already been publishing for a while might cause confusion with old versions not being properly evicted when they're pulled in as a transitive dependency?

ktoso: > Why only on these two submodules?

Because that's "akka discovery" does not have to live at all with management, could be a core akka thing. Always was intended as such and I did not notice the package was set wrong in build.sbt

> changing it for artifacts that we have already been publishing for a while might

They were first released just yesterday, as 0.7, and that package was a mistake to begin with, even docs pointing out the "com.lightbend.akka" package, so I consider this a bug fix, not a change of package to be honest.

raboof: i agree it makes sense to make this change for akka-discovery - doing it as well for the other submodules would be more problematic

johanandren: Very confusing that the same project has submodules publishing to different groups IMO.

ktoso: Should discovery be moved into akka then? Makes its quick evolution during 0.x cycle difficult, thus keeping it here and evolving and perhaps splitting off I guess.

johanandren: Nah, It's not that important I guess.

hseeberger:
But if it is "com.lightbend.akka" then does it have the same version number like "main" Akka?

I think each project (with its submodules which get released under one version number) should have its unique group id. You guys are mixing up things.

patriknw:
Then we have messed up all things since we only use `com.lightbend.akka` and `com.typedsafe.akka` for all things.

hseeberger:
Do you have so many distinct things?
Anyway, as long as renaming is an option, I think considering a more granular and specific group id doesn‘t hurt.

patriknw:
currently 8 things

ktoso:
That's a fun idea Heiko, but not based in reality -- groups are groups, not the artifact, and it's up to anyone to decide how they group things. Akka has certainly never used the scheme you came out with now, neither did many of my other open source libraries. It does not affect anything as uniqueness is still in-fact, and as such just a matter of taste.

hseeberger:
No worries, thanks for the discussion.

PR #78, opened 2017-12-19
config using names instead of classnames actually works now #77

### LINKED ISSUE #77: Discovery loading via config section name did not work in 0.7, missing test
michaelHampel @michaelHampel 15:39
@ktoso Hello - just tried to run(set run) the main class in akka-management/bootstrap-joining-demo, but its fails with: [error] (run-main-0) java.lang.IllegalArgumentException: Illegal akka.discovery.impl value (akka-dns) or incompatible class! The implementation class MUST extend akka.discovery.SimpleServiceDiscovery and take an ExtendedActorSystem as constructor argument.
java.lang.IllegalArgumentException: Illegal akka.discovery.impl value (akka-dns) or incompatible class! The implementation class MUST extend akka.discovery.SimpleServiceDiscovery and take an ExtendedActorSystem as constructor argument.
should this work?

ktoso:
Resolves #77

### DIFF
--- a/discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
+++ b/discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
@@ -38,21 +38,26 @@ final class ServiceDiscovery(implicit system: ExtendedActorSystem) extends Exten
         .createInstanceFor[SimpleServiceDiscovery](clazzName, (classOf[ExtendedActorSystem] → system) :: Nil)
         .recoverWith {
           case _ ⇒
-            dynamic.createInstanceFor[SimpleServiceDiscovery](_simpleImplMethod,
-              (classOf[ActorSystem] → system) :: Nil)
+            dynamic.createInstanceFor[SimpleServiceDiscovery](clazzName, (classOf[ActorSystem] → system) :: Nil)
         }
         .recoverWith {
           case _ ⇒
-            dynamic.createInstanceFor[SimpleServiceDiscovery](_simpleImplMethod, Nil)
+            dynamic.createInstanceFor[SimpleServiceDiscovery](clazzName, Nil)
         }
 
-    val i = create("akka.discovery." + _simpleImplMethod + ".class").recoverWith {
-      case _ ⇒ create(_simpleImplMethod + ".class")
-    }.recoverWith { case _ ⇒ create(_simpleImplMethod) }
+    // format: OFF
+    val i = create {
+      classNameFromConfig("akka.discovery." + _simpleImplMethod + ".class")
+    }.recoverWith {
+      case _ ⇒ create(classNameFromConfig(_simpleImplMethod + ".class"))
+    }.recoverWith {
+      case _ ⇒ create(_simpleImplMethod) // so perhaps, it is a classname?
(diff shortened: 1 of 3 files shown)


discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
```diff
@@ -38,21 +38,26 @@ final class ServiceDiscovery(implicit system: ExtendedActorSystem) extends Exten
...
+      case _ ⇒ create(classNameFromConfig(_simpleImplMethod + ".class"))
+    }.recoverWith {
+      case _ ⇒ create(_simpleImplMethod) // so perhaps, it is a classname?
+    }
+    // format: ON
 
     i.getOrElse(
       throw new IllegalArgumentException(
-          s"Illegal `akka.discovery.impl` value (${_simpleImplMethod}) or incompatible class! " +
+          s"Illegal `akka.discovery.method` value '${_simpleImplMethod}' or incompatible class! " +
```
raboof: Confused me for a sec that you might expect a method (rather than a class) here, but makes sense if documented I guess :)

discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
```diff
@@ -38,21 +38,26 @@ final class ServiceDiscovery(implicit system: ExtendedActorSystem) extends Exten
         .createInstanceFor[SimpleServiceDiscovery](clazzName, (classOf[ExtendedActorSystem] → system) :: Nil)
         .recoverWith {
           case _ ⇒
-            dynamic.createInstanceFor[SimpleServiceDiscovery](_simpleImplMethod,
-              (classOf[ActorSystem] → system) :: Nil)
+            dynamic.createInstanceFor[SimpleServiceDiscovery](clazzName, (classOf[ActorSystem] → system) :: Nil)
```
raboof: 👍

discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
ktoso: Yeah, documented as:

```
  # Users MUST configure this value to set the default discovery mechanism.
  #
  # The value can be an implementation config path name, such as "akka-dns",
  # which would attempt to resolve as `akka.discovery.akka-dns` which is expected
  # to contain a `class` setting. As fallback, the root `akka-dns` setting scope
  # would be used. If none of those contained a `class` setting, then the value is
  # assumed to be a class name, and an attempt is made to intanciate it.
  method = "<method>"
```

2m: I might be too late to the bike-shedding party, but it would be more consistent to name this `provider`, as we have for [`actor.provider`](https://github.com/akka/akka/blob/master/akka-actor/src/main/resources/reference.conf#L98) which can be either a short name or a FQCN.

ktoso: Provider I wanted to keep for "provides the routes", this is a "discovery method". But yeah, let's not change now I think..

Issue #68, opened 2017-12-07
Pull in Kubernetes API based Service Discovery from reactive-lib


ktoso:
It's implemented here https://github.com/lightbend/reactive-lib/pull/27/files#diff-2130427a780c9a4e08d28efe21102c61R59 and could be a project here in akka-management, so it'd be plug-and-play.

This can be done once the modularize https://github.com/akka/akka-management/issues/61 issue is done

ktoso:
cc @longshorej

longshorej:
:+1:

PR #79, opened 2017-12-19
Discovery implementation based on Kubernetes API

longshorej:
This adds a K8s-API based implementation of `SimpleServiceDiscovery`. By using this (instead of the DNS one), the cluster can bootstrap itself before passing its readiness checks.

### DIFF
--- a/discovery-kubernetes-api/src/main/resources/reference.conf
+++ b/discovery-kubernetes-api/src/main/resources/reference.conf
@@ -18,8 +18,8 @@ akka.discovery {
     # Namespace to query for pods
     pod-namespace = "default"
 
-    # selector value to query pod API with. %s will be replaced with the configured effective name, which defaults to
-    # the actor system name
+    # Selector value to query pod API with.
+    # `%s` will be replaced with the configured effective name, which defaults to the actor system name
     pod-label-selector = "app=%s"
 
     # The name of the akka management port
--- a/discovery-kubernetes-api/src/main/resources/reference.conf
+++ b/discovery-kubernetes-api/src/main/resources/reference.conf
@@ -1,0 +1,28 @@
+######################################################
+# Akka Service Discovery Kubernetes API Config       #
+######################################################
+
+akka.discovery {
+  # Set the following in your application.conf if you want to use this discovery mechanism:
+  # impl = kubernetes-api
+
+  kubernetes-api {
+    class = akka.discovery.kubernetes.KubernetesApiSimpleServiceDiscovery
+
+    # API server, cert and token information. Currently these are present on K8s versions: 1.6, 1.7, 1.8, and perhaps more
+    api-ca-path = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+    api-token-path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+    api-service-host-env-name = "KUBERNETES_SERVICE_HOST"
+    api-service-port-env-name = "KUBERNETES_SERVICE_PORT"
+
+    # Namespace to query for pods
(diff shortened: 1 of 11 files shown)


ktoso:
That was quick, awesome! Will review :)

discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -0,0 +1,30 @@
+######################################################
+# Akka Service Discovery DNS Config                  #
+######################################################
+
+akka.discovery {
+
+  # Set the following in your application.conf if you want to use this discovery mechanism:
+  # impl = akka-dns
```
ktoso: `# method = kubernetes-api` AFAIR, this is so people can copy paste this to enable a given impl

discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -0,0 +1,30 @@
+######################################################
+# Akka Service Discovery DNS Config                  #
```
ktoso: Kubernetes API

longshorej: Oops :) this is fixed

discovery-kubernetes-api/src/main/resources/reference.conf
longshorej: Also fixed

docs/src/main/paradox/discovery.md
```diff
@@ -119,6 +119,96 @@ An improved way of DNS discovery are `SRV` records, which are not yet supported
...
+      compile group: "com.lightbend.akka", name: "akka-discovery-kubernetes-api_$scala.binaryVersion$", version: "$version$"
+    }
+    ```
+    @@@
+    
+And configure it to be used as default discovery implementation in your `application.conf`:
+
+```
+akka.discovery {
+  impl = akka.discovery.kubernetes-api
```
ktoso: method = kubernetes-api 

:)

docs/src/main/paradox/discovery.md
```diff
@@ -119,6 +119,96 @@ An improved way of DNS discovery are `SRV` records, which are not yet supported
...
+    metadata:
+      labels:
+        app: example
+    spec:
+      containers:
+      - name: example
+        image: example/image:1.0.0
+        imagePullPolicy: IfNotPresent
+        ports:
+        - name: akka-mgmt-http
```
ktoso: Awesome :)
Maybe also show the remoting port exposed in this listing so people are not confused when they copy this?

ktoso:
Let me know when you think this is ready to be merged. I will try to add integration tests with mini Kube here some day btw

docs/src/main/paradox/discovery.md
longshorej: Yeah, good idea - this is done.

longshorej:
@ktoso this is ready from my perspective, I tested it out with a local app.

discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -0,0 +1,28 @@
+######################################################
+# Akka Service Discovery Kubernetes API Config       #
+######################################################
+
+akka.discovery {
+  # Set the following in your application.conf if you want to use this discovery mechanism:
+  # impl = kubernetes-api
```
jonas: Should this be ‘method’?

ktoso: Yeap, it should I will fix it

ktoso:
Resolves https://github.com/akka/akka-management/issues/68

Issue #71, opened 2017-12-13
EC2 discovery impl


ktoso:
Which can use the EC2 APIs same as https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/Ec2Snitch.java does

easel:
Another implementation to look at is the for elastiscearch. See https://github.com/elastic/elasticsearch/tree/master/plugins/discovery-ec2/src/main/java/org/elasticsearch/discovery/ec2

On ec2, the trick becomes how do you determine which nodes you discover get picked up? Is it members of a load balancer, instances with a given tag, or perhaps members of an ECS service?

My inclination is towards the ECS option, but I haven't really kicked the tires on making it work yet.

ktoso:
That's part of the problem exactly ;-)
A simple one being "everyone on the same sub-net" perhaps?

easel:
I think I've got it picking up by the "cluster" and "service" on ECS.

Another issue on AWS is the aws sdk for java. It's a pretty heavy dependency and pulls in a bunch of fiddly stuff like jackson and apache http. Are we ok with such a heavy dependency?

sebastianharko:
@ktoso @easel I made a POC (PR open) implementation of this that is plain-old-EC2 and uses "instances with a given tag" mechanism. Not sure if there's any interest to merge here but who knows..

easel:
@sebastianharko I'll take a look, definitely interested.

sebastianharko:
@easel thanks! https://github.com/akka/akka-management/pull/81

ktoso:
Initial impl is in. Not closing ticket for now, there's more to come.
Unless you could open tickets about what specific part you're working on @easel ?

easel:
I can open separate ticket for it, probably easier to keep track of that way.

ktoso:
Great, thanks :)

PR #81, opened 2018-01-04
POC: EC2 discovery impl. #71

sebastianharko:
A working(!) proof-of-concept for EC2 discovery (just plain old EC2, not AWS container service or Kubernetes on AWS)...

This is based on EC2 **tags**. It simply discovers all running instances matching a tag: service -> name.

There are other ways to do discovery on EC2 though. I think this is the easiest though. Let me know what you think :-) 

Cheers!


ktoso:
Noice!

> There are other ways to do discovery on EC2 though. I think this is the easiest though. Let me know what you think :-)

Yeah, I think we may end up providing multiple and that's fine... This one sounds pretty nice!

build.sbt
```diff
@@ -10,6 +10,7 @@ lazy val `akka-management-root` = project
     `akka-discovery`,
     `akka-discovery-dns`,
     `akka-discovery-kubernetes-api`,
+    `akka-discovery-aws`,
```
ktoso: Should it be called EC2 specifically? Since they also have ECS and other kubernetes specific ones etc

discovery-aws/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
...
+
+  aws-ec2-tag-based {
+
+    class = akka.discovery.aws.Ec2TagBasedSimpleServiceDiscovery
+
+    tag-key = "service"
+
+  }
+
+}
```
ktoso: important to have the newline here or things don't get combined properly :)

discovery-aws/src/main/scala/akka/discovery/aws/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.discovery.aws
```
ktoso: `...aws.ec2`?

build.sbt
sebastianharko: your'e right - aws-ec2 sounds like a better choice, since it's more specific ;-)

sebastianharko:
@ktoso yes, working on some basic docs

build.sbt
```diff
@@ -10,6 +10,7 @@ lazy val `akka-management-root` = project
     `akka-discovery`,
     `akka-discovery-dns`,
     `akka-discovery-kubernetes-api`,
+    `akka-discovery-aws-ec2`,
```
ktoso: nice

discovery-aws-ec2/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
+######################################################
+# Akka Service Discovery AWS EC2 Config              #
+######################################################
+
+akka.discovery {
+  # Set the following in your application.conf if you want to use this discovery mechanism:
+  # method = aws-ec2-tag-based
```
ktoso: good

discovery-aws-ec2/src/main/resources/reference.conf
```diff
@@ -0,0 +1,17 @@
...
+
+  aws-ec2-tag-based {
```
ktoso: good :)

docs/src/main/paradox/discovery.md
```diff
@@ -176,6 +176,43 @@ spec:
           protocol: TCP
 ```
 
+## Discovery Method: AWS EC2 Discovery
+
+If you're an AWS user, you can use tags to simply mark the instances that belong to the same cluster. Use a tag that
+has "service" as the key and set the value to be the name of your service. Note that this implementation is adequate
+for users running service clusters on *vanilla* EC2 instances. If you're using Amazon EKS or Amazon ECS, this is not
+the right choice.
```
ktoso: "and you may instead want to use the kubernetes discovery method"?

docs/src/main/paradox/discovery.md
```diff
@@ -176,6 +176,43 @@ spec:
           protocol: TCP
 ```
 
+## Discovery Method: AWS EC2 Discovery
+
+If you're an AWS user, you can use tags to simply mark the instances that belong to the same cluster. Use a tag that
+has "service" as the key and set the value to be the name of your service. Note that this implementation is adequate
```
ktoso: Could it be useful to show how to add such tag using the CLI or the UI a screenshot?

sebastianharko: done :-)

build.sbt
easel: Would it perhaps be best to name the sub-module `discovery-aws` give there are likely to be other implementations that would likely share much of the same config and dependencies?

ktoso: Will they? All the EC2 ones will you mean ;-)
The container engine ones etc would use kubernetes things, wouldn't they?

easel:
This looks like a great start to me, tags meet a lot of use cases and will also work for ECS / EKS if you're using host-based networking (which, tbh, for akka you probably ought to be). Left one comment about the name, but I'm going to try it out. LGTM

build.sbt
easel: I'd expect the existing kube api stuff to work against EKS and live there. 

Amazon specific stuff will all need the AWS API for java and it's transitive dependencies. For instance, ECS, route53 or load balancer integrated implementations.

ktoso: Yeah, but that's most likely for EC2 apps anyway? Or would one do those on EKS as well?

easel: Perhaps a compromise would be to call it `akka-discovery-aws-api`? Make it clear this is stuff that uses the api but doesn't necessarily preclude other implementations on aws that don't use the aws api for java?

easel: I mention this because I have the beginnings of an ECS implementation and will need to know where to put it!

ktoso: I see, the `aws-api` proposal is nice indeed! Let's go with that :)

sebastianharko: @easel @ktoso akka-discovery-aws-api sounds good, renaming right now :-) cheers!

docs/src/main/paradox/discovery.md
```diff
@@ -176,11 +176,55 @@ spec:
...
+for users running service clusters on *vanilla* EC2 instances. If you're using Amazon EKS (Amazon Elastic Container 
+Service for Kubernetes) or plain Amazon ECS, then this is not the right choice and you may want to use 
+the @ref:['Kubernetes API'-based discovery method](discovery.md#discovery-method-kubernetes-api) or some other solution.
+
```
easel: For reference, auto-scaling groups can apply a tag to every new instance, which should handle propagation nicely. Not sure if it's worth calling all that out here or not.

sebastianharko: @easel I think it's worth calling out since auto-scaling is something akka users generally want :-) You want to PR ? cheers!

build.sbt
sebastianharko: @ktoso @easel 
For the Java package name are we ok with:
`akka.discovery.aws_api` as the main name space ? Java package names can't have "-" but can have "_" :-) 

Then, EC2-based methods can be under `akka.discovery.aws_api.ec2` whereas methods based on something like ECS can be under `akka.discovery.aws_api.ecs` ?? 

@easel What do you think ? 
Cheers!

ktoso:
Fixed formatting issues https://travis-ci.org/akka/akka-management/jobs/325110442 which broke the build (formatting is automatic, on compilation, so make sure to compile before PRing)

once it passes will merge

build.sbt
ktoso: That'd be a bit weird, just `awsapi` :)

docs/src/main/paradox/discovery.md
easel: Ok, will do once this lands, there's no good way to pr against an open pr is there?

ktoso: It's possible actually, one has to go to the persons repo and target that branch that is this PR. But anwyay, this will shortly be in :)

sebastianharko: @ktoso i think the build server says ok now, so this can be merged ? cheers!

ktoso:
Yup!

ktoso:
Refs #71

docs/src/main/paradox/discovery.md
sebastianharko: @easel this is merged now so feel free to PR

PR #90, opened 2018-01-08
Discovery: EC2 Tag Based - now can support additional filters

sebastianharko:
This allows the user to add additional filters to the tag based discovery. For example, you usually may want to subset by region or some tag that indicates the environment (test, production etc.). This is optional though.

### DIFF
--- a/docs/src/main/paradox/discovery.md
+++ b/docs/src/main/paradox/discovery.md
@@ -187,7 +187,7 @@ Screenshot of two tagged EC2 instances:
 
 ![EC2 instances](images/discovery-aws-ec2-tagged-instances.png)
 
-Note the tag **service** -> *products-api*. It's set on both instances. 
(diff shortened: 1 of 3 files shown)


docs/src/main/paradox/discovery.md
```diff
@@ -180,17 +180,17 @@ spec:
...
-defined in `application.conf`.
+has "service" as the key and set the value equal to the name of your service (same value as `akka.cluster.bootstrap.contact-point-discovery.service-name` 
+defined in `application.conf`, if you're using this module for bootstrapping your Akka cluster).
  
 Screenshot of two tagged EC2 instances:
 
 ![EC2 instances](images/discovery-aws-ec2-tagged-instances.png)
 
-Note the tag **service** -> *products-api*. 
+Note the tag **service** -> *products-api*. It's set on both instances.
```
ktoso: we try to avoid short forms in the docs => `it is` please :)

PR #96, opened 2018-01-11
Add setting to prevent a new cluster from being formed

### LINKED ISSUE #95: Feature to never form cluster; only join existing
For some workflows, it would be useful to be able to specify that Akka management should never form a cluster and only join an already formed one. This can sort of be emulated by setting contact points to a very high value, but then the logging isn't very nice.

I'd be happy to PR this if we agree this is useful.

longshorej:
This a setting, `allow-formation` which when disabled will prevent a new cluster from being formed, i.e. it will only attempt to join others.

Fixes #95

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -8,6 +8,10 @@ akka.management {
   http.route-providers += "akka.management.cluster.bootstrap.ClusterBootstrap$"
 
   cluster.bootstrap {
+    # When enabled, a new cluster will be formed via the bootstrapping mechanism. This can be disabled to only
+    # allow a node to join already existing clusters.
+
+    allow-formation = true
 
     # Configuration for the first phase of bootstraping, during which contact points are discovered
     # using the configured service discovery mechanism (e.g. DNS records).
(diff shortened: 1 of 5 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HttpContactPointBootstrap.scala
```diff
@@ -113,17 +113,10 @@ private[dns] class HttpContactPointBootstrap(
...
-        //
-        // Summing up, one of the following will happen:
-        // A) this node is allowed to join itself, and does so, and stops this probing actor -- our job is done.
-        // B) some other node triggers the same process and joins itself
-        //    - in which case we'll notice seed-nodes in our probing sooner or later!
-        scheduleNextContactPointProbing()
+        // we have seed nodes, so we notify the parent. Our job is then done -- the parent
+        // will join the cluster
+
+        notifyParentSeedNodesWereFoundWithinDeadline(response)
```
longshorej: Comment and method name were outdated/inaccurate

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -8,6 +8,10 @@ akka.management {
   http.route-providers += "akka.management.cluster.bootstrap.ClusterBootstrap$"
 
   cluster.bootstrap {
+    # When enabled, a new cluster will be formed via the bootstrapping mechanism. This can be disabled to only
+    # allow a node to join already existing clusters.
+
+    allow-formation = true
```
ktoso: Maybe `form-new-cluster = true`?

longshorej:
Thanks @ktoso - I addressed your review, let me know what you think.

PR #101, opened 2018-01-12
Docs: fix issue with dependency

### LINKED ISSUE #94: Wrong dependency in docs
```
"com.lightbend.akka.management" % "akka-management" % "0.8.0"
```

It should be with `%%`, and the maven section doesn't include `_2.12`

sebastianharko:
Fixes #94 

Also fixes a broken link in the README page.

Cheers!

### DIFF
--- a/docs/src/main/paradox/akka-management.md
+++ b/docs/src/main/paradox/akka-management.md
@@ -40,7 +40,7 @@ actually want to use (and load) in their project.
 
 @@dependency[sbt,Gradle,Maven] {
   group="com.lightbend.akka.management"
-  artifact="akka-management"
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/akka-management.md
```diff
@@ -40,7 +40,7 @@ actually want to use (and load) in their project.
 
 @@dependency[sbt,Gradle,Maven] {
   group="com.lightbend.akka.management"
-  artifact="akka-management"
+  artifact="akka-management_$scala.binaryVersion$"
```
ktoso: Though a bit weird, I though this just works hm... perhaps it only does in Akka, I will have a look later on why that is.

PR #98, opened 2018-01-11
Discovery mechanism for Marathon API

longshorej:
This adds another discovery mechanism, this time for Marathon. This allows you to discover other nodes for the bootstrapping process while also defining health checks. It also allows you to take advantage of DC/OS's dynamic port declarations.

I've verified this with an example I'm preparing in another repository. I'll try to clean it up and submit it as an example in this repo via a follow-up PR.

PR #99, opened 2018-01-11
Mesos bridge mode support

ygree:
In order to support network bridge mode on Mesosphere, had to introduce bind-hostname to be able bind to interface within container, but still be able to address it externally and compare it with the list of seed nodes provided by bootstrap.

It solves the issue described in https://github.com/akka/akka-management/issues/64

PR #104, opened 2018-01-16
Add example for using Marathon API

longshorej:
Adds an example for #98. Example won't work until a release is done with that PR in (version used is 0.8.1 below, can change if released with different version)

### DIFF
--- a/bootstrap-joining-demo/marathon-api/marathon/app.json
+++ b/bootstrap-joining-demo/marathon-api/marathon/app.json
@@ -1,0 +1,34 @@
+{
+  "id": "/bootstrap-joining-demo-marathon-api-0.1.0",
+  "cmd": "./bootstrap-joining-demo-marathon-api-0.1.0/bin/bootstrap-joining-demo-marathon-api -java-home $(echo $(pwd)/jre*) -Dakka.remote.netty.tcp.hostname=$HOST -Dakka.remote.netty.tcp.port=$PORT_AKKAREMOTE -Dakka.management.http.hostname=$HOST -Dakka.management.http.port=$PORT_AKKAMGMTHTTP",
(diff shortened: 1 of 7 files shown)


longshorej:
cc @ygree may be useful for the customer in addition to your example

bootstrap-joining-demo/marathon-api/src/main/resources/application.conf
```diff
@@ -0,0 +1,39 @@
+akka {
+
+  discovery.method = aws-api-ec2-tag-based
```
ygree: @longshorej Looks like it should be `marathon-api`

bootstrap-joining-demo/marathon-api/src/main/resources/application.conf
```diff
@@ -0,0 +1,39 @@
...
+
+  actor {
+    provider = "cluster"
+  }
+
+  remote {
+    netty.tcp {
+      port = 2551
```
ygree: @longshorej Looks like these port settings can be omitted here, because they get overridden in the start command defined in `app.json`

bootstrap-joining-demo/marathon-api/marathon/app.json
```diff
@@ -0,0 +1,34 @@
+{
```
ygree: @longshorej Just a note. This example demonstrates HOST mode. We could add bridge-mode example later, when my PR is merged: https://github.com/akka/akka-management/pull/99

ktoso: Could you PR that and update the readme here or else where discussing when what mode should be used? I'm not sure what those modes imply

PR #111, opened 2018-01-19
Fix dependency to support host mode.

ygree:
Add a note about necessity to have fixed http-mgmt-port because DNS SRV lookup is not supported at the moment.
Organize marathon app description files.


ygree:
@ktoso Can we merge this minor fix?

ktoso:
FYI I've created and added you all who contributed here (@longshorej @longshorej @sebastianharko) to the "akka management team" so you have write access here: https://github.com/orgs/akka/teams/akka-management-team/members

Feel free to move fast and merge things.
Let's try to focus on adding tests as we go as well.

bootstrap-joining-demo/dns-api-mesos/README.md
```diff
@@ -13,4 +13,14 @@ Push image into DockerHub
 Mesosphere DC/OS
 ================
 
-Use `mesos.host-mode.json` service description.
\ No newline at end of file
+Host mode
+---------
+Use `marathon/app.host-mode.json` service description.
```
ktoso: What do those modes mean, could you add a short explanation? Like 1 sentence or two..?

bootstrap-joining-demo/dns-api-mesos/build.sbt
```diff
@@ -5,7 +5,7 @@ version := "1.0"
 
 scalaVersion := "2.12.4"
 
-val akkaManagementVersion = "0.8.0"
+val akkaManagementVersion = "0.9.0"
```
sebastianharko: @ygree I have the same issue in my AWS EC2 demo since it's an independent project: every time a new version of akka management gets published , we basically have to update our demos :-( I propose we revert to a solution where our demos are actual sbt (sub)projects of the main project (the kubernetes api is set that way). The disadvantage is that it's harder for prospective users to copy-and-paste from the build.sbt file though .... @ygree what do you think ?

raboof: Hmm, perhaps something like https://github.com/akka/akka-management/pull/120 could be a nice solution (rough draft)?

ktoso:
Not merging this, still really want to have some docs about it... See https://github.com/akka/akka-management/pull/111/files#r163432932

ktoso:
Chance to get some docs here @ygree ?

Issue #114, opened 2018-01-22
Insufficient contact points discovered


jedossa:
When I run `boostrap-joininig-demo-kubernetes-api`, I got not contact points (k8s `v1.8`): `[akka.tcp://Appka@172.17.0.5:2552/system/headlessServiceDnsBootstrap] Discovered [0] observation, which is less than the required [2], retrying (interval: 1.000 s)`.

longshorej:
I'm not sure that example was ever finished. I'll look into this tomorrow and verify if it works / fix it if it doesn't.

Also relevant are the integration tests discussion that's currently WIP, but perhaps we should ensure all examples are tested nightly or something...

PR #115, opened 2018-01-22
Change port configuration for contact point discovery #114

jedossa:
Change management port according to deployment configuration.

### DIFF
--- a/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
+++ b/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
@@ -26,7 +26,7 @@ spec:
               fieldPath: status.podIP
         livenessProbe:
           tcpSocket:
-            port: 8080
+            port: 19999
         ports:
         # akka remoting
         - name: remoting
@@ -38,7 +38,7 @@ spec:
           protocol: TCP
         # external http
         - name: http
-          containerPort: 8080
+          containerPort: 19999
           protocol: TCP
 ---
 kind: Service
--- a/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
+++ b/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
@@ -13,6 +13,7 @@ spec:
     metadata:
       labels:
         app: appka
+        actorSystemName: appka
     spec:
       containers:
       - name: appka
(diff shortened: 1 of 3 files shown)


bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
```diff
@@ -6,8 +6,13 @@ akka {
 
   discovery {
     # pick the discovery method you'd like to use:
-//    method = akka-dns
+    # method = akka-dns
     method = kubernetes-api
+
+    kubernetes-api {
+      pod-label-selector = "app=appka"
```
longshorej: For this example, I think we should add labels for `actorSystemName` and use that. So, `pod-label-selector=actorSystemName=%s`, and add a label to the pod `actorSystemName: appka`.

Also, would you mind removing line 9 (the commented out akka-dns)?

bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
```diff
@@ -6,8 +6,13 @@ akka {
...
+      pod-port-name = "http"
```
longshorej: Rather leave this out -- i.e. the defaults here for consistency with another project Lightbend is working on.

ktoso: you mean to remove the pod-port-name? The latter part of the sentence had even me confused...

longshorej: Sorry for being unclear - Yes, defaults should be sufficient, i.e. no need to declare `pod-port-name`.

ktoso:
I THINK this is correct now... could you check @longshorej ?

bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
```diff
@@ -26,12 +26,11 @@ spec:
               fieldPath: status.podIP
         livenessProbe:
           tcpSocket:
-#            port: 2551
-            port: 19999
+            port: 8080
```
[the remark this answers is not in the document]
jedossa: I have not looked into the reason but if liveness port and management port does not are the same I get: `Unsupported Content-Type, supported: application/json` when probing for seed nodes.

longshorej:
@jedossa sorry I haven't had a chance to look at this in the past few days. I'll make some time tomorrow. Thanks for the improvement!

ktoso:
Thanks!

PR #120, opened 2018-01-25
Auto-detect latest akka-management version in examples

raboof:


bootstrap-joining-demo/dns-api-mesos/build.sbt
```diff
@@ -1,15 +1,13 @@
 enablePlugins(JavaAppPackaging)
 name := "bootstrap-joining-demo-dns-api"
 
-version := "1.0"
-
 scalaVersion := "2.12.4"
 
-val akkaManagementVersion = "0.8.0"
+def akkaManagementVersion(version: String) = version.split('+')(0)
```
sebastianharko: @raboof this has advantages and disadvantages, the disadvantage is that it's not copy-and-paste friendly, so a developer who is browsing github.com (via web) cannnot copy-and-paste but would have to git clone the repo, if I'm not mistaken.

ktoso: Not being copy paste friendly here is fine I'd say.
People should be checking docs and using latest things anyway

sebastianharko: @ktoso true, so let's use this solution then ! :-)

ktoso: Cool, merging

PR #129, opened 2018-02-08
Kubernetes API discovery now ignores pods that are being terminated

### LINKED ISSUE #126: K8s API should honor deletionTimestamp
When an operator requests that a pod be deleted, the K8s REST API includes a non-null value for `deletionTimestamp` in its metadata.

I think we should filter out any such pods from discovery such that we don't attempt to make contact with pods that are in the process of being terminated (i.e. K8s has sent them `SIGTERM` and it's waiting for them to exit)

longshorej:
Fixes #126 

/cc @mitkus


ktoso:
Nice thanks

PR #135, opened 2018-02-15
Make akka.discovery.aws-api-ec2-tag-based.filters optional

lustefaniak:
RIght now if no filters are defined, and default value `""` is used it will fail with `java.lang.AssertionError: assertion failed: failed to parse one of the key-value pairs in filters`


sebastianharko:
Hi @lustefaniak thanks for this PR ! Indeed, it's a bug and observed it too :-) 
Cheers!

Issue #82, opened 2018-01-04
Implement AWS Discovery via ECS Service


easel:

dhpiggott:
I implemented something this weekend that's working fairly nicely. To start with I implemented it directly in my downstream project.

I'm hoping I'll be able to turn it into a contribution to akka-management in the next week or two.

ktoso:
Sounds promising, please do loop back here, it would be an awesome contribution :)

dhpiggott:
I made a start: https://github.com/akka/akka-management/pull/140

PR #140, opened 2018-03-13
ECS Service Discovery

dhpiggott:
1. Implement AWS Discovery via ECS Service API

   The initial implementation was done downstream in Liquidity
(https://github.com/dhpcs/liquidity).

   The first approach was to use the conventions established by the
Ec2TagBasedSimpleServiceDiscovery implementation as a starting point.
The downstream commit on which that part of this contribution is based:
https://github.com/dhpcs/liquidity/commit/0513aacb154ae94b67e2f6ee54a2f2906d75c2c6.

   That approach works, but by limiting ourselves to the old/stable AWS SDK
we are forced to accept blocking IO between our nodes and the AWS APIs.
To overcome this, I enhanced the initial implementation to use the
preview non-blocking AWS SDK.

   The downstream commit on which that part of this contribution is based:
https://github.com/dhpcs/liquidity/commit/9dfbb701fba4e613b2efe4181494284f70a21bea.

   In the downstream project the non-blocking version was developed
immediately after the blocking version - so the blocking version has not
actually been tested in anger. However, as can be seen in the downstream
commit, a main method was created and used to validate it works as a
unit (by doing runMain locally with an appropriate AWS config pointing
it to a real ECS service).

   The non-blocking version was subject to the same test, but is also now
in use running Liquidity (albeit as a one-node-cluster most of the
time...).

   The downstream commit in which the relevant configuration changes were
made:
https://github.com/dhpcs/liquidity/commit/1e376d25f02c4f8c0a17c2e3978660e0d2d87dec.

   For me the non-blocking version is preferable (I'm happy using the
preview non-blocking AWS SDK), but I recognise it might not be desirable
to provide that as the only option in this contribution. Hence
including both, with appropriate namespace separations, so that other
downstream users can choose which they are most comfortable with.

   In this commit I'm providing the implementation (which is the easy part,
because it was already done) in order to get a PR open sooner rather
than later. There's a TODO for documentation - that's the bit which
could take longer - hopefully within the next week or two though.

   Note that the ECS service this was built for and is tested against is
using awsvpc mode networking*
(https://aws.amazon.com/about-aws/whats-new/2017/11/amazon-ecs-introduces-awsvpc-networking-mode-for-containers-to-support-full-networking-capabilities/).
awsvpc mode networking makes things a lot simpler by eliminating the
need for discovering ports too (as would be necessary with bridge mode),
while not also limiting containers to one per container instance (unlike
host mode). In other words, this solution is _not_ intended for use with
older ECS networking approaches; it's only for awsvpc mode.

   *it's also using Fargate, but that shouldn't make a difference one way
or the other here.

2. Create aws-api-ecs bootstrap-joining-demo

   Includes:
   1. CloudFormation template for creation of ECR repository, plus
   code-as-documentation launch script.
   2. Minimal application.
   3. Configuration of sbt-native-packager for generation of Docker image.
   4. Script for publishing local image provided by (3) to ECR repository
   created by (1). We use docker:publishLocal and then tag and push the
   image manually in favour of just using docker:publish because this
   avoids the need to set dockerRepository within the SBT build (which
   would then require that we inject the AWS account ID into the build).
   4. CloudFormation template for creation of ECS service (and dependencies),
   plus code-as-documentation launch script.

   The security group includes a rule to allow public access to 19999 in
order that the formation of the cluster can be observed by running:

   watch -n 1 -c "curl http://$taskInstanceIp:19999/cluster/members/ | python
-mjson.tool"

   where $taskInstanceIp is the public IP of any of the task instances.

   The remaining TODOs are:
   1. Write the README for this addition.

   2. Add to the README in the main docs.

3. Add README for aws-api-ecs bootstrap-joining-demo

4. Update to AWS ECS SDK 2.0.0-preview-9

5. Complete discovery docs for AWS ECS Discovery


discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -45,8 +45,6 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
       .withFilters(runningInstancesFilter)
       .withFilters(otherFilters)
 
-    implicit val timeout = resolveTimeout
```
dhpiggott: Wasn't being used, so removed to fix the compiler warning but also so as to not lead anyone into a false sense of security.

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,136 @@
...
+            .desiredStatus(DesiredStatus.RUNNING)
+            .build()
+        )
+      )
+      accumulatedTasksArns = accumulator ++ listTasksResponse.taskArns().asScala
+      taskArns <- listTasksResponse.nextToken() match {
+        case null =>
+          Future.successful(accumulatedTasksArns)
+
+        case nextPageToken =>
```
dhpiggott: I've only tested with a maximum of two tasks, so this path hasn't been tested (and I don't see an obvious way to do so without spending launching 100+ on my personal account, which I'm not keen on doing) - so it's worth looking at this especially carefully.

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,113 @@
...
+        .withServiceName(serviceName)
+        .withNextToken(pageTaken.orNull)
+        .withDesiredStatus(DesiredStatus.RUNNING)
+    )
+    val accumulatedTasksArns = accumulator ++ listTasksResult.getTaskArns.asScala
+    listTasksResult.getNextToken match {
+      case null =>
+        accumulatedTasksArns
+
+      case nextPageToken =>
```
dhpiggott: Same point here as earlier.

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
dhpiggott: ...thought about it for a few more seconds and realised that, IIRC, the SDK does support a maximum page size. So (_if_ I do recall correctly) it should be simple enough to do a manual test with a smaller number of tasks running to verify this.

ktoso:
Refs  #82

ktoso:
This is awesome, I'll try to make time to review soon; This week is a bit crazy for me with meetings/hackathongs/conference around ScalaMatsuri... Hope to be back here soon, thanks a ton for your work!

sebastianharko:
@dhpiggott would it be possible to move everything related to AWS (async api and sync api) to the akka-discovery-aws-api folder ? I image that once the aws sdk async is no longer in preview, I'm going to migrate the ec2-tag-based discovery to that new aws api anyway ;-)

dhpiggott:
@sebastianharko My hesitation is that I'm not sure it's a good idea to inflict _both_ the old/stable and new/preview SDK dependencies on current/future users of the akka-discovery-aws module. Even if they don't conflict, I know I would prefer that the "flavour" of discovery module I use in my downstream project only brings in the dependencies that it strictly needs - and nothing more.

So my thinking was that the `akka-discovery-aws-api-async` module would serve to give users control over which dependencies they end up inheriting, but that once the async AWS SDK is out of preview its content would be moved into `akka-discovery-aws-api` (at which point  ec2-tag-based discovery would be converted to use the async SDK too). `akka-discovery-aws-api-async` itself could then simply be left empty but made to depend on `akka-discovery-aws-api` so that users of e.g. https://github.com/rtimush/sbt-updates won't miss out on spotting new releases.

sebastianharko:
@dhpiggott Sounds like a plan, thanks for the explanation behind the reasoning ;-)

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
dhpiggott: I've now covered this by (temporarily) adding `.maxResults(1)` to the request and bootstrapping a cluster of 5 task instances.

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
dhpiggott: And here by (temporarily) adding `.withMaxResults(1)` to the request.

bootstrap-joining-demo/aws-api-ec2/README.md
```diff
@@ -10,10 +10,6 @@ Examine the source code
 Examine the source code. See how simple it is: there is no hardcoding
 of seed node IPs or host names in the application.conf file.
 
-Examine the source code
```
dhpiggott: Deleted because it's a duplicate of the paragraph directly above.

dhpiggott:
With a demo app and guide added, the remaining piece (bar responding to review feedback) should just be in adding a section to the main discovery doc here: https://github.com/akka/akka-management/pull/140/files#diff-2f9309809fa0ed116cc955a91dc2c3b0R341. Once I've done that I'll remove the WIP flag.

docs/src/main/paradox/discovery.md
```diff
@@ -184,6 +184,7 @@ then binds the default Service Account to the `Role` by creating a `RoleBinding`
 Adjust as necessary.
 
 ```yaml
```
dhpiggott: Added because IntelliJ's Markdown editor gets very confused without it. Seemed like it might be worth trading a rogue blank line in the final output for a smoother editing experience for contributors, but I'll happily revert it if not.

bootstrap-joining-demo/aws-api-ecs/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,138 @@
...
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable.Seq
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+// This is duplicated from the discovery-aws-api-async (for now...).
+// TODO: Remove when 0.11.0 is released.
+class AsyncEcsSimpleServiceDiscovery(system: ActorSystem)
```
dhpiggott: It would be nice to avoid doing this (even though it is temporary). I'm open to suggestions for alternative approaches!

dhpiggott:
I've updated the main discovery docs now, so this is no longer WIP and should be ready for review.

ktoso:
Thanks, I will dedicate time next week to review and spend time on
akka-management and this PR then :)

docs/src/main/paradox/discovery.md
ktoso: I think it's ok; though I also think all those examples should be eventually made into paradox (and exposed in documentation page), let's do that separately though

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+class AsyncEcsSimpleServiceDiscovery(system: ActorSystem)
+    extends SimpleServiceDiscovery {
+
+  override def lookup(name: String,
+                      resolveTimeout: FiniteDuration): Future[Resolved] = {
+    val ecsClient = ECSAsyncClient.create()
+    implicit val ec: ExecutionContext = system.dispatcher
```
ktoso: val could be a field

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
ktoso: Thanks for calling it out, we can make a separate ticket for testing this

ktoso: https://github.com/akka/akka-management/issues/145

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  // InetAddress.getLocalHost.getHostAddress throws an exception when running
+  // in awsvpc mode because the container name cannot be resolved.
+  // ECS provides a metadata file
+  // (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-metadata.html)
+  // that we ought to be able to use instead to find our IP address, but the
+  // metadata file does not get set when running on Fargate. So this is our
+  // only option for determining what the canonical Akka and akka-management
+  // hostname values should be set to.
+  def getContainerAddress: Either[String, InetAddress] =
```
ktoso: should get some documentation for users why they'd use this -- (i.e. explain how you used it in the demo app for example)

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,113 @@
...
+          Some(nextPageToken),
+          accumulatedTasksArns
+        )
+    }
+  }
+
+  private[this] def describeTasks(ecsClient: AmazonECS, taskArns: Seq[String]): Seq[Task] =
+    for {
+      // Each DescribeTasksRequest can contain at most 100 task ARNs.
+      group <- taskArns.grouped(100).to[Seq]
```
ktoso: should this be configurable? hmm

project/Dependencies.scala
```diff
@@ -81,7 +81,16 @@ object Dependencies {
...
-      DependencyGroups.AkkaActor ++ Seq("com.amazonaws" % "aws-java-sdk-ec2" % "1.11.257")
+      DependencyGroups.AkkaActor ++ Seq(
+        "com.amazonaws" % "aws-java-sdk-ec2" % "1.11.292",
+        "com.amazonaws" % "aws-java-sdk-ecs" % "1.11.292")
+  )
+
+  val DiscoveryAwsApiAsync = Seq(
+    libraryDependencies ++=
+      DependencyGroups.AkkaActor ++ Seq(
+        "software.amazon.awssdk" % "ecs" % "2.0.0-preview-9",
```
ktoso: Ok to use the new one but we should mention that we do so in the docs

bootstrap-joining-demo/aws-api-ecs/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
dhpiggott: I was thinking the same, but then I thought there must be a reason the EC2 bootstrap demo did it this way. Without knowing if that was the case and what the reason was, I thought it better to stick with consistency for this PR.

I'd be happy to unify it all into one build. Most of my opportunity for working on this has been at weekends though and I'm away all of the next weekend, so I'm not certain I'd do it promptly.

bootstrap-joining-demo/aws-api-ecs/build.sbt
```diff
@@ -0,0 +1,18 @@
+scalaVersion := "2.12.4"
+name := "bootstrap-joining-demo-aws-api-ecs"
+version := "1.0"
+
+libraryDependencies ++= Seq(
+  "com.lightbend.akka.management" %% "akka-management-cluster-http" % "0.10.0",
+  "com.lightbend.akka.management" %% "akka-management-cluster-bootstrap" % "0.10.0"
+)
+
+// TODO: To be removed when 0.11.0 is released (see note on AsyncEcsSimpleServiceDiscovery).
```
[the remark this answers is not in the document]
dhpiggott: Yup, that was it. As per the above comment I might do it myself, but if you can that's great too.

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
dhpiggott: Done.

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
dhpiggott: I can't see an obvious reason for it to be.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+**Note:**
+
+The ECS Service Discovery module and this demo are both written solely
+for use with awsvpc mode networking. Use with bridge mode or host mode
+networking is _not_ supported.
+
+**Note:**
+
+The provided CloudFormation application stack configures the task definition
+and service for use with Fargate. This is done in the interest of simplicity
```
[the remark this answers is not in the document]
dhpiggott: Oops! I meant to do this first time. I've added it now (here and in discovery.md too).

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+However, the ECS Service Discovery is _not_ limited to use with Fargate. By
+removing the two Fargate configuration lines and adding a container-instance
+resource you can also run it with "classic" bring-your-own-container-instance
+ECS (and thus in any region).
+
+
+# Step 1: Examine the source code
+
+Examine the source code. See how simple it is. There is no hard coding
+of seed node IPs or host names in the application config.
```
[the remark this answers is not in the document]
dhpiggott: I was thinking the same :-) I was borrowing a bit from the EC2 guide. I've replaced it with your suggestion.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+# Step 2: Create the ECR repository
+
+You can do this manually via the web console by creating an ECR repo called
+`bootstrap-joining-demo-aws-api-ecs`, _or_ by running the included
+CloudFormation wrapper script:
+
+`./scripts/create-infrastucture.sh`
+
```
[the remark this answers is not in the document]
dhpiggott: :+1: Yup, I should have done that first time. I've added a note now.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+# Step 1: Examine the source code
+
+Examine the source code. See how simple it is. There is no hard coding
+of seed node IPs or host names in the application config.
+
+
+# Step 2: Create the ECR repository
+
+You can do this manually via the web console by creating an ECR repo called
```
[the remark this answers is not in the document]
dhpiggott: Done.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+`bootstrap-joining-demo-aws-api-ecs`, _or_ by running the included
+CloudFormation wrapper script:
+
+`./scripts/create-infrastucture.sh`
+
+
+# Step 3: Build and publish the Docker image
+
+The easiest way to do this is by running the included script (which you read
+earlier, right?):
```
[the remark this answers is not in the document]
dhpiggott: Done.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+The script uses `docker:publishLocal` and then tags and pushes the image
+manually (in favour of just using `docker:publish` as might normally be done)
+because this avoids the need to set `dockerRepository` within the SBT build
+(which would then require that we inject the AWS account ID into the build).
+
+
+# Step 4: Identify your subnet IDs
+
+The application stack configures the demo ECS service to launch with five task.
+instances.
```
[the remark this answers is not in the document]
dhpiggott: :+1: Nice catch.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+Before you can create the cluster itself, you need to identify which subnets
+you want the task instances to be launched in.
+
+Go to the VPC service area of the AWS Management Console, then go to the
+Subnets tab. You can compare with the screenshot below as a guide to what
+you're looking for.
+
+![Identifying subnet IDs](screenshots/identify-subnet-ids.png)
+
+For the purposes of the demo we'll assume you're using your default VPC (the
```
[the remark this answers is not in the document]
dhpiggott: :+1: Done. I used to write more formally (and in many cases still do), but in the last couple of years, inspired by [Write Like You Talk](http://www.paulgraham.com/talk.html), I've started deliberately being less formal. Perhaps this is not an appropriate context though, and either way, consistency with the existing conventions trumps individual preferences here. It's only because I'm less tuned in to the conventions that I inadvertently didn't conform :-).

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+one AWS creates for you automatically), though this need not be the case.
+
+Make a note of the IDs of the subnets that your VPC has. You'll need them in
+the following format for CloudFormation:
+
+`subnet-fd6f218a,subnet-2d73e606,subnet-9f2644c6,subnet-2c32b849,subnet-1eaa1512,subnet-a9e5dd93`
+
+(don't use this literal value!)
```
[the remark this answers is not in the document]
dhpiggott: Done.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+# Step 5: Create and validate the cluster
+
+## Create the cluster
+
+Again, the included wrapper script can be used for this (ultimately it
+just delegates to `aws cloudformation create-stack`):
+
+`./scripts/create-application.sh $subnetsIdentifiedInStepFour`
```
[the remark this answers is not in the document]
dhpiggott: :+1: Reworked it to assign to a variable.

bootstrap-joining-demo/aws-api-ecs/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+## Watch it form
+
+The security group that the stack defines and associates with the task
+definition includes a rule to allow public access to port 19999 so that the
+formation of the cluster can be observed as follows:
+
+`watch -n 1 -c "curl http://$taskInstanceIp:19999/cluster/members/ | python
```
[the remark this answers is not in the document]
dhpiggott: :+1: Done.

bootstrap-joining-demo/aws-api-ecs/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
+######################################################
+# Akka Service Discovery AWS Config                  #
```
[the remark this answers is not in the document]
dhpiggott: Once the demo becomes part of the main build and this gets deleted, there will only be `discovery-aws-api/src/main/resources/reference.conf` and `discovery-aws-api-async/src/main/resources/reference.conf`. The former includes config for EC2 and ECS (hence the generalisation of the title), and the latter is expected to also cover both as soon as there's an async EC2 version.

bootstrap-joining-demo/aws-api-ecs/src/main/scala/akka/cluster/bootstrap/DemoApp.scala
```diff
@@ -0,0 +1,50 @@
...
+import akka.management.AkkaManagement
+import akka.management.cluster.bootstrap.ClusterBootstrap
+import com.typesafe.config.ConfigFactory
+
+object DemoApp {
+
+  def main(args: Array[String]): Unit = {
+    val privateAddress = getPrivateAddressOrExit
+    val config = ConfigFactory
+      .systemProperties()
```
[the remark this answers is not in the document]
dhpiggott: `load().withFallback` doesn't work because when the inline config defined here is the fallback, `akka.remote.netty.tcp.hostname` ends up being "" (from the akka-remote `reference.conf`), and `akka.management.http.hostname` ends up being "\<hostname\>" (from the akka-management `reference.conf`).

That results in the respective modules then attempting to call `InetAddress.getLocalHost.getHostAddress` (and failing as described in discovery.md).

So this inline config does need to have precedence over any `ConfigFactory` method which returns something including the reference configs.

 I was going for minimalism* with `ConfigFactory.systemProperties()`, but one working alternative would be `ConfigFactory.parseString("...").withFallback(ConfigFactory.load())`, if that's preferable?

*Which works because I see that `ActorSystem.scala` does this:
```
    final val config: Config = {
      val config = cfg.withFallback(ConfigFactory.defaultReference(classLoader))
      config.checkValid(ConfigFactory.defaultReference(classLoader), "akka")
      config
    }
```

bootstrap-joining-demo/aws-api-ecs/src/main/scala/akka/cluster/bootstrap/DemoApp.scala
```diff
@@ -0,0 +1,50 @@
...
+      )
+    val system = ActorSystem("bootstrap-joining-demo-aws-api-ecs", config)
+    AkkaManagement(system).start()
+    ClusterBootstrap(system).start()
+  }
+
+  private[this] def getPrivateAddressOrExit: InetAddress =
+    AsyncEcsSimpleServiceDiscovery.getContainerAddress match {
+      case Left(error) =>
+        Console.out.println(s"$error. Halting.")
```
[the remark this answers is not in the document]
dhpiggott: Oops. Nice catch - should have been stderr first time!

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
dhpiggott: I included something in the main discovery docs: https://github.com/akka/akka-management/pull/140/files#diff-2f9309809fa0ed116cc955a91dc2c3b0R439

Or do you mean the source comment should be clearer?

project/Dependencies.scala
dhpiggott: I included something in the main discovery docs: https://github.com/akka/akka-management/pull/140/files#diff-2f9309809fa0ed116cc955a91dc2c3b0R374.

Did you have something else in mind?

bootstrap-joining-demo/aws-api-ecs/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
...
+######################################################
+
+// TODO: To be removed when 0.11.0 is released (see note on AsyncEcsSimpleServiceDiscovery).
+akka.discovery {
+  # Set the following in your application.conf if you want to use this discovery mechanism:
+  # method = async-aws-api-ecs
+
+  async-aws-api-ecs {
```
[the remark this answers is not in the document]
dhpiggott: :+1: Done. Not sure why I did it this way! I'd like to think I had a reason but I've no idea what it would have been right now.

dhpiggott:
Thanks for the quick and thorough review! I think I've covered most things now, though there may be a bit more discussion needed on one or two points.

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+class AsyncEcsSimpleServiceDiscovery(system: ActorSystem)
+    extends SimpleServiceDiscovery {
+
+  override def lookup(name: String,
+                      resolveTimeout: FiniteDuration): Future[Resolved] = {
+    val ecsClient = ECSAsyncClient.create()
```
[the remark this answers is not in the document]
dhpiggott: Nope. I shouldn't have ever done it this way really! Have extracted it out as a field now.

bootstrap-joining-demo/aws-api-ecs/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
sebastianharko: @ktoso @dhpiggott you are right - the PR that I have open (that adds integration testing EC2 tag based) makes the EC2 bootstrap demo a subproject of the main one :-)

dhpiggott:
Cleaned up the dependency mess and fixed incorrect Scalafmting (surprised Travis didn't fail it because of that before).

ktoso:
Thanks, this looks great now :-)

ktoso:
We'll merge some other PRs and merge this shortly :)

PR #149, opened 2018-03-27
Kubernetes discovery debugability / parsing improvements

### LINKED ISSUE #147: PodList parsing is too strict, needs higher-level error reporting
The parsing of the response in `KubernetesApiSimpleServiceDiscovery.scala` is too strict. Also,  failures, which prevent cluster formation, are difficult to diagnose.

The newly added error message pointing to RBAC settings can be helpful, but I've seen other reasons for failure.

Changing the type of `PodList.Port.name` to `Option[String]` would eliminate one type of parse failure I've seen.

Another parse error could be caused (and I think I've seen that one too) by `PodList.Container.ports`, which should also be [optional](https://kubernetes-v1-4.github.io/docs/api-reference/v1/definitions/#_v1_container).

That might be all, but we should check that these are the only changes required to comply with the Kubernetes PodList schema.

If after this more lax parsing, our service discovery needs fields that were missing, it should ideally report which field it is missing, and from which pod.

An empty list of returned pods should likely issue a message like "No pods found with label ...".

longshorej:
* Fixes the Kubernetes discovery to be inline with the types defined by the K8s docs, notably several fields made optional
* Renames the local types to what they're actually named in the K8s docs
* Some logging improvements to help debug failures

Fixes #147

### DIFF
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/JsonFormat.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/JsonFormat.scala
@@ -9,11 +9,11 @@ import spray.json._
 import PodList._
 
 object JsonFormat extends SprayJsonSupport with DefaultJsonProtocol {
-  implicit val portFormat: JsonFormat[Port] = jsonFormat2(Port)
+  implicit val containerPortFormat: JsonFormat[ContainerPort] = jsonFormat2(ContainerPort)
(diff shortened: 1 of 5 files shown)


ktoso:
So the worrying thing with all those is that we don't test them except "in someone's prod" nowadays. Which is why it would be so good to get the other testing PRs in, but they need infra as far as I understood

ktoso:
FYI @patriknw

Issue #133, opened 2018-02-13
Make "lowest address" pluggable


longshorej:
*Wanted to get the discussion started on this. I think the only "bullet proof" way to do this is to have the scheduler coordinate termination with the application itself (we've discussed this before), but this feature would at least be an improvement over what we have now and seems like low hanging fruit.*

------------------------

I think the lowest address sort that is done now needs to be pluggable. The scheduler may have access to additional information that akka-management itself doesn't, e.g. which instance was started first (via timestamp metadata). As it is now, the hard-coded behavior of "lowest address" isn't consistent with that, and thus termination strategies such as youngest may not be as effective as they could be, causing singletons to bounce around more than they'd have to.

Alternatively,we could modify the contract of discovery so that it should return the nodes already sorted. Given K8s/Marathon APIs, they would have access to the required information already to do this sort. DNS impl could simply return them sorted by address. This may be conflating concepts a bit, not sure.

Thoughts?

blublinsky:
Did you look at k8 stateful sets. They seem to provide all of the machinery and guarantees. Too bad DC/OS does not have the same thing

longshorej:
Yes, there's a number of issues, the biggest for this particular problem being that use of `StatefulSet` precludes the use of readiness checks.

Even if this wasn't a problem, the deployment strategy for `StatefulSet` isn't ideal for Cluster Singleton. It terminates instances from `n` to `0`. This works nicely for the first rolling upgrade. Given a second rolling upgrade, this will bound the singleton from `n` to `n-1` .. `0` for each step. The third rolling upgrade will be fine, but the fourth will suffer the same problem as the second.

You can choose to terminate from `0` to `n` as well, this suffers from the same problem but is offset by one.

ktoso:
@longshorej that's very interesting info, could you please open a top level ticket about "why not (yet) stateful sets"? People ask this question and my answer was a bit fuzzy "i heard something does not quite match", so being able to direct to such ticket would help a lot (services too).

ktoso:
> As it is now, the hard-coded behavior of "lowest address" isn't consistent with that, and thus termination strategies such as youngest may not be as effective as they could be, causing singletons to bounce around more than they'd have to.

[...]

Issue #144, opened 2018-03-23
Make joining decision pluggable


patriknw:
For cases where only using the lowest address of the discovered node is not desired.

The strategy should be given information from the discovery, and additional information that could be useful, e.g. for how long it has been probing and such.

This would probably also solve https://github.com/akka/akka-management/issues/133 (perhaps even a duplicate).

A decision can also be to continue probing.

ktoso:
Yeah it's likely duplicate but let's keep both in the milestone as they have nice discussion

PR #154, opened 2018-03-29
Make joining decision pluggable, #144

patriknw:
* This centralized decision point also makes the actors easier to understand
* This probably fixes some issues with the old implementation also

Refs #144


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/JoinDecider.scala
```diff
@@ -0,0 +1,208 @@
...
+              "Exceeded stable margins but missing seed node information from some contact points [{}] (out of [{}])",
+              contactPointsWithoutSeedNodesObservations.mkString(", "), info.contactPoints.mkString(", "))
+
+        KeepProbing.asCompletedFuture
+      }
+
+    }
+
+  // FIXME Do we really need settings.contactPoint.noSeedsStableMargin? Isn't the stable-margin enough?
+  // It has been probing all known contact points for that duration
```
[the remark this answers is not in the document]
ktoso: Yes I think you're right; those two can be collapsed into just using stable-margin for the decision point

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -279,16 +243,27 @@ final class HeadlessServiceDnsBootstrap(discovery: SimpleServiceDiscovery, setti
...
+
+      // filter out old observations, in case the probing failures are not triggered
+      def isObsolte(obs: SeedNodesObservation): Boolean =
+        Duration.between(obs.observedAt, currentTime).toMillis > settings.contactPoint.probingFailureTimeout.toMillis
+
+      val seedObservations = seedNodesObservations.valuesIterator.filterNot(isObsolte).toSet
+      val info =
+        new SeedNodesInformation(currentTime, contacts.observedAt, contacts.observedContactPoints, seedObservations)
+
+      joinDecider.decide(info).recover { case _ => KeepProbing }.foreach(self ! _)
```
ktoso: recover is good, but also log a warning then?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -135,130 +103,126 @@ final class HeadlessServiceDnsBootstrap(discovery: SimpleServiceDiscovery, setti
   import HeadlessServiceDnsBootstrap._
   import context.dispatcher
 
+  // FIXME move/rename these classes, doesn't have to be DNS
```
ktoso: yes good point Neither is it about Headless now as well, which is nice I think...? Headless is a specific k8s term, but all this is very generic now :)

ktoso:
Resolves https://github.com/akka/akka-management/issues/133 as well

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -74,7 +74,7 @@ akka.management {
 
       # If some discovered seed node will keep failing to connect for specified period of time,
       # it will initiate rediscovery again instead of keep trying.
-      probing-failure-timeout = 5 seconds
+      probing-failure-timeout = 3 seconds
```
ktoso: I think this needs some cleanup... The request coming back does not have a timeout it seems; yet this timeout is about "if we KEEP FAILING for this period, we'll resolve a new contact point and try again". I think we need to add a probe-timeout = 1.second perhaps and then should be good to have this at 3, so it's be 2/3 attempts after which we resolve again

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
ktoso: I guess no timeout is fine here hm; Technically we could end up invoking it again while it's still internally working on the previous decision future -- could lead to conflicting decisions? I guess we'd have acted upon the first thing already but maybe could show up as confusing logs if the decider decides to log things. 

Some more guard rails around this case would be nice... maybe if only logging that previous future decision still not completed etc

patriknw:
Incorporated some of the feedback, in 56ec7ac
Will fix the other things also.

patriknw:
@ktoso Incorporated remaining feedback and updated documentation. This is ready for final review.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -248,20 +251,34 @@ private[akka] final class HeadlessServiceDnsBootstrap(discovery: SimpleServiceDi
       }
   }
 
-  private def decide(): Unit =
-    lastContactsObservation.foreach { contacts =>
-      val currentTime = timeNow()
+  private def decide(): Unit = {
+    if (decisionInProgress)
+      log.debug("Previous decision still in progress")
```
ktoso: Good though I was thinking warning; but debug is fine I guess

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -248,20 +251,34 @@ private[akka] final class HeadlessServiceDnsBootstrap(discovery: SimpleServiceDi
...
-        new SeedNodesInformation(currentTime, contacts.observedAt, contacts.observedContactPoints, seedObservations)
+        decisionInProgress = true
 
-      joinDecider.decide(info).recover { case _ => KeepProbing }.foreach(self ! _)
+        joinDecider
+          .decide(info)
+          .recover {
+            case e =>
+              log.error(e, "Join decision failed: {}", e)
+              KeepProbing
```
ktoso: good :)

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala
```diff
@@ -248,20 +251,34 @@ private[akka] final class HeadlessServiceDnsBootstrap(discovery: SimpleServiceDi
...
-      val info =
-        new SeedNodesInformation(currentTime, contacts.observedAt, contacts.observedContactPoints, seedObservations)
+        decisionInProgress = true
 
-      joinDecider.decide(info).recover { case _ => KeepProbing }.foreach(self ! _)
+        joinDecider
+          .decide(info)
+          .recover {
+            case e =>
+              log.error(e, "Join decision failed: {}", e)
```
ktoso: e.getMessage) likely? Will change

patriknw:
@ktoso I fixed one thing that I had missed. `form-new-cluster` was lost in the refactoring, fixed in 3a6c9ee

I also moved the `LowestAddressJoinDecider` to a separate file as you suggested previously.

I'll do the other renaming of the classes in a separate PR, don't want to mix that with this diff.

ktoso:
Very nice, thanks!

PR #168, opened 2018-04-09
minor rename of internal classes

patriknw:


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -22,7 +22,7 @@ import akka.event.Logging
 import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.server.Route
 import akka.management.cluster.bootstrap.contactpoint.HttpClusterBootstrapRoutes
-import akka.management.cluster.bootstrap.dns.HeadlessServiceDnsBootstrap
+import akka.management.cluster.bootstrap.internal.BootstrapCoordinator
```
ktoso: yes good :)

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
  */
-package akka.management.cluster.bootstrap.dns
+package akka.management.cluster.bootstrap.internal
```
ktoso: ok

Issue #167, opened 2018-04-06
Little inconsistency when using effective-name and custom pod-label-selector


mbuccini:
Hello everyone,

I was digging into the code to understand how `cluster-bootstrap` is connected to `discovery-kubernetes-api`, and I found out that we have the following:

https://github.com/akka/akka-management/blob/master/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/Settings.scala#L26

```
  def podLabelSelector(name: String): String =
    kubernetesApi.getString("pod-label-selector").format(name)
```

which is based on the assumption that the config file (https://github.com/akka/akka-management/blob/master/discovery-kubernetes-api/src/main/resources/reference.conf) may have a placeholder:
```
akka.discovery {
  kubernetes-api {
    pod-label-selector = "app=%s"
```
If this is the case, it will resolve to `akka.management.cluster.bootstrap.contact-point-discovery.effective-name` (which is built based on `<service-name>.<service-namespace>`, unless it's set differently). This happens because if we use `akka.discovery.method=kubernetes-api`, when we call `discovery.lookup(serviceName, ...)` (here: https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala#L205), it will do the following:

```
  def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] =
    for {
      token <- apiToken()

      labelSelector = settings.podLabelSelector(name)

      _ = system.log.info("Querying for pods with label selector: [{}]", labelSelector)

      request <- optionToFuture(podRequest(token, settings.podNamespace, labelSelector),
        s"Unable to form request; check Kubernetes environment (expecting env vars ${settings.apiServiceHostEnvName}, ${settings.apiServicePortEnvName})")
```
meaning that we may have an inconsistency. 

**Long story short**: if we set  

```akka.discovery {
  kubernetes-api {
     pod-label-selector = "app=my-custom-label"
```

and at the same time we set:

```
akka.management.cluster.bootstrap.contact-point-discovery.effective-name=my-cool-service-name
```

The following line (https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/dns/HeadlessServiceDnsBootstrap.scala#L126):

```override def receive: Receive = {
    case InitiateBootstrapping ⇒
[...]

PR #172, opened 2018-04-10
Fix log inconsistency when using kubernetes-api

markon:
**Problem**
When you use `kubernetes-api` as a discovery mechanism, and the property `pod-label-selector` is different from `effective-name`, you will see in the logs still `effective-name` being used (while it's not the case *only* with kubernetes-api). 

**Solution**
The solution proposed here is to remove `effective-name` from the logged statement. Then, once we retrieve the Resolved points, we can log the service-name used to locate the members. For this reason, we also need to change what `KubernetesApiSimpleServiceDiscovery.lookup` returns, because it doesn't return a Resolved object with `serviceName=pod-label-selector` but  `serviceName=effective-name`. I think this is a good trade-off, because we don't lose the log (which may be useful).

This is a fix for https://github.com/akka/akka-management/issues/167.

Note: I signed the CLA.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
@@ -121,8 +121,8 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
   /** Awaiting initial signal to start the bootstrap process */
   override def receive: Receive = {
     case InitiateBootstrapping ⇒
-      log.info("Locating service members; Lookup [{}]. Using discovery [{}], join decider [{}]", serviceName,
-        discovery.getClass.getName, joinDecider.getClass.getName)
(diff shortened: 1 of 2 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -135,6 +135,7 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
 
     case SimpleServiceDiscovery.Resolved(name, contactPoints) ⇒
       serviceName = name
+      log.info("Located service members with name: [{}].", serviceName)
```
patriknw:
```
log.info("Located service members with name: [{}]: [{}]", serviceName, contactPoints.mkString(", "))
```

?

Later, I think we should revisit the log level of some of the logging. It's rather verbose on info level now and might be better to turn some of it into debug.

markon:
@patriknw I applied the change you mentioned. Could you re-review? :) Thanks!

PR #182, opened 2018-04-21
ECS service discovery improvements

### LINKED ISSUE #180: AsyncEcsSimpleServiceDiscovery -- listTaskArns needs a cluster name
Otherwise, per AWS documentation, the default cluster is assumed, and if wrong, the discovery will fail.

dhpiggott:
1. Make ECS clients lazily initialized
    
   The commit message of
   https://github.com/dhpcs/liquidity/commit/ad946d39c8a941914ce0825ef35e48c2dbecaa37
   contains the full rationale, but in summary it's so that any failures in
   creating the client (typically due to the AWS SDK provider chains not
   finding credentials and/or region values) manifest as exceptions from
   the call to lookup as opposed to the constructor. The reason this is
   preferable is because of the way ServiceDiscovery itself instantiates
   discovery implementations (with chained Try.recoverWith calls but
   without ever logging failures).

2. Fix minor typo in EcsApiDemoApp
    
   The error message that getContainerAddress may produce already includes
   a ".", so appending one here is excessive.

3. Set dockerBaseImage := "openjdk:10-jre-slim" for ECS demo app

4. Automate the identification of subnet IDs in the ECS demo app
    
   This eliminates the only manual step that was required, making it that
   much simpler to try.
    
   Also make the wrapper scripts wait for stack actions to complete and add
   support for updating the stacks via the scripts.

5. Add support for discovery on non-default ECS clusters
    
   Resolves #180.

6. Make ECS demo app resource names shorter, update README and screenshots

7. s/CloudWatchLogGroup/LogGroup/
    
  For consistency with the convention used for other resource type names.

8. Add docs for ECS discovery cluster name settings

9. Read configs to named vals before reading cluster name


bootstrap-joining-demo/aws-api-ecs/cfn-templates/ecs-bootstrap-demo-app-infrastructure.yaml
```diff
@@ -1,6 +1,6 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Resources:
-  ECRRepository:
+  Repository:
```
ktoso: was the previous value invalid?

discovery-aws-api/src/main/resources/reference.conf
```diff
@@ -24,5 +24,7 @@ akka.discovery {
 
     class = akka.discovery.awsapi.ecs.EcsSimpleServiceDiscovery
 
+    cluster = "default"
```
ktoso: please document this key. When to set it, what;'s the use case for changing / having this setting

ktoso: is this value also default in ECS?

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
```diff
@@ -22,9 +22,11 @@ import scala.concurrent.duration._
 
 class EcsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
-  private[this] implicit val ec: ExecutionContext = system.dispatcher
+  private[this] val cluster = system.settings.config.getString("akka.discovery.aws-api-ecs.cluster")
```
ktoso: please first `val c = system.settings.config.getConfig("akka.discovery.aws-api-ecs")` and then get values from that one

discovery-aws-api/src/main/resources/reference.conf
ktoso: It is for https://github.com/akka/akka-management/issues/180 I asume, please document this.
Could you please also explain this in documentation? Users should not be expected to read source details to find this feature

bootstrap-joining-demo/aws-api-ecs/cfn-templates/ecs-bootstrap-demo-app-infrastructure.yaml
dhpiggott: Nope, it was fine. Only reason for the change was for consistency with what's done in most of the application template - e.g. the service resource (https://github.com/akka/akka-management/pull/182/files/dafd6d042811fc43c423825110c337bcad8b5669#diff-f41a72438d909aa2145be9a2a27c3a46R106) is just labeled `Service` (not `ECSService`).

In explaining this I realise I could/should have also changed `CloudWatchLogGroup` (https://github.com/akka/akka-management/pull/182/files/dafd6d042811fc43c423825110c337bcad8b5669#diff-f41a72438d909aa2145be9a2a27c3a46R10) to just `LogGroup`.

dhpiggott: I pushed s/CloudWatchLogGroup/LogGroup/ too for consistency.

discovery-aws-api/src/main/resources/reference.conf
dhpiggott: Added something simple that I think should cover it. To add more would really just amount to duplicating AWS docs which users will need to have read anyway.

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsSimpleServiceDiscovery.scala
dhpiggott: I would definitely have written it as `val config = ...` if two or more values were being read from it. It just seemed a bit redundant as long as there's only one thing being read. I take it that's the preferred style in either case though, so I've made and pushed the change.

docs/src/main/paradox/discovery.md
```diff
@@ -419,6 +423,10 @@ And in your `application.conf`:
 ```
 akka.discovery {
   method = aws-api-ecs-async
+  aws-api-ecs-async {
+    # Defaults to "default" to match the AWS default cluster name if not overridden
+    cluster = "your-ecs-cluster-name"
+  }
```
ktoso: thanks!

ktoso:
Thanks for the follow ups!

Issue #166, opened 2018-04-06
Provide "hardcoded" discovery mechanism, mostly for dev time


ktoso:
As silly as it sounds, it can be pretty useful in real deployments or testing.
It should be able to configure a range of hosts to return, i.e. `2.2.2.2 - 2.2.2.10` as well as CIDR notation which can be useful when someone knows that they deploy only into this subnet and mostly fill it with nodes anyway (not sure how typical, but I can see some people wanting this) `a.b.c.d/24` (likely wrong example, bear with me ;-)).

For bootstrap this can be useful for local development, or simply when people run their own thingies on small scale and know the nodes, and don't have any k8s or anything else available.

patriknw:
How is this different from `seed-nodes` in config?

ktoso:
I think the only difference is the ease of swapping between the two.

`java -Dakka.discovery=mocked` which would use the configuration from application.conf
or without that, which would use the discovery mechanism that application.conf has configured.

With seed nodes you'd have to either:
- do a -D...seed-nodes= (setting arrays is a pain in typesafe config via parameters, since you have to do the `-D...0 = ... -D...1 = ...` syntax if I remember correctly?
- do your own -DdevMode = ... which would be picked up in your app, and load some additional config -- this is essentially what we'd prepackage here.

ktoso:
Note also that discovery is not only used for bootstrap, but can be used in general as a service locator -- so it would be the same "im working local now, use the mocked resolutions" mechanism to mock out IPs to other things or services.

mbuccini:
Good idea, it'd be nice to have a mechanism to allow some local testing that doesn't involve changing completely the setup. 

BTW: I think this is somewhat related to https://github.com/akka/akka-management/issues/162, therefore I am going to close that one and keep it for documentation.

ktoso:
> @markusthoemmes Apr 13 19:58
> Hi, trying to grok akka/akka-management#166 in terms of actionability. Not sure I'm following your train of thought there: What you essentially want is to be able to specify an alternate list of nodes (not seed-nodes) which can be "activated" via akka.discovery.method = mocked?

yeah that's basically it; This is about changing the discovery of the "contact points" how we call them, and from there on the other joining mechanisms would take over

dhpiggott:
I just hacked something together out of necessity that does this in my downstream project. If anyone wants to pinch it, refine it, and then contribute it as their own, please feel free! (I'm not sure I would have time to do so myself anytime soon).

[...]

Issue #157, opened 2018-04-03
"Localhost" (file?) discovery


ktoso:
We could implement a discovery mechanism that works when you're on the same localhost (no containers, etc, just 2 apps started on same box).

The discovery mechanism could memory map a specific "discovery file" and all nodes would monitor it (for discovery).

mikx:
Yes, please add this.

patriknw:
Sounds like an excellent ticket for contributors to pick up.

herzrasen:
Hi, I'm currently working on an implementation for this. Any ideas how to best register the nodes in the local service file?

mikx:
That is what I have now:

```
	akka.discovery.inline {
	  targets = "_:19998,_:19999"
	  class = akka.discovery.config.InlineSimpleServiceDiscovery
	}
```

```
class InlineSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] = {
    val targets = system.settings.config.getString("akka.discovery.inline.targets")
    val host = InetAddress.getLocalHost().getHostAddress()
    val resolved = targets.split(',').to[collection.immutable.Seq] map { _.trim().split(':') } map {
      case Array("_", port) => ResolvedTarget(host, Option(port).map(_.toInt))
      case Array(hst, port) => ResolvedTarget(hst,  Option(port).map(_.toInt))
    }
    Future.successful(Resolved(name, resolved))
  }
}
```

herzrasen:
But doesn't that statically resolve to the targets you have configured? Even if only one instance is running, say 19998 it resolves 19999 too.
I think we need a file where each starting instance is registering in. I'm just not sure about where to implement the registration. Sounds like ClusterBootstrap but putting special "Localhost resolving" code in there feeld odd to me.

PR #188, opened 2018-05-03
Add demo that can be run locally and improve bootstrap docs

chbatey:


bootstrap-joining-demo/local/src/main/resources/second.conf
```diff
@@ -0,0 +1,39 @@
...
+    }
+  }
+
+  discovery {
+    method = akka-dns
+  }
+
+  management {
+    http {
+      port = 19999
```
patriknw: isn't this already taken by `first`?

bootstrap-joining-demo/local/src/main/resources/first.conf
```diff
@@ -0,0 +1,38 @@
...
+    http {
+      port = 19999
+      # akka1 won't work see https://github.com/akka/akka-management/issues/187
+      hostname = "127.0.0.1"
+    }
+
+    cluster.bootstrap {
+      contact-point-discovery {
+        service-name = "akka1"
+        required-contact-point-nr = 1
```
patriknw: why 1 instead of 2?

bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala
```diff
@@ -0,0 +1,19 @@
...
+object Main extends App {
+
+  val system = ActorSystem()
+
+  //#akka-management
+  AkkaManagement(system).start()
+  //#akka-management
+
+  //#bootstrap
+  ClusterBootstrap(system).start()
```
patriknw: Are both AkkaManagement and ClusterBootstrap `start` needed?

can't ClusterBootstrap also start AkkaManagement

then probably make them idempotent so that it can only be started once

bootstrap-joining-demo/local/README.md
```diff
@@ -0,0 +1,23 @@
+# Local bootstrap demo
+
+To run the following need to resolve:
+
+```
+127.0.0.1   akka1
+127.0.0.2   akka2
+127.0.0.3   akka3
+```
```
TimMoore: Isn't the intended use more that you have a single hostname resolve to multiple addresses?

The docs say:

> In the simplest version, we query for a service name – which each cluster manager, such as Kubernetes, Mesos or others define using their own naming schemes, and expect to get back a list of IPs that are related to this service.

and

> The simplest way of resolving multiple hosts of a (micro-)service is to perform a DNS lookup and treat all returned A records as hosts of the same service cluster.

With this configuration, akka1 is a single point of failure.

I know this is just meant as an example, but many users will imitate examples very literally.

docs/src/main/paradox/bootstrap.md
```diff
@@ -62,6 +64,14 @@ akka.discovery.method = akka-dns
...
 can pick the one most fitting to your cluster manager or runtime -- such as Kubernetes, Mesos or other cloud or tool specific methods.
 
+The http routes used for bootstrap are automatically added to Akka management. Your application will need to start Akka management:
+
+@@snip [akka-cluster.yml](../../../../bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala) { #akka-management }
+
+Then to start the boot strap process:
+
+@@snip [akka-cluster.yml](../../../../bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala) { #bootstrap }
+
```
TimMoore: Thanks for adding this! I think a few people have tripped over this already.

bootstrap-joining-demo/local/src/main/resources/second.conf
chbatey: They are listening on different interfaces akka1/2/3 which are 127.0.1/2/3

bootstrap-joining-demo/local/src/main/resources/first.conf
chbatey: akka1 is the seed node so they all discover via this name

bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala
chbatey: You can also add the akka management routes to your own rather than have akka management start a http server

bootstrap-joining-demo/local/README.md
chbatey: Yeah I wanted to do this but it would require running a DNS server locally to do an A record with multiple IPs which defeated my goal of a super simple one to be able to run without deps.

TimMoore: Oh, that's a shame... I didn't realize that `/etc/hosts` doesn't support multiple addresses for the same host.

Maybe it would be more applicable for more people to have an example specifically designed to run in `minikube`.

raboof: or with `docker-compose`? if you have kubernetes we'd want to suggest against using DNS and use the k8s API instead...

chbatey: Im very reluctant to add another demo (we already have 4, one of which is a kubernetes one you could run with minikube) that requires dependencies. I just wanted something you can run without installing things. Perhaps we should just be very clear in the readme about this demos limitation and purpose or not include it.

TimMoore: That's fair enough... I think the nature of cluster bootstrap requires supporting infrastructure though. You would never use this method for a real application unless you also have a properly configured DNS server. I think either of your suggestions (clarify the readme or drop it) would help.

docs/src/main/paradox/bootstrap.md
ktoso: Hmm, quite a waste to close this PR without taking these additions... I'll revive them

ktoso:
> Perhaps we could come up with a bootstrap mechanism that could in fact work on one machine - perhaps using multicast could work?

That's these tickets

- https://github.com/akka/akka-management/issues/166 Provide "hardcoded" discovery mechanism, mostly for dev time 
-  https://github.com/akka/akka-management/issues/157 "Localhost" (file?) discovery

ktoso:
Are we sure we want to just drop it like that? Seemed like a good example and improvement

docs/src/main/paradox/bootstrap.md
```diff
@@ -62,6 +64,14 @@ akka.discovery.method = akka-dns
 Be sure to read about the @ref[alternative discovery methods](discovery.md) that Akka provides, and you
 can pick the one most fitting to your cluster manager or runtime -- such as Kubernetes, Mesos or other cloud or tool specific methods.
 
+The http routes used for bootstrap are automatically added to Akka management. Your application will need to start Akka management:
+
+@@snip [akka-cluster.yml](../../../../bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala) { #akka-management }
+
+Then to start the boot strap process:
+
+@@snip [akka-cluster.yml](../../../../bootstrap-joining-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala) { #bootstrap }
```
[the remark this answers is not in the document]
patriknw: wow, this was completely missing

ktoso:
Reopening as there's good stuff here...

ktoso:
Any chance to revive / make this mergeable? @chbatey

chbatey:
I'll try and do this with the akka dns stuff

bootstrap-joining-demo/local/src/main/resources/second.conf
ktoso: Please use the 8558 port? we moved everything to it in akka-management

chbatey:
I'll update this to use config rather than DNS

chbatey:
I've updated this to use `config` service discovery so can be run from within an IDE with no deps. Ready for re-review @raboof

bootstrap-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object Node3 extends App {
+  new Main(3)
+}
+
+class Main(nr: Int) {
+
+  val config = ConfigFactory.parseString(
+    s"""
+      akka.remote.artery.canonical.hostname = "127.0.0.$nr"
```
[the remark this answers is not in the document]
chbatey: Ah on Mac it appears you have to do:

```
sudo ifconfig lo0 alias 127.0.0.2 up
sudo ifconfig lo0 alias 127.0.0.3 up
sudo ifconfig lo0 alias 127.0.0.4 up
```

I can add that to the instructions or change the port

docs/src/main/paradox/bootstrap/demos/local-config.md
```diff
@@ -0,0 +1,31 @@
...
+
+Akka bootstrap is then configured to lookup `local-cluster` as the service name:
+
+@@snip [application.conf]($management$/bootstrap-demo/local/src/main/resources/application.conf) { #bootstrap }
+
+Three main methods can be run, only overriding the host so the ActorSystem's can all bind to the same port:
+
+@@snip [application.conf]($management$/bootstrap-demo/local/src/main/scala/akka/cluster/bootstrap/Main.scala) { #main }
+
+Run the three mains: `Node1`, `Node2` and `Node3` and they will form a cluster.
```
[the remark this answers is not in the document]
chbatey: Good idea

docs/src/main/paradox/bootstrap/demos/local-config.md
```diff
@@ -22,7 +22,23 @@ Three main methods can be run, only overriding the host so the ActorSystem's can
...
-Run the three mains: `Node1`, `Node2` and `Node3` and they will form a cluster.
+The example uses three loopback addresses: `127.0.0.2-4`. On Mac you'll need to set these up:
+
+```
+sudo ifconfig lo0 alias 127.0.0.2 up
+sudo ifconfig lo0 alias 127.0.0.3 up
+sudo ifconfig lo0 alias 127.0.0.4 up
+```
+
+On Linux this should not be required.
```
raboof: I'll test on Windows :)

chbatey: If that turns out to be an issue i'll just update the ports

raboof: Tested on windows8, works like a charm!

Issue #160, opened 2018-04-04
Management bind port and fallback port of contact points should default to the same 8558


ktoso:
This: https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/resources/reference.conf#L67

should be the same as:
https://github.com/akka/akka-management/blob/master/management/src/main/resources/reference.conf#L17

I propose setting both to 8558

We do want to keep both settings though to allow for extra "wiggle room" in deploys.

ktoso:
Could be a bit nicer, empty value in bootstrap causes reuse of value set in management

PR #192, opened 2018-05-07
Align management bind and contact point

StephenKing:
It is very confusing that the management bind port (19999) and the bootstrap contact point fallback port (8558) do not match.

To improve the new user's experience, use port 19999 everywhere.

Fixes:  #160


ktoso:
I agree, and thanks for the PR however in the opposite direction as mentioned in the ticket :-)
Please make all ports be `8558` since it is the "complement" of the akka port `2552` -- akka-management "complements" akka itself in the same way the port numbers do here.

StephenKing:
okay, that was a bigger search & replace :-)

PR #201, opened 2018-05-16
Disable AmazonEC2Client's retries

### LINKED ISSUE #193: Agressive usage of AWS EC2 API leads to rate limiting
It might be worth to reduce the aggressiveness of the API usage during (unsuccessful) bootstrapping:

> WARN  a.m.c.b.i.BootstrapCoordinator akka.tcp://esc@10.89.16.140:2541/system/bootstrapCoordinator   - Resolve attempt failed! Cause: com.amazonaws.services.ec2.model.AmazonEC2Exception: Request limit exceeded. (Service: AmazonEC2; Status Code: 503; Error Code: RequestLimitExceeded; Request ID: cca286ee-7ff4-4e9b-8dc0-8350904c11b2)

StephenKing:
As our service discovery should throttle down its pace of calling APIs, this moves the control outside about retries outside of the AmazonEC2Client.

Fixes #193

### DIFF
--- a/discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
+++ b/discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
@@ -11,6 +11,8 @@ import akka.discovery.SimpleServiceDiscovery
 import akka.discovery.SimpleServiceDiscovery.{ Resolved, ResolvedTarget }
 import akka.discovery.awsapi.ecs.AsyncEcsSimpleServiceDiscovery._
 import akka.pattern.after
+import software.amazon.awssdk.core.config.ClientOverrideConfiguration
+import software.amazon.awssdk.core.retry.RetryPolicy
 import software.amazon.awssdk.services.ecs._
 import software.amazon.awssdk.services.ecs.model._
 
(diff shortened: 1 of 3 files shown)


discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -34,7 +36,10 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
 
   override def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] = {
 
-    val ec2Client = AmazonEC2ClientBuilder.defaultClient
+    // we have our own retry/backoff mechanism, so we don't need EC2Client's in addition
+    val clientConfiguration = new ClientConfiguration()
+    clientConfiguration.setRetryPolicy(PredefinedRetryPolicies.NO_RETRY_POLICY)
+    val ec2Client = AmazonEC2ClientBuilder.standard().withClientConfiguration(clientConfiguration).build()
```
ktoso: good thanks :) We don't have a backoff yet but yes, let's keep as is here as we want to have a backoff on the outside indeed

StephenKing:
Oh dear, I was talking about _Akka HTTP_, feels a bit weird. I reworded the first commit and also added the same for the ECS discovery.

I'd be happy if somebody could pick up the additional part for backing off (probably in another PR, right?). I'm afk for a bit longer after this week btw., so excuse if I'm unresponsive.

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsSimpleServiceDiscovery.scala
```diff
@@ -25,7 +27,10 @@ class AsyncEcsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceD
   private[this] val config = system.settings.config.getConfig("akka.discovery.aws-api-ecs-async")
   private[this] val cluster = config.getString("cluster")
 
-  private[this] lazy val ecsClient = ECSAsyncClient.create()
+  private[this] lazy val ecsClient = {
+    val conf = ClientOverrideConfiguration.builder().retryPolicy(RetryPolicy.NONE).build()
```
dhpiggott: Is there a reason not to do this in the sync version too (EcsSimpleServiceDiscovery.scala)?

PR #202, opened 2018-05-16
Consul based service discovery 

### LINKED ISSUE #198: Add Consul based service discovery
I would like to add Consul based Service Discovery plugin and since I have a working implementation (although it's very specific and designed for our case) I would like to discuss the case.

The most basic approach would be to base the search on service name (defaulting to `ActorSystem` name in Service Discovery) but if you are using cluster roles then probably you register every node under different service name in consul (i.e. public API might be registered under a name that might be used by Consul Template to generate LB configuration and nodes responsible for processing would be registered under different names). Considering that I would suggest using tags in registered services and base Service Discovery on that. 

Another issue is the fact that Service Discovery requires access to node's Akka Management endpoint and it might not be the exposed under the same port as the registered HTTP check for the service in Consul, so also this should be IMO done on tags.

To sum up, I would like to propose a plugin for Consul Service Discovery that requires the following tags to be present in Consul for the services:
`cluster-port:${AKKA_MANAGEMENT_PORT}` 
`cluster-system-name:some-system`
(of course this could be configurable).

What do you think? Would it be a good thing to add or maybe my approach is wrong?

mlosiewicz-pl:
Fixes #198


mlosiewicz-pl:
Does anyone have any idea how to fix:
```
[error] (akka-discovery-aws-api-async/compile:compileIncremental) scala.reflect.internal.FatalError: Error accessing /home/travis/.ivy2/cache/org.apache.httpcomponents/httpcore/jars/httpcore-4.4.7.jar
```
in travis CI?

raboof:
That seems like some sort of infra problem - I restarted the build.

ktoso:
Thanks a lot, I'll review in depth on monday :-)

ktoso:
Well, 5 days late but reviewing now

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  implicit class ConsulResponseFutureDecorator[T](f: ConsulResponseCallback[T] => Unit) {
+    def asFuture: Future[ConsulResponse[T]] = {
+      val callback = new ConsulResponseFutureCallback[T]
+      f(callback)
+      callback.future
+    }
+  }
+
+  case class ConsulResponseFutureCallback[T]() extends ConsulResponseCallback[T] {
```
ktoso: `final case class` please

discovery-consul/src/test/resources/reference.conf
```diff
@@ -0,0 +1,3 @@
+akka {
+
+}
```
ktoso: trailing \n is needed in reference confs to avoid issues in general, but here perhaps we can just remove it

ktoso: (the file I mean)

discovery-consul/src/test/scala/akka/cluster/bootstrap/discovery/ConsulDiscoverySpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      val lookupService = new ConsulSimpleServiceDiscovery(system)
+      val resolved = lookupService.lookup("test", 10 seconds).futureValue
+      resolved.addresses should contain(ResolvedTarget("127.0.0.1", Some(1234)))
+    }
+  }
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    consul = ConsulStarterBuilder.consulStarter().withHttpPort(8500).build().start()
+  }
```
ktoso: please move the beforeAll above the tests (right next to the var)

discovery-consul/src/test/scala/akka/cluster/bootstrap/discovery/ConsulDiscoverySpec.scala
```diff
@@ -0,0 +1,75 @@
...
+
+      val lookupService = new ConsulSimpleServiceDiscovery(system)
+      val resolved = lookupService.lookup("test", 10 seconds).futureValue
+      resolved.addresses should contain(ResolvedTarget("127.0.0.1", Some(1234)))
+    }
+  }
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    consul = ConsulStarterBuilder.consulStarter().withHttpPort(8500).build().start()
```
ktoso: seems there's no reason to keep it a var, can we make it a val and innit in constructor?

project/Dependencies.scala
```diff
@@ -67,6 +67,16 @@ object Dependencies {
...
   )
 
+  val DiscoveryConsul = Seq(
+    libraryDependencies ++=
+      DependencyGroups.AkkaActor ++
+      DependencyGroups.AkkaTesting ++
+      Seq(
+        "com.orbitz.consul" % "consul-client" % "1.1.2",
+        "com.pszymczyk.consul" % "embedded-consul" % "1.0.2" % "test"
+      )
```
ktoso: Please add a comment with the license next to each of those lines, we have to make sure they are apache compatible (i.e. no GPL code)

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+    val port = catalogService.getServiceTags.asScala
+      .find(_.startsWith(settings.applicationAkkaManagementPortTagPrefix))
+      .map(_.replace(settings.applicationAkkaManagementPortTagPrefix, ""))
+      .flatMap { maybePort =>
+        Try(maybePort.toInt).toOption
+      }
+    ResolvedTarget(catalogService.getServiceAddress, Some(port.getOrElse(catalogService.getServicePort)))
+  }
+
+  private val getServicesWithTags = ((callback: ConsulResponseCallback[util.Map[String, util.List[String]]]) =>
```
ktoso: could this be a normal method instead?

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+       consul.catalogClient().getService(name, QueryOptions.BLANK, callback)).asFuture
+
+}
+
+object ConsulSimpleServiceDiscovery {
+
+  implicit class ConsulResponseFutureDecorator[T](f: ConsulResponseCallback[T] => Unit) {
+    def asFuture: Future[ConsulResponse[T]] = {
+      val callback = new ConsulResponseFutureCallback[T]
+      f(callback)
```
ktoso: could `f` throw? Let's make sure that then the Future is failed rather than the asFuture throwing

discovery-consul/src/main/resources/reference.conf
```diff
@@ -0,0 +1,20 @@
...
+  # impl = akka-consul
+
+  # configure the akka-consul provider
+  akka-consul {
+    class = akka.discovery.consul.ConsulSimpleServiceDiscovery
+
+    consul-host = "127.0.0.1"
+    consul-port = 8500
+
+    application-name-tag-prefix = "system:"
```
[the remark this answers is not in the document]
ktoso: So here we would have to explain that we expect values to be present under "system:name-of-my-actor-system" right?

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
+
+class ConsulSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
+
+  private lazy val settings = ConsulSettings.get(system)
+  private lazy val consul =
+    Consul.builder().withHostAndPort(HostAndPort.fromParts(settings.consulHost, settings.consulPort)).build()
```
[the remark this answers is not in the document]
mlosiewicz-pl: nope,

bootstrap-joining-demo/aws-api-ec2/build.sbt
```diff
@@ -1,6 +1,6 @@
 enablePlugins(JavaAppPackaging)
 
-packageName in Universal := "app" // should produce app.zip
+com.typesafe.sbt.SbtNativePackager.autoImport.packageName in Universal := "app" // should produce app.zip
```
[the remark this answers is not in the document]
mlosiewicz-pl: My sbt was complaining about, but it's probably a local issue.

discovery-consul/src/main/resources/reference.conf
```diff
@@ -0,0 +1,20 @@
+######################################################
+# Akka Service Discovery Consul Config                  #
+######################################################
```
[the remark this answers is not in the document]
mlosiewicz-pl: fixed

discovery-consul/src/main/resources/reference.conf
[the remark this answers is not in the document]
mlosiewicz-pl: Added

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[SimpleServiceDiscovery.Resolved] = {
+    import system.dispatcher
```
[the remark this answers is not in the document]
mlosiewicz-pl: fixed

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,105 @@
...
+    Future.firstCompletedOf(
+      Seq(
+        after(resolveTimeout, using = system.scheduler)(
+          Future.failed(new TimeoutException("Future timed out!"))
```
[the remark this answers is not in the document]
mlosiewicz-pl: Right, fixed

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
mlosiewicz-pl: done

discovery-consul/src/test/resources/reference.conf
mlosiewicz-pl: removed

discovery-consul/src/test/scala/akka/cluster/bootstrap/discovery/ConsulDiscoverySpec.scala
mlosiewicz-pl: it works without `var`

discovery-consul/src/test/scala/akka/cluster/bootstrap/discovery/ConsulDiscoverySpec.scala
mlosiewicz-pl: Done

project/Dependencies.scala
mlosiewicz-pl: Added

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
mlosiewicz-pl: done

discovery-consul/src/main/scala/akka/discovery/consul/ConsulSimpleServiceDiscovery.scala
mlosiewicz-pl: I think I did what you asked for, but please take another look

mlosiewicz-pl:
About the registration method - I assume that most of the people doing this already have something in place like `ConstructR` and have a way of registering services in Consul ( for example if you use Nomad then that part is done for you ) so I would advise against adding a method for registration. The tags assigned to services are not key-value pairs but a set of strings, that's why I need prefixes. Also if we add a method for registration in this case then I would have to make sure that what I register is properly handled by everything set up by consul-template (it is common to set up load balancers based on services in consul) so that there is no duplication. Consul template is also the reason I added the port for akka-management, I don't want it to be accessible through load balancer.

I based the decisions on implementation on an actual production running service and it worked in my case (I mean, I deployed it on prod already ;) to check if it actually work) but in other cases it might not be accurate, but the generic approach with tags should work most of the times.

I will add the discovery.md as soon as I find some spare time and write it all down.

ktoso:
> About the registration method - I assume that most of the people doing this already have something in place like ConstructR and have a way of registering services in Consul ( for example if you use Nomad then that part is done for you ) so I would advise against adding a method for registration.

Note that this project *replaces* constructr, which is already in archived mode even: https://github.com/hseeberger/constructr/issues/174

mlosiewicz-pl:
I meant that most of the setup for existing application is already done somewhere and since ConstructR does not register the service in Consul there has to be another mechanism in place already. Adding a registration method in my opinion is not easy (you have to create health checks and know what is exposed where in application - you probably have all the information anyway but still a generic solution would be hard to implement) and will be omitted most of the time. 

Said that, I agree that this would be a powerful tool and would allow for quick set up of the cluster with low effort and very quickly. While implementing registration in Consul a lot of configuration options would be have to be added, but it's doable.

ktoso:
That's a good summary – I think it's valuable, but we'll see when/if we end up providing it or not.

As for this PR, I would like to merge it but it needs a bit of documentation -- would you be able to add it? Thanks a lot

mlosiewicz-pl:
Sorry, I was quite busy with some other stuff. 

I suck at writing documentation so please check it and if it does not make sense I'll try to write it in some other way.

ktoso:
Thanks! It's a good start :) I'll merge and we'll see what feedback we get. I'll also want to mark it somehow as community contriubted / "not supported" or something similar until we're confident in it... hope that's fine. Exact wording for these things we'll figure out during this week (we got a team meeting about it :))

PR #205, opened 2018-05-22
implement backoff in discovery usage in bootstrap #193

ktoso:
Will resolve https://github.com/akka/akka-management/issues/193 I'd hope.
It makes the timeouts more staggered as well as implements a backoff on failure

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -49,7 +49,25 @@ akka.management {
       stable-margin = 5 seconds
 
       # Interval at which service discovery will be polled in search for new contact-points
+      #
+      # Note that actual timing of lookups will be the following:
+      # - perform initial lookup; interval is this base interval
+      # - await response within resolve-timeout
+      #   (this can be larger than interval, which means interval effectively is resolveTimeout + interval,
+      #    this has been specifically made so, to not hit discovery services with requests while the lookup is being serviced)
+      #   - if failure happens apply backoff to interval (the backoff growth is exponential)
+      # - if no failure happened, and we receive a resolved list of services, schedule another lookup in interval time
+      #   - if previously failures happened during discovery, a successful lookup resets the interval to `interval` again
+      # = repeat until stable-margin is reached
       interval = 1 second
+
+      # Adds "noise" to vary the intervals between retries slightly (0.2 means 20% of base value).
+      # This is important in order to avoid the various nodes performing lookups in the same interval,
+      # potentially causing a thundering heard effect. Usually there is no need to tweak this parameter.
+      exponential-backoff-random-factor = 0.2
+
+      # Maximum interval to which the exponential backoff is allowed to grow
+      exponential-backoff-max = 15 seconds
 
       # The smallest number of contact points that need to be discovered before the bootstrap process can start.
       # For optimal safety during cluster formation, you may want to set these value to the number of initial
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
@@ -4,12 +4,13 @@
 package akka.management.cluster.bootstrap
 
 import java.util.Locale
-import java.util.concurrent.TimeUnit
+import java.util.concurrent.{ ThreadLocalRandom, TimeUnit }
 
 import akka.actor.ActorSystem
 import com.typesafe.config.Config
 
 import scala.concurrent.duration.{ FiniteDuration, _ }
+import scala.util.Try
 
 final class ClusterBootstrapSettings(config: Config) {
   private implicit class HasDefined(val config: Config) {
@@ -59,6 +60,14 @@ final class ClusterBootstrapSettings(config: Config) {
 
     val interval: FiniteDuration =
       discoveryConfig.getDuration("interval", TimeUnit.MILLISECONDS).millis
+
+    val exponentialBackoffRandomFactor: Double =
+      discoveryConfig.getDouble("exponential-backoff-random-factor")
+
+    val exponentialBackoffMax: FiniteDuration =
+      discoveryConfig.getDuration("exponential-backoff-max", TimeUnit.MILLISECONDS).millis
+
+    require(exponentialBackoffMax >= interval, "exponential-backoff-max has to be greater or equal to interval")
 
     val requiredContactPointsNr: Int = discoveryConfig.getInt("required-contact-point-nr")
 
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
@@ -3,11 +3,10 @@
  */
 package akka.management.cluster.bootstrap.internal
 
-import java.time.Duration
 import java.time.LocalDateTime
+import java.util.concurrent.ThreadLocalRandom
 
 import scala.collection.immutable
-
 import akka.actor.Actor
 import akka.actor.ActorLogging
 import akka.actor.ActorRef
@@ -29,7 +28,10 @@ import akka.management.cluster.bootstrap.JoinSelf
 import akka.management.cluster.bootstrap.KeepProbing
 import akka.management.cluster.bootstrap.SeedNodesInformation
 import akka.management.cluster.bootstrap.SeedNodesObservation
-import akka.pattern.pipe
+import akka.pattern.{ pipe, BackoffSupervisor }
+
+import scala.concurrent.duration._
+import scala.util.Try
 
 /** INTERNAL API */
 @InternalApi
@@ -53,7 +55,7 @@ private[akka] object BootstrapCoordinator {
     final case class ProbingFailed(contactPoint: ResolvedTarget, cause: Throwable) extends DeadLetterSuppression
   }
 
-  private case object ResolveTick extends DeadLetterSuppression
+  private case object DiscoverTick extends DeadLetterSuppression
   private case object DecideTick extends DeadLetterSuppression
 
   protected[bootstrap] final case class DnsServiceContactsObservation(observedAt: LocalDateTime,
@@ -87,7 +89,14 @@ private[akka] object BootstrapCoordinator {
  *
  * CAVEATS:
  * There is a slight timing issue, that may theoretically appear in this bootstrap process.
- * FIXME explain the races
+ * One such example is the timing between the lookups becoming an "stable observation",
+ * so an decision is made with regards to joining (e.g. self-joining) on the nodes,
+ * yet exactly right-after the discoverContactPoints returned, a new node with lowest address appears (though is not observed).
+ * Technically this is a race and "wrong decision" made by then, however since the node will itself also probe
+ * the other nodes visible in discovery, it should soon realise that a cluster exists/is-forming and join that
+ * one instead of self-joining, so the race in most timing situations should be harmless, however remains possible
+ * under very unlucky timing -- where the new lowest node would not observe the new cluster being formed within the
+ * stable timeout.
  */
 // also known as the "Baron von Bootstrappen"
 /** INTERNAL API */
@@ -101,11 +110,11 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
 
   import BootstrapCoordinator.Protocol._
   import BootstrapCoordinator._
-  import context.dispatcher
 
+  implicit private val ec = context.dispatcher
   private val cluster = Cluster(context.system)
 
-  private val ResolveTimerKey = "resolve-key"
+  private val DiscoverTimerKey = "resolve-key"
   private val DecideTimerKey = "decide-key"
 
   private var serviceName = settings.contactPointDiscovery.effectiveName(context.system)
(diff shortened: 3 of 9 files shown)


ktoso:
Will see to add a test for this after lunch

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -49,8 +49,24 @@ akka.management {
...
+      #   (this can be larger than interval, which means interval effectively is resolveTimeout + interval,
+      #    this has been specifically made so, to not hit discovery services with requests while the lookup is being serviced)
+      #   - if failure happens apply backoff to interval (grow interval by multiplying by the exponential-backoff-factor)
+      # - if no failure happened, and we receive a resolved list of services, schedule another lookup in interval time
+      #   - if previously failures happened during discovery, a successful lookup resets the interval to `interval` again
+      # = repeat until stable-margin is reached
       interval = 1 second
 
+      # When lookups in discovery fail an exponential backoff will be performed
+      exponential-backoff-factor = 1.2
```
patriknw: I don't think this property is needed. It's too detailed for anyone to care. Using pow 2 with a max cap is enough. That's how we do it in other backoff features, such as `RestartFlow` and `BackoffSupervisor`. The exact calculation is in https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala#L210

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -87,7 +87,14 @@ private[akka] object BootstrapCoordinator {
...
  * There is a slight timing issue, that may theoretically appear in this bootstrap process.
- * FIXME explain the races
+ * One such example is the timing between the lookups becoming an "stable observation",
+ * so an decision is made with regards to joining (e.g. self-joining) on the nodes,
+ * yet exactly right-after the discoverContactPoints returned, a new node with lowest address appears (though is not observed).
+ * Technically this is a race and "wrong decision" made by then, however since the node will itself also probe
+ * the other nodes visible in discovery, it should soon realise that a cluster exists/is-forming and join that
+ * one instead of self-joining, so the race in most timing situations should be harmless, however remains possible
+ * under very unlucky timing -- where the new lowest node would not observe the new cluster being formed within the
+ * stable timeout.
```
patriknw: good explanation

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -60,6 +60,17 @@ final class ClusterBootstrapSettings(config: Config) {
     val interval: FiniteDuration =
       discoveryConfig.getDuration("interval", TimeUnit.MILLISECONDS).millis
 
+    val exponentialBackoffFactor: Double =
+      discoveryConfig.getDouble("exponential-backoff-factor")
+
+    val exponentialBackoffMax: Duration =
```
patriknw: why not `FiniteDuration`?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -60,6 +60,17 @@ final class ClusterBootstrapSettings(config: Config) {
...
+      discoveryConfig.getDuration("exponential-backoff-max", TimeUnit.MILLISECONDS).millis
```
patriknw: require exponential-backoff-max >= interval ?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -114,34 +121,55 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
+
+  private var currentDiscoveryInterval: Duration = settings.contactPointDiscovery.interval
+  def resetDiscoveryInterval(): Unit =
+    currentDiscoveryInterval = settings.contactPointDiscovery.interval
+  def backoffDiscoveryInterval(): Unit =
+    currentDiscoveryInterval = settings.contactPointDiscovery.backoffInterval(currentDiscoveryInterval)
+  def startSingleDiscoveryTimer(): Unit =
+    currentDiscoveryInterval match {
+      case d: FiniteDuration => timers.startSingleTimer(DiscoverTimerKey, DiscoverTick, d)
+      case _ => // ignore, infinite interval
```
patriknw: how can it be non-finite?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -101,11 +108,11 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
 
   import BootstrapCoordinator.Protocol._
   import BootstrapCoordinator._
-  import context.dispatcher
 
+  implicit private val ec = context.dispatcher
```
raboof: changed because of https://github.com/akka/akka/issues/25040 ?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -114,34 +121,55 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
       context become bootstrapping(sender())
   }
 
   /** In process of searching for seed-nodes */
   def bootstrapping(replyTo: ActorRef): Receive = {
-    case ResolveTick ⇒
-      lookup()
+    case DiscoverTick ⇒
+      discoverContactPoints()
+    // the next round of discovery will be performed once this one returns
```
raboof: The (scalariform?) indent suggests that this comment belongs to the next case - perhaps move it up a line?

cluster-bootstrap/src/main/resources/reference.conf
ktoso: Thanks, will do that then :)

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
ktoso: AFAIR one of the calculations returns a Duration so it bubbled up though realistically it should always be Finite ones, I'll change that

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
ktoso: ok

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
ktoso: yeah, as general style recommendation rather than an issue that I encountered here tho

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
ktoso: yeah weird... will check

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
raboof: yeah :+1:

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -49,8 +49,26 @@ akka.management {
       stable-margin = 5 seconds
 
       # Interval at which service discovery will be polled in search for new contact-points
+      #
+      # Note that actual timing of lookups will be the following:
+      # - perform initial lookup; interval is this base interval
+      # - await response within resolve-timeout
+      #   (this can be larger than interval, which means interval effectively is resolveTimeout + interval,
+      #    this has been specifically made so, to not hit discovery services with requests while the lookup is being serviced)
+      #   - if failure happens apply backoff to interval (grow interval by multiplying by the exponential-backoff-factor)
```
patriknw: exponential-backoff-factor is no more

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -192,15 +235,20 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
+          // child actor will have terminated now, so we ride on another discovery round to cause looking up
+          // target nodes and if the same still exists, that would cause probing it again
+          //
+          // we do this in order to not keep probing nodes which simply have been removed from the deployment
         }
       }
 
       // remove the previous observation since it might be obsolete
       seedNodesObservations -= contactPoint
+      startSingleDiscoveryTimer()
```
patriknw: not needed, but I guess it doesn't hurt to wait another interval before discovery here

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+        val lookup = resolveProbe.expectMsgType[DiscoveryRequest]
+        lookup.time
+      }
+      // ends up being very flaky:
+      //      (times(1) - times(0)).toInt shouldBe 2000 +- 500
+      //      (times(2) - times(1)).toInt shouldBe 4000 +- 500
+      //      // so it backs off, later it shall reset to "normal" intervals:
+      //      (times(5) - times(4)).toInt shouldBe 1000 +- 500
+
+      called >= 6 shouldBe true
```
patriknw: I think one can write something like this: `called should >= (6)` for better error messages

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+      bootstrapA.discovery.getClass should ===(classOf[MockDiscovery])
+
+      bootstrapA.start()
+      // bootstrapB.start() // deliberately not discovering from this node, however its management port is available
+
+      val pA = TestProbe()(systemA)
+      clusterA.subscribe(pA.ref, classOf[MemberUp])
+
+      pA.expectMsgType[CurrentClusterState]
+      val up1 = pA.expectMsgType[MemberUp](45.seconds)
```
patriknw: how long does it take to run this test?

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
patriknw: it's good to only test the number of calls at the end of the test and not try to check intervals, but I think it can be somewhat more specific so that the test would fail without the new backoff implementation.  I think that currently it will pass with old implementation also? i.e. some check that called is also <= X

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
ktoso: around 16 seconds I think

ktoso: [info] - should poll discovery in exponentially increasing backoffs until eventually joining the returned nodes (20 seconds, 110 milliseconds)

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
ktoso: Right, here it really it just slows it down by rescheduling the same timer key again -- I'll add a comment

cluster-bootstrap/src/main/resources/reference.conf
ktoso: Thanks

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
ktoso: It's a bit fuzzy to put a <  on it since the latter requests could happen a few more or less... by the delays we increased the number of calls since they are failures.. not sure how to better assert -- and the timing assertions are painful on travis :\

raboof: Would this be a good opportunity to try out the ExplicitlyTriggeredScheduler (https://github.com/akka/akka-http/pull/1221/files)?

ktoso:
travis having issues fetching from github...

ktoso:
Looks green to me: https://travis-ci.org/akka/akka-management/builds/382468606

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapDiscoveryBackoffIntegrationSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+      info("" + up1)
+
+      called shouldBe >=(5)
+
+      val durationBetweenCall2And3 = (call3Timestamp - call2Timestamp).nanos
+      info(s"duration between call 2 and 3 ${durationBetweenCall2And3.toMillis} ms")
+      durationBetweenCall2And3 shouldBe >=(ClusterBootstrap(systemA).settings.contactPointDiscovery.interval * 2)
```
patriknw: @ktoso I added this test, so that it actually fails without the backoff

reduced the interval in test to 500ms

also squashed, should be ready for merge now

ktoso: Ah nice! Yeah that's nice and simpler than I was trying to do here... Thanks a ton :)
Let's merge once validates then

ktoso:
It was indeed running in parallel which caused timing issues in our timing sensitive assertions (the backoff one etc)

ktoso:
So I disabled that, still a bit fleaky though... looking into it more

Issue #72, opened 2017-12-14
Provide support for SRV records in DNS discovery


ktoso:
https://github.com/kubernetes/dns/blob/master/docs/specification.md#232---srv-records

cc @fsat

ktoso:
This will be done in Akka itself, but keeping the ticket here too as it will use it and we shall document it then too

TimMoore:
Raised #215 separately to provide API support for passing the port/protocol name in, which can then be used to construct SRV queries.

TimMoore:
This is happening in https://github.com/akka/akka/pull/25214

Issue #136, opened 2018-02-16
Add support for DNS SRV


pocman:
I'm trying to use Akka Cluster Bootstrap in replacement for ConstructR-consul with akka-dns in my Lagom project.

Currently DNS SRV is not supported but I implemented the following to solve this issue : 
```scala
package akka.discovery.dns

import akka.actor.ActorSystem
import akka.discovery._
import akka.event.Logging
import akka.pattern.ask
import com.lightbend.dns.locator.{ServiceLocator => ServiceLocatorService}

import scala.concurrent.Future
import scala.concurrent.duration.FiniteDuration

/**
 * Looks for DNS SRV records for a given service.
 */
class DnsSrvServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {

  import SimpleServiceDiscovery._

  private val log = Logging(system, getClass)
  private val dns = system.actorOf(ServiceLocatorService.props, ServiceLocatorService.Name + "_DnsSrvServiceDiscovery")

  import system.dispatcher

  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] =
    dns
        .ask(ServiceLocatorService.GetAddresses(name))(resolveTimeout)
        .mapTo[ServiceLocatorService.Addresses]
        .map {
          case ServiceLocatorService.Addresses(addresses) =>
            log.info("Resolved Dns.Resolved: {}", addresses)
            addresses.map(sa => ResolvedTarget(sa.host, Some(sa.port))).to[collection.immutable.Seq]
        }.map(Resolved(name, _))
}
```

And modified the application.conf as follow :

```yaml
akka {
  discovery {
    method = akka-dns

    akka-dns {
      # currently no special configuration is present for this module
      class = akka.discovery.dns.DnsSrvServiceDiscovery
    }
  }
```

ktoso:
Thanks! The issue about SRV is here: https://github.com/akka/akka-management/issues/72

We want to implement it natively, and then consolidate the other service locator API with this one, splitting out akka-discovery as it's own thing.

pocman:
ServiceLocatorService is only resolving over UDP, this is a problem because [DNS through UDP has a 512 bytes limit](https://serverfault.com/questions/587625/why-dns-through-udp-has-a-512-bytes-limit).
I'm going to try to use https://github.com/spotify/dns-java

pocman:
Here you go 
```
package akka.discovery.dns

import akka.actor.ActorSystem
import akka.discovery._
[...]

PR #214, opened 2018-06-07
Use akka async-dns and default to SRV records

chbatey:
Relies on https://github.com/akka/akka/pull/25214 so will need a new akka release before merging

### DIFF
--- a/discovery-aggregate/src/main/scala/akka/discovery/aggregate/AggregateSimpleServiceDiscovery.scala
+++ b/discovery-aggregate/src/main/scala/akka/discovery/aggregate/AggregateSimpleServiceDiscovery.scala
@@ -4,9 +4,9 @@
 package akka.discovery.aggregate
 
 import akka.actor.ExtendedActorSystem
-import akka.discovery.SimpleServiceDiscovery.{ Lookup, Resolved }
+import akka.discovery.SimpleServiceDiscovery.Resolved
 import akka.discovery.aggregate.AggregateSimpleServiceDiscovery.Mechanisms
-import akka.discovery.{ ServiceDiscovery, SimpleServiceDiscovery }
+import akka.discovery.{ Lookup, ServiceDiscovery, SimpleServiceDiscovery }
 import akka.event.Logging
 import akka.util.Helpers.Requiring
 import com.typesafe.config.Config
@@ -14,6 +14,7 @@ import com.typesafe.config.Config
 import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
 
 final class AggregateSimpleServiceDiscoverySettings(config: Config) {
 
@@ -66,7 +67,7 @@ final class AggregateSimpleServiceDiscovery(system: ExtendedActorSystem) extends
               Future.successful(resolved)
           }
           .recoverWith {
-            case t: Throwable =>
+            case NonFatal(t) =>
               log.error(t, "[{}] Service discovery failed. Trying next discovery mechanism", mechanism)
               resolve(tail, query, resolveTimeout)
           }
--- a/discovery-aggregate/src/main/scala/akka/discovery/aggregate/AggregateSimpleServiceDiscovery.scala
+++ b/discovery-aggregate/src/main/scala/akka/discovery/aggregate/AggregateSimpleServiceDiscovery.scala
@@ -4,7 +4,7 @@
 package akka.discovery.aggregate
 
 import akka.actor.ExtendedActorSystem
-import akka.discovery.SimpleServiceDiscovery.Resolved
+import akka.discovery.SimpleServiceDiscovery.{ Lookup, Resolved }
 import akka.discovery.aggregate.AggregateSimpleServiceDiscovery.Mechanisms
 import akka.discovery.{ ServiceDiscovery, SimpleServiceDiscovery }
 import akka.event.Logging
@@ -45,31 +45,30 @@ final class AggregateSimpleServiceDiscovery(system: ExtendedActorSystem) extends
   /**
    * Each discovery mechanism is given the resolveTimeout rather than reducing it each time between mechanisms.
    */
-  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[SimpleServiceDiscovery.Resolved] = {
-    resolve(mechanisms, name, resolveTimeout)
-  }
+  override def lookup(lookup: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] =
+    resolve(mechanisms, lookup, resolveTimeout)
 
-  private def resolve(sds: Mechanisms, name: String, resolveTimeout: FiniteDuration): Future[Resolved] = {
+  private def resolve(sds: Mechanisms, query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
     sds match {
       case (mechanism, next) :: Nil =>
-        log.debug("Looking up [{}] with [{}]", name, mechanism)
-        next.lookup(name, resolveTimeout)
+        log.debug("Looking up [{}] with [{}]", query, mechanism)
+        next.lookup(query, resolveTimeout)
       case (mechanism, next) :: tail =>
-        log.debug("Looking up [{}] with [{}]", name, mechanism)
+        log.debug("Looking up [{}] with [{}]", query, mechanism)
         // If nothing comes back then try the next one
         next
-          .lookup(name, resolveTimeout)
+          .lookup(query, resolveTimeout)
           .flatMap { resolved =>
             if (resolved.addresses.isEmpty) {
-              log.debug("Mechanism [{}] returned no ResolvedTargets, trying next", name)
-              resolve(tail, name, resolveTimeout)
+              log.debug("Mechanism [{}] returned no ResolvedTargets, trying next", query)
+              resolve(tail, query, resolveTimeout)
             } else
               Future.successful(resolved)
           }
           .recoverWith {
             case t: Throwable =>
               log.error(t, "[{}] Service discovery failed. Trying next discovery mechanism", mechanism)
-              resolve(tail, name, resolveTimeout)
+              resolve(tail, query, resolveTimeout)
           }
       case Nil =>
         // this is checked in `discoveryMechanisms`, but silence compiler warning
--- a/discovery-dns/src/main/resources/reference.conf
+++ b/discovery-dns/src/main/resources/reference.conf
@@ -2,14 +2,19 @@
 # Akka Service Discovery DNS Config                  #
 ######################################################
 
-akka.discovery {
+akka {
+  discovery {
 
-  # Set the following in your application.conf if you want to use this discovery mechanism:
-  # impl = akka-dns
+    # Set the following in your application.conf if you want to use this discovery mechanism:
+    # impl = akka-dns
 
-  # configured the akka-dns provider
-  akka-dns {
-    # currently no special configuration is present for this module
-    class = akka.discovery.dns.DnsSimpleServiceDiscovery
+    # configured the akka-dns provider
+    akka-dns {
+      # currently no special configuration is present for this module
+      class = akka.discovery.dns.DnsSimpleServiceDiscovery
+
+      # possible values: ip srv. Ip will do a A/AAAA DNS query. Srv will do an SRV DNS query.
+      lookup-type = srv
+    }
   }
 }
--- a/discovery-dns/src/main/resources/reference.conf
+++ b/discovery-dns/src/main/resources/reference.conf
@@ -12,9 +12,6 @@ akka {
     akka-dns {
       # currently no special configuration is present for this module
       class = akka.discovery.dns.DnsSimpleServiceDiscovery
-
-      # possible values: ip srv. Ip will do a A/AAAA DNS query. Srv will do an SRV DNS query.
-      lookup-type = srv
     }
   }
 }
--- a/discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
+++ b/discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
@@ -6,28 +6,23 @@ package akka.discovery.dns
 import akka.AkkaVersion
 import akka.actor.ActorSystem
 import akka.event.Logging
-import akka.io.{ Dns, IO }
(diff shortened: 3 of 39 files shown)


discovery-dns/src/main/resources/reference.conf
```diff
@@ -2,14 +2,25 @@
 # Akka Service Discovery DNS Config                  #
 ######################################################
 
-akka.discovery {
+akka {
+  io {
+    dns {
+      resolver = "async-dns"
```
patriknw: Trying to override reference.conf value from another reference.conf will not work. I think it must be documented that this is required to define in application.conf (with runtime check that fails if it isn't)

chbatey:
Updated this based on changes to async-dns and added docs

discovery-dns/src/main/resources/reference.conf
```diff
@@ -2,14 +2,19 @@
...
+    # Set the following in your application.conf if you want to use this discovery mechanism:
+    # impl = akka-dns
 
-  # configured the akka-dns provider
-  akka-dns {
-    # currently no special configuration is present for this module
-    class = akka.discovery.dns.DnsSimpleServiceDiscovery
+    # configured the akka-dns provider
+    akka-dns {
+      # currently no special configuration is present for this module
```
TimMoore: Should this comment be deleted now that there is a `lookup-type` config below?

discovery-dns/src/main/resources/reference.conf
```diff
@@ -2,14 +2,19 @@
...
+      class = akka.discovery.dns.DnsSimpleServiceDiscovery
+
+      # possible values: ip srv. Ip will do a A/AAAA DNS query. Srv will do an SRV DNS query.
+      lookup-type = srv
```
ktoso: This is confusing hm. I was under the impression one per each query knows if it's srv or normal, since srv records are supposed to start with an `_`

https://en.wikipedia.org/wiki/SRV_record

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -11,26 +12,48 @@ import akka.pattern.ask
...
 
 /**
  * Looks for A records for a given service.
  */
 class DnsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
+
+  // Required for async dns
+  // AkkaVersion.require("discovery-dns", "2.5.14")
+  require(system.settings.config.getString("akka.io.dns.resolver") == "async-dns",
+    "Akka discovery DNS requires akka.io.dns.resolver to be set to async-dns")
```
ktoso: good checks

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -11,26 +12,48 @@ import akka.pattern.ask
...
   override def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] = {
     def cleanIpString(ipString: String): String =
       if (ipString.startsWith("/")) ipString.tail else ipString
 
-    dns.ask(Dns.Resolve(name))(resolveTimeout) map {
-      case resolved: Dns.Resolved =>
-        log.info("Resolved Dns.Resolved: {}", resolved)
-        val addresses = resolved.ipv4.map { entry ⇒
-          ResolvedTarget(cleanIpString(entry.getHostAddress), port = None)
+    log.debug("Resolving {} type {}", name, settings.LookupType)
```
ktoso: are we sure this should not be `if name starts with "_" => srv lookup`?
that's how it was in the other impl as well

docs/src/main/paradox/discovery.md
```diff
@@ -19,27 +19,25 @@ We recommend using the DNS implementation as good default choice, and if an impl
...
 
-And configure it to be used as default discovery implementation in your `application.conf`:
+And configure it to be used as discovery implementation in your `application.conf` and `async-dns` to be uses
+as the Akka DNS resolver:
 
 ```
-akka.discovery {
-  method = akka-dns
+akka {
+  discovery.method = akka-dns
```
ktoso: okey so we agree we call it `akka-dns` :)

discovery-dns/src/main/resources/reference.conf
ktoso: Hmm, I'm dumb -- since this is auto discovery and name can be taken from system name... this will not have _...
However, I think it would be best if we issue both request perhaps -- try srv and if none found then issue an A request -- so people don't have to configure this at all.

WDYT?

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -11,26 +12,48 @@ import akka.pattern.ask
...
+
+    dns.ask(DnsProtocol.Resolve(name, settings.LookupType))(resolveTimeout) map {
```
ktoso: what if we did SRV lookup and if nothing then do an A lookup?

chbatey: Okay let's do that and remove the setting

ktoso: Ping for today? :)

ktoso:
Resolves https://github.com/akka/akka-management/issues/72

discovery-dns/src/main/resources/reference.conf
TimMoore: I think this relates to #224. If you pass a port name/protocol, do SRV. Otherwise, A record.

chbatey: Yes i'll rebase this on top of 224 and do that

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
chbatey: I'm going to rebase this on top of 215 and if a user does a Full lookup they get a SRV and a Simple lookup a A record

chbatey:
Picking this back up and basing on the #215

chbatey:
Hey @ktoso this is ready for another review when you have a chance. It'll need to be switched to akka 2.5.14 later, made it rely on a nightly for 2.12 for now.

Also started pulling in code examples from files rather than inline.

ktoso:
okey

discovery-aggregate/src/main/scala/akka/discovery/aggregate/AggregateSimpleServiceDiscovery.scala
```diff
@@ -4,7 +4,7 @@
 package akka.discovery.aggregate
 
 import akka.actor.ExtendedActorSystem
-import akka.discovery.SimpleServiceDiscovery.Resolved
+import akka.discovery.SimpleServiceDiscovery.{ Lookup, Resolved }
```
ktoso: 👍

discovery-dns/src/main/resources/reference.conf
ktoso: Seen this @chbatey ?

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -11,32 +12,63 @@ import akka.pattern.ask
...
 /**
  * Looks for A records for a given service.
  */
 class DnsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
+
+  // FIXME once 2.5.12 is released
+  // Required for async dns
+  // AkkaVersion.require("discovery-dns", "2.5.14")
+  require(system.settings.config.getString("akka.io.dns.resolver") == "async-dns",
+    "Akka discovery DNS requires akka.io.dns.resolver to be set to async-dns")
```
ktoso: good

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -11,32 +12,63 @@ import akka.pattern.ask
...
+            log.warning("Resolved UNEXPECTED (resolving to Nil): {}", resolved.getClass)
+            Resolved(query.name, Nil)
         }
-        Resolved(name, addresses)
 
-      case resolved ⇒
-        log.warning("Resolved UNEXPECTED (resolving to Nil): {}", resolved.getClass)
-        Resolved(name, Nil)
+      case Full(name, port, protocol) =>
+        val srvRequest = s"_$port._$protocol.$name"
```
ktoso: okey I see, makes sense

discovery-dns/src/test/scala/akka/cluster/bootstrap/discovery/DnsDiscoverySpec.scala
```diff
@@ -3,11 +3,144 @@
...
+  "Dns Discovery" must {
+
+    // Requires DNS server, see above
+    pending
+
+    "work with SRV records" in {
+      val discovery = ServiceDiscovery(system).discovery
+      val name = "_service._tcp.akka.test."
+      val result =
+        discovery.lookup(Full("akka.test.", "service", "tcp"), resolveTimeout = 500.milliseconds).futureValue
```
ktoso: nice, I like this. The name Full also works well

docs/src/main/paradox/discovery.md
```diff
@@ -13,33 +13,69 @@ If you're looking for a way to discover Actors in a Cluster, you may want to loo
 pattern](https://doc.akka.io/docs/akka/current/typed/actor-discovery.html#receptionist) from Akka 
 Typed instead. Since it provides a more fine-tuned towards Actors mechanism of doing the discovery.
 
+Discovery is done in a mechanism agnostic way, where mechanisms are things like DNS, Consul and configuration.
```
ktoso: "Consul or other key-value stores"?

docs/src/main/paradox/discovery.md
```diff
@@ -13,33 +13,69 @@ If you're looking for a way to discover Actors in a Cluster, you may want to loo
...
+Then either a simple of a full lookup can be done. How these are interpreted is discovery mechamism dependent e.g. 
+DNS does an A/AAAA record for a simple lookup and a SRV query for a full look up:
+
+Scala
+:  @@snip [CompileOnlySpec.scala]($management$/discovery/src/test/scala/doc/akka/discovery/CompileOnlySpec.scala) { #simple }
+
+Java
+:  @@snip [CompileOnlyTest.java]($management$/discovery/src/test/java/jdoc/akka/discovery/CompileOnlyTest.java) { #simple }
+
+A full lookup includes a `port` and `protocol` which can be interpreted as a mechanism chooses.
```
ktoso: Mention that this is SRV lookups?

discovery-dns/src/main/resources/reference.conf
chbatey: We've removed lookup-type now

pocman:
Hi @chbatey ,
you should be careful using DNS SRV to list all lookups.
There is payload size limitation and in some case the lookup will only return 2 fully resolved ip/port.
I'm using this code to list all allocs on consul dns srv (the payload size is higher on kubernetes but it doesn't respect the rfc).

```
package akka.discovery.dns

import akka.actor.ActorSystem
import akka.discovery
import akka.discovery.SimpleServiceDiscovery
import com.spotify.dns.{ DnsSrvResolver, DnsSrvResolvers, LookupResult }
import com.typesafe.config.{ Config, ConfigFactory }

import scala.collection.JavaConverters._
import scala.concurrent.Future
import com.spotify.dns.statistics.{ DnsReporter, DnsTimingContext }
import com.typesafe.scalalogging.LazyLogging

/**
 * Looks for DNS SRV records via tcp for a given service.
 */
class TcpDnsSrvServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery with LazyLogging {

  import SimpleServiceDiscovery._

  private val configuration: Config = ConfigFactory.load()
  System.setProperty(
    "dns.server",
    configuration
      .getStringList("akka.discovery.akka-dns.nameservers")
      .asScala
      .mkString(",")
  )
  org.xbill.DNS.Lookup.getDefaultResolver.setTCP(true)

  private val reporter = new LoggingReporter
  private val resolver: DnsSrvResolver = DnsSrvResolvers
    .newBuilder()
    .cachingLookups(false)
    .retainingDataOnFailures(true)
    .metered(reporter)
    .dnsLookupTimeoutMillis(1000)
    .build()

  private val AResolver = new DnsAResolver()(system.dispatcher)
  import system.dispatcher

  override def lookup(lookup: discovery.Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
    lookup.portName.foreach(
      portValue => logger.warn(s"TcpDnsSrvServiceDiscovery lookup doesn't support defined portName (here $portValue).")
    )
    lookup.protocol.foreach(
      protocolValue =>
        logger.warn(s"TcpDnsSrvServiceDiscovery lookup doesn't support defined protocolValue (here $protocolValue).")
    )

    val fLookupResults = Future { resolver.resolve(lookup.serviceName).asScala.toList }
    fLookupResults.flatMap { lookupResults =>
      logger.info("Resolved Dns.Resolved: {}", lookupResults.mkString(", "))

      Future.sequence(lookupResults.map { lookupResult: LookupResult =>
        AResolver.resolve(lookupResult.host()).collect {
          case Some(address) =>
            ResolvedTarget(address.getHostAddress, Some(lookupResult.port))
        }
      })
    }.map(Resolved(lookup.serviceName, _))
  }

  private class LoggingReporter extends DnsReporter {
    override def resolveTimer: DnsTimingContext = new DnsTimingContext() {
      final private val start = System.currentTimeMillis

      override def stop(): Unit = {
        val now = System.currentTimeMillis
        val diff = now - start
        logger.debug("Request took " + diff + "ms")
      }
    }

    override def reportFailure(error: Throwable): Unit = {
      logger.error("Error when resolving: " + error)
    }

    override def reportEmpty(): Unit = {
      logger.error("Empty response from server.")
    }
  }

}
```

raboof:
@pocman hmm, seems you mentioned this before in https://github.com/akka/akka-management/issues/136 but it slipped through the cracks? Might be good to register this as a separate github issue?

PR #216, opened 2018-06-08
Discovery local

herzrasen:
This is a simple implementation for #157. Using the `akka.discovery.method = akka-local` uses `akka.discovery.akka-local.service-file` as a register for locally running services. When shutting down the service the shutdownHook will unregister the stopping service. 

Input and suggestions are welcome.

### DIFF
--- a/discovery-local/src/main/scala/akka/discovery/local/LocalServiceDiscovery.scala
+++ b/discovery-local/src/main/scala/akka/discovery/local/LocalServiceDiscovery.scala
@@ -5,7 +5,7 @@ package akka.discovery.local
 
 import java.nio.file.Paths
 
-import akka.actor.ActorSystem
+import akka.actor.{ActorSystem, CoordinatedShutdown}
 import akka.discovery.SimpleServiceDiscovery
 import akka.discovery.local.registration.LocalServiceRegistration
 import akka.event.Logging
@@ -14,9 +14,10 @@ import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 
 /**
- * Looks for an entry in the service file
- */
-class LocalServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
+  * Looks for an entry in the service file
+  */
+class LocalServiceDiscovery(system: ActorSystem)
+    extends SimpleServiceDiscovery {
   import SimpleServiceDiscovery._
 
   private val log = Logging(system, getClass)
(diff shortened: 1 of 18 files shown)


discovery-local/src/main/scala/akka/discovery/local/registration/LocalServiceRegistration.scala
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Removes an <host>:<port> from the service file if registered.
+   * @param address the hostname or ip address
+   * @param port the port
+   */
+  def remove(address: String, port: Int): Unit =
+    writeToFile(localServiceEntries.filterNot(_ == LocalServiceEntry(address, port)))
+
+  private def writeToFile(localServiceEntries: Seq[LocalServiceEntry]) =
+    Files.write(serviceFile, localServiceEntries.toJson.compactPrint.getBytes(utf8))
```
ktoso: You could get 2 discoveries attempting to write and corrupt the file this way. Needs to take filesystem lock https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileLock.html around these operations. This likely is good enough as we're not talking about any high performance stuff here, just avoiding writing into it from more apps

discovery-local/src/main/scala/akka/discovery/local/LocalServiceDiscovery.scala
```diff
@@ -0,0 +1,74 @@
...
+    log.debug(s"Using $serviceFile as service file")
+    val addresses =
+      localServiceRegistration.localServiceEntries.map { entry =>
+        ResolvedTarget(entry.addr, Some(entry.port))
+      }
+
+    // remove self from the service file on shutdown
+    sys.addShutdownHook {
+      unregisterSelf()
+    }
```
ktoso: this should rather use the coordinated shutdown hooks, not a system wide one.
See https://doc.akka.io/docs/akka/current/actors.html#coordinated-shutdown

herzrasen:
@ktoso thanks for your remarks. I updated the sources accordingly.

herzrasen:
Any chance this is going to be merged?

chbatey:
It is very similar to the config based service discovery. However maybe there's a usecase for file based that is not inside the application's config? WDYT @herzrasen @ktoso ?

herzrasen:
@chbatey it is indeed very similar. I use it in testing environments where no DNS is present. I could basically do the same with an SEED_NODES environement variable which is only set in my test envs.

ktoso:
I was somewhat in progress of reviewing this, but the general feel is that this is not quite what i had envisioned as the one I had in mind was also about registering into it on startup becoming a "super easy no need to configure anything mode" for demos and playing around. Not sure if we'd like to head there with this PR, WDYT?

herzrasen:
I understand what you mean. I'm just not quite sure how this could be implemented since some kind of configuration is needed in any case.

Issue #209, opened 2018-05-28
Cluster bootstrapping with Istio


idoshamun:
Istio blocks direct connection between pods, is there a way to overcome this to allow the cluster bootstrapping?

This is the error I get right now:
```
21:03:21.123 [error] akka.management.cluster.bootstrap.internal.HttpContactPointBootstrap - Probing [http://10.60.1.32:3311/bootstrap/seed-nodes] failed due to: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests
```

ktoso:
We have not looked into Istio so far in this context; if you could look into it and share your findings that would be very useful

idoshamun:
The current conclusion is that Istio doesn't allow pod to pod communication and there is no way to disable this restriction. Meaning that we can't form an Akka cluster using this method on an Istio cluster.
Here is the relevant thread:
[https://groups.google.com/forum/#!topic/istio-users/d-THsO19oAM](https://groups.google.com/forum/#!topic/istio-users/d-THsO19oAM)

Any other ideas to form a cluster?

thomschke:
It's works by using pod dns names [see](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods).

- service discovery by pod name ( #217 )

- set hostname =`${POD_IP//./-}.default.pod.cluster.local`:

```
apiVersion: "apps/v1"
kind: Deployment
metadata:
  name: akka
spec:
  replicas: 2
  selector:
    matchLabels:
      app: akka
  template:
    metadata:
      labels:
        app: akka
    spec:
      containers:
      - name: akka
        ports:
        - containerPort: 10001
          name: "akka-remote"
        - containerPort: 10002
          name: "akka-mgmt-http"
        env:
        - name: "DYN_JAVA_OPTS"
          value: "-Dakka.remote.netty.tcp.hostname=${POD_IP//./-}.default.pod.cluster.local -Dakka.management.http.hostname=${POD_IP//./-}.default.pod.cluster.local"

```

- configure an internal ServiceEntry [see](https://istio.io/docs/reference/config/istio.networking.v1alpha3/#ServiceEntry) to allow outgoing calls:

```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: akk
spec:
  hosts:
  - "*.default.pod.cluster.local"
  location: MESH_INTERNAL
  ports:
  - number: 10001
    name: "akka-remote"
    protocol: TCP
[...]

PR #217, opened 2018-06-19
k8s pod lookup by name (#209)

thomschke:
Istio service mesh requires that every service (e.g. bootstrap contact points) be requested by name (not by ip-addr).

Pods are assigned a DNS A record in the form of “**pod-ip-address**.my-namespace.**pod**.cluster.local”. [see](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods)

This is a simple implementation for k8s pod lookup by name.


ktoso:
Resolves #209

idoshamun:
@ktoso When should it be released so I can validate it solves Istio inter-communication issue?

Issue #176, opened 2018-04-18
Aggregate service discovery


ktoso:
Nowadays people can do AkkaDiscovery(). to get a "default one".
The default one is configured to a specific one.
It is useful to be able to configure by address range or prefix match, which lookup mechanism should be used for given input. Imagine a set of known hardcoded things, and a set of not known service located things which one has to go via DNS or other discovery -- this way, once configured, one does not have to change code once these change etc.

One can use hardcoded for locally being developed services, but keep the located ones for other cloud things etc and mix it depending how you need.

```
akka.discovery { 
  default = aggregate // or composite

  aggregate { 
    switch { 
      "*.example.com": akka-dns
      "google.com": gcloud-api 
     "*service": 127.0.0.1 // hardcode it for development time if you want (can be set via -D value on starting app etc)
      etc etc
```

ignasi35:
This would be of great use for Lagom, for example. 

In current Lagom, configuring lookup for Kafka and Cassandra slightly differs. Both default to use Lagom's Service Locator but overriding each of the to use hardcoded [brokers lists](https://www.lagomframework.com/documentation/1.4.x/java/ProductionOverview.html#Using-static-Cassandra-contact-points) (Kafka) or [contact points](https://www.lagomframework.com/documentation/1.4.x/java/KafkaClient.html#Lagom-Kafka-Client1) (Cassandra) differs in both the format and the settings required.

TimMoore:
You'll also need to be able to specify ports for hardcoded services.

Given that, I think something simpler than what's proposed above might be enough. Do we really need pattern matching on the names and multiple destinations?

I'd rather see an ordered sequence of lookup mechanisms, where anything not found in an earlier one is looked up in the next one.

The configuration mechanism could just be a straight map of `"name": "host:port"` pairs.

It could also be useful to pass even the results of successful lookups on to the next stage, so you could provide config like `"user-service": "_https._tcp.user-service.some-namespace.svc.cluster.local"` and then it will still get resolved by the DNS SRV mechanism.

alanngai:
this is not a well formed thought, but I imagine at some point this may run up against alternative solutions to the same general problem of "selecting" different configurations for different environments.  In the case of kubernetes for example, the "canonical" solution is to use ConfigMap objects, which would be different for each environment, to house environment specific settings.  There is a request, for example, for reactive-lib to support ConfigMaps (https://github.com/lightbend/reactive-lib/issues/51).

[...]

PR #218, opened 2018-06-21
Aggregate and config discovery

chbatey:
Refs #176 and #166 

This adds a config and aggregate discovery. I haven't updated it to the default yet.


docs/src/main/paradox/discovery.md
```diff
@@ -521,6 +569,53 @@ Notes:
...
+  artifact="akka-discovery-aggregate_2.12"
+  version="$version$"
+}
+
+Configure `akka-aggregate` as `akka.discovery.method` and which discovery mechanisms are tried and in which order.
+
+```
+akka {
+  discovery {
+    method = akka-aggregate
```
ktoso: not sure why "akka"

docs/src/main/paradox/discovery.md
```diff
@@ -521,6 +569,53 @@ Notes:
...
+    akka-aggregate {
+      discovery-mechanisms = ["akka-dns", "akka-config"]
+    }
+    akka-config {
+      services = {
```
ktoso: would prefer ` { ` rather than ` = {` for maps

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -21,6 +21,7 @@ import scala.concurrent.duration.FiniteDuration
 @ApiMayChange
 object SimpleServiceDiscovery {
 
+  // TODO for use cases like aggregate it might be worth making this an ADT with a NoTargets case
```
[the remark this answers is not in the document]
chbatey: Maybe not worth it. Just some times it might make sense to distinguish between a service being there but no registered endpoints vs not being there at all e.g. in config, consul etc

docs/src/main/paradox/discovery.md
chbatey: All of the inbuilt ones have the prefix akka-dns, akka-config etc

docs/src/main/paradox/discovery.md
chbatey: Yep

discovery-config/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
+akka {
+  discovery {
+    akka-config {
+      class = akka.discovery.config.ConfigSimpleServiceDiscovery
+
+      # Location of the services
+      services-path = "akka.discovery.akka-config.services"
+
+      # Services. See docs for examples of format.
```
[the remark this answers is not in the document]
chbatey: Will add

discovery-config/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
+akka {
+  discovery {
+    akka-config {
```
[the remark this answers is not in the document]
chbatey: Think i'll remove all the akka prefixes

docs/src/main/paradox/discovery.md
ktoso: That is the complaint ;) about both akka-dns and akka-config actually hm... akka-dns i could see at least a bit since it is "the akka impl" rather than some other one, but this is not "akka config"

ktoso:
WDYT about https://github.com/akka/akka-management/pull/218#discussion_r197701177 ? and then merge plz

discovery-config/src/main/scala/akka/discovery/config/ConfigSimpleServiceDiscovery.scala
```diff
@@ -0,0 +1,50 @@
...
+  private val log = Logging(system, getClass)
+
+  private val resolvedServices = ConfigServicesParser.parse(
+    system.settings.config.getConfig(system.settings.config.getString("akka.discovery.akka-config.services-path"))
+  )
+
+  log.debug("Config discovery serving: {}", resolvedServices)
+
+  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[SimpleServiceDiscovery.Resolved] = {
+    // TODO or fail or change the Resolved type to an ADT?
```
TimMoore: I think this is good as is… especially given the implementation of the aggregate above. You wouldn't want it to print a noisy error every time you look up a service that is not in the config.

If there were an ADT, then I suppose you would still want to handle the case of a successful empty list for completeness sake, so IMHO there's no reason to add another case to handle.

docs/src/main/paradox/discovery.md
```diff
@@ -521,6 +569,53 @@ Notes:
...
+      }
+    }
+  }
+}
+
+```
+
+The above configuration will result in `akka-dns` first being checked and if it fails or returns no
+targets for the given service name then `akka-config` is queried which i configured with one service called
+`service1` which two hosts `host1` and `host2`.
```
TimMoore: I think in most realistic scenarios you'd want to go the other way around. I know it's just an example, but people tend to copy and paste things :smile:

docs/src/main/paradox/discovery.md
TimMoore: I agree that it would read better without the prefix.

chbatey:
I've removed the prefix and will merge once validation has run

PR #220, opened 2018-06-26
upgrade akka to 2.5.13 and akka-http to 10.1.3

shkoder:


project/Dependencies.scala
```diff
@@ -3,8 +3,8 @@ import Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.5.7"
-  val AkkaHttpVersion = "10.0.13" // TODO update to 10.1.x as soon as possible
+  val AkkaVersion = "2.5.13"
```
ktoso: yes

project/Dependencies.scala
```diff
@@ -3,8 +3,8 @@ import Keys._
...
+  val AkkaHttpVersion = "10.1.3"
```
ktoso: no. you can already use it with 10.1 but bumping the version like this may break still existing 10.0 users.
We are not doing this right now. We will shortly though.

Please remove the change on line 7 :)

shkoder: sure, removed

Issue #222, opened 2018-06-26
Kubernetes Service Discovery returns a label


eed3si9n:
### steps

1. Run Lagom on GKS?

### problem

```
akka.actor.ActorSystemImpl [sourceThread=application-akka.actor.default-dispatcher-6, akkaTimestamp=xxxx, akkaSource=akka.actor.ActorSystemImpl(application), sourceActorSystem=application] 
- Querying for pods with label selector: [appName=appName=foo]

....

akka.actor.ActorSystemImpl [sourceThread=application-akka.actor.default-dispatcher-3, akkaTimestamp=xxxx, akkaSource=akka.actor.ActorSystemImpl(application), sourceActorSystem=application]
- Kubernetes API entity: [{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"found '=', expected: ',' or 'end of string'","reason":"BadRequest","code":400}
```

### expectation

`appName=foo`

### notes

I suspect this is the problem:
https://github.com/akka/akka-management/blob/a489147c4c44bd35cbaaf36a7971e368b8e71597/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala#L97

PR #223, opened 2018-06-26
Fixes Kubernetes API Revolved value

eed3si9n:
Fixes https://github.com/akka/akka-management/issues/222

The `Revolved` record expects the service name, but the Kubernetes API discovery's `Resolved` record currently returns the label selector instead of the queried service name.

This causes Lagom to fail on startup on Kubernetes with a strange error.

```
akka.actor.ActorSystemImpl [sourceThread=application-akka.actor.default-dispatcher-3, akkaTimestamp=xxxx, akkaSource=akka.actor.ActorSystemImpl(application), sourceActorSystem=application]
- Kubernetes API entity: [{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"found '=', expected: ',' or 'end of string'","reason":"BadRequest","code":400}
```

This is likely caused by querying  `app=app=foo`.


ktoso:
Thanks! Great catch

ktoso:
Also: giving you both write access :)

Issue #215, opened 2018-06-08
Service discovery API for named ports/protocols


TimMoore:
Many service discovery implementations have some mechanism for distinguishing between multiple endpoints of a single service (for example, a service might listen to ports for http, akka-management-http, and akka-remote).

The details of how this works vary:

- [DNS SRV](https://www.ietf.org/rfc/rfc2782.txt) uses lookups of the form `_<service>._<protocol>.<name>`
- In [Kubernetes](https://github.com/kubernetes/dns/blob/master/docs/specification.md#232---srv-records), this translates to the form `_<port>._<proto>.<service>.<ns>.svc.<zone>.`, where each service defines a list of [port names and protocols](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services)
- [Marathon](https://mesosphere.github.io/marathon/docs/networking.html#specifying-ports) works similarly to Kubernetes
- In [Consul's DNS SRV interface](https://www.consul.io/docs/agent/dns.html#rfc-2782-lookup), the service name is the first component, and the `<protocol>` component is translated to a Consul tag lookup.
- Consul's HTTP interface also support [filtering queries by tag](https://www.consul.io/api/catalog.html#tag)
- Most others have some similar kind of metadata/tagging facility

The Akka `SimpleServiceDiscovery` API should allow some way to optionally specify this additional port name and protocol along with the name. Then, individual implementation can decide the best way to use that information to query the back end.

TimMoore:
The type of case we need to handle is when we want to build something on top of the service discovery API that should be independent of the underlying implementation. So for example in Lagom, if you make a request to a service named "user-service" (where the name is defined in the service API descriptor) then it looks up "user-service" in the service locator.

But then, service locator might be listening on several different ports: one for the main http server, one for akka-management-http, one for akka-remote, and now maybe a separate grpc port (still TBD whether we can have that share the same port as the main http server in early iterations without needing a lot more work).

So we need to translate "user-service" to some SRV name inside the DNS service discovery implementation, but how do we know which prefix to use?

Right now, Lightbend Orchestration hardcodes "_http", which it is also hardcoded to generate as a port name for Lagom services when creating the Kubernetes service definition YAML.

Not only is this too inflexible to handle a separate grpc port, but it also doesn't work correctly right now for Kafka and Cassandra, which require manual configuration.

[...]

PR #224, opened 2018-06-29
Optional port and protocol for lookup API

chbatey:
Just raising to get ppl's opinions about what an API for ports and protocols might look like.

I wouldn't like to not take in a service for SRV otherwise we'd need to pull off the service to put the protocol in as it is: `_service._proto.name`

Should it just be a map?

They all need to be optional apart from name as even for a single mechanism they aren't mandatory e.g. protocol is for SRV DNS but not for A DNS.

### DIFF
--- a/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
+++ b/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
@@ -35,14 +35,16 @@ final class MockDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
   private val log = Logging(system, getClass)
 
-  override def lookup(name: String, resolveTimeout: FiniteDuration): Future[Resolved] =
-    MockDiscovery.data.get().get(name) match {
+  override def lookup(query: SimpleServiceDiscovery.Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
+    MockDiscovery.data.get().get(query.name) match {
       case Some(res) ⇒
         val items = res()
-        log.info("Mock-resolved [{}] to [{}:{}]", name, items, items.value)
+        log.info("Mock-resolved [{}] to [{}:{}]", query.name, items, items.value)
         items
       case None ⇒
-        log.info("No mock-data for [{}], resolving as 'Nil'", name)
-        Future.successful(Resolved(name, Nil))
+        log.info("No mock-data for [{}], resolving as 'Nil'", query.name)
+        Future.successful(Resolved(query.name, Nil))
     }
+  }
+
 }
--- a/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
+++ b/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
@@ -35,15 +35,15 @@ final class MockDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
   private val log = Logging(system, getClass)
 
-  override def lookup(name: String, meta: LookupMetadata, resolveTimeout: FiniteDuration): Future[Resolved] = {
-    MockDiscovery.data.get().get(name) match {
+  override def lookup(query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
+    MockDiscovery.data.get().get(query.serviceName) match {
       case Some(res) ⇒
         val items = res()
-        log.info("Mock-resolved [{}] to [{}:{}]", name, items, items.value)
+        log.info("Mock-resolved [{}] to [{}:{}]", query.serviceName, items, items.value)
         items
       case None ⇒
-        log.info("No mock-data for [{}], resolving as 'Nil'", name)
-        Future.successful(Resolved(name, Nil))
+        log.info("No mock-data for [{}], resolving as 'Nil'", query.serviceName)
+        Future.successful(Resolved(query.serviceName, Nil))
     }
   }
 
--- a/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
+++ b/cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
@@ -35,15 +35,15 @@ final class MockDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
   private val log = Logging(system, getClass)
 
-  override def lookup(query: SimpleServiceDiscovery.Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
-    MockDiscovery.data.get().get(query.name) match {
+  override def lookup(name: String, meta: LookupMetadata, resolveTimeout: FiniteDuration): Future[Resolved] = {
+    MockDiscovery.data.get().get(name) match {
       case Some(res) ⇒
         val items = res()
-        log.info("Mock-resolved [{}] to [{}:{}]", query.name, items, items.value)
+        log.info("Mock-resolved [{}] to [{}:{}]", name, items, items.value)
         items
       case None ⇒
-        log.info("No mock-data for [{}], resolving as 'Nil'", query.name)
-        Future.successful(Resolved(query.name, Nil))
+        log.info("No mock-data for [{}], resolving as 'Nil'", name)
+        Future.successful(Resolved(name, Nil))
     }
   }
 
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
@@ -24,6 +24,8 @@ import scala.util.Try
 import JsonFormat._
 import SimpleServiceDiscovery.{ Resolved, ResolvedTarget }
 
+import scala.util.control.NonFatal
+
 object KubernetesApiSimpleServiceDiscovery {
 
   /**
@@ -73,11 +75,11 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
 
   private val httpsContext = http.createClientHttpsContext(httpsConfig)
 
-  override def lookup(name: String, metadata: LookupMetadata, resolveTimeout: FiniteDuration): Future[Resolved] =
+  override def lookup(query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] =
     for {
       token <- apiToken()
 
-      labelSelector = settings.podLabelSelector(name)
+      labelSelector = settings.podLabelSelector(query.serviceName)
 
       _ = system.log.info("Querying for pods with label selector: [{}]", labelSelector)
 
@@ -94,7 +96,7 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
         val unmarshalled = Unmarshal(entity).to[PodList]
 
         unmarshalled.failed.foreach { t =>
-          system.log.error("Failed to unmarshal Kubernetes API response status [{}]; check RBAC settings",
+          system.log.error(t, "Failed to unmarshal Kubernetes API response status [{}]; check RBAC settings",
             response.status.value)
         }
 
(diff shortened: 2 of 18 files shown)


ktoso:
Please link back to the issue; Relates to https://github.com/akka/akka-management/issues/215

TimMoore:
How is `service` different from `port`? I expected three parameters, service name, port and protocol.

chbatey:
You're right @TimMoore we don't need both, i was taking the union of two sets of terminologies

I've updated this to use sealed class hierarchy for the lookup types to make using it nicer. I think having a bunch of optional values ugly to deal with.

chbatey:
This is ready for review now. Next step is to implement the "Full" lookup for DNS once we have a new Akka version.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -54,6 +55,29 @@ object SimpleServiceDiscovery {
...
+  @ApiMayChange
+  sealed abstract class Lookup {
+    def name: String
+  }
+
+  @ApiMayChange
+  final case class Simple(name: String) extends Lookup
+
+  /**
+   * The meaning of each field is up to the discovery mechanism.
```
ktoso: Yeah that's probably okey though should include a `see Lookup` for more details or copy that info?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -54,6 +55,29 @@ object SimpleServiceDiscovery {
...
+   * same query being sent to multiple mechanisms via the aggregate service
+   * discovery
+   */
+  @ApiMayChange
+  sealed abstract class Lookup {
+    def name: String
+  }
+
+  @ApiMayChange
+  final case class Simple(name: String) extends Lookup
```
ktoso: needs docs, that it is simple "juist by name, see also the Full one for more specific selection criterias?"

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -54,6 +55,29 @@ object SimpleServiceDiscovery {
...
+
+  /**
+   * The meaning of each field is up to the discovery mechanism.
+   */
+  @ApiMayChange
+  final case class Full(name: String, port: String, protocol: String) extends Lookup
```
ktoso: port more like portName perhaps?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -85,9 +118,15 @@ abstract class SimpleServiceDiscovery {
...
    */
-  def lookup(name: String, resolveTimeout: java.time.Duration): CompletionStage[Resolved] = {
+  def lookup(query: Lookup, resolveTimeout: java.time.Duration): CompletionStage[Resolved] = {
     import scala.compat.java8.FutureConverters._
-    lookup(name, FiniteDuration(resolveTimeout.toMillis, TimeUnit.MILLISECONDS)).toJava
+    lookup(query, FiniteDuration(resolveTimeout.toMillis, TimeUnit.MILLISECONDS)).toJava
   }
 
+  /**
+   * Java API
```
ktoso: but also actual docs please :)

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -63,18 +87,27 @@ object SimpleServiceDiscovery {
...
+   */
+  def lookup(query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved]
+
+  /**
+   * Scala API: Perform lookup using underlying discovery implementation.
    *
-   * The returned future SHOULD be failed once resolveTimeout has passed.
+   * @param name A name, see discovery-mechanism's docs for how this is interpreted
+   * @param resolveTimeout Timeout. Up to the discovery-mechanism to adhere to his
+   * @return
```
ktoso: remove or properly document return value?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -63,18 +87,27 @@ object SimpleServiceDiscovery {
...
   /**
-   * Scala API: Perform basic lookup using underlying discovery implementation.
+   * Scala API: Perform lookup using underlying discovery implementation.
    *
-   * While the implementation may provide other settings and ways to configure timeouts,
-   * the passed `resolveTimeout` should never be exceeded, as it signals the application's
-   * eagerness to wait for a result for this specific lookup.
+   * @param query A simple of full query, see discovery-mechanism's docs for how this is interpreted
+   * @param resolveTimeout Timeout. Up to the discovery-mechanism to adhere to his
+   * @return
```
ktoso: remove or properly document return value?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -54,6 +55,29 @@ object SimpleServiceDiscovery {
...
+   * Implementations are free to ignore port/protocol from a Full lookup
+   * but should not throw if they are set unexpectedly as this prevents the
+   * same query being sent to multiple mechanisms via the aggregate service
+   * discovery
+   */
+  @ApiMayChange
+  sealed abstract class Lookup {
+    def name: String
+  }
+
```
ktoso: I now realised that these Simple/Full may be better perhaps as factory methods on `Lookup` -- instead of the separate classes?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
ktoso: perhaps also `require(!=null)` on those? Document if null or empty is suggested etc

docs/src/main/paradox/discovery.md
```diff
@@ -13,6 +13,33 @@ If you're looking for a way to discover Actors in a Cluster, you may want to loo
...
+    val result: Future[Resolved] = discovery.lookup("service-name", resolveTimeout = 500 milliseconds)
+    ```
+
+Java
+:   ```java
+    import akka.discovery.ServiceDiscovery; 
+    ActorSystem system = ActorSystem.create("Example");
+    // ... 
+    SimpleServiceDiscovery discovery = ServiceDiscovery.get(system).discovery();
+    Future<SimpleServiceDiscovery.Resolved> result = discovery.lookup("service-name", Duration.create("500 millis"));
```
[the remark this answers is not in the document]
chbatey: I've made this pull code from java/scala files in https://github.com/akka/akka-management/pull/214/files so will fix there

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -56,28 +56,32 @@ object SimpleServiceDiscovery {
...
+}
 
-  @ApiMayChange
-  final case class Simple(name: String) extends Lookup
+/**
+ * Additional information for a service lookup. It is up to each mechanism to decide
+ * what to do with this.
+ */
+@ApiMayChange
+class LookupMetadata(portName: Option[String] = None, protocol: Option[String] = None) {
```
ktoso: uhm... Don't be angry but isn't this a bit weird...?
A normal Lookup(name, portname, protocol) would have been fine I thought?

ktoso: This makes it a bit harder to pass around something to be idenfitied -- "the thing" is IDd after all using name+portName+protocol for SVC.

If you really love this idea we can keep it but what I thought we had agreed on was the `Lookup(name, opt port name, opt protocol)`

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -54,6 +55,23 @@ object SimpleServiceDiscovery {
         case None => host
       }
   }
+
+}
+
+/**
+ * A service lookup. It is up to each mechanism to decide
+ * what to do with the optional portName and protocol fields.
```
patriknw: I don't think `portName` and `protocol` are self describing so would be good to clarify what those are and examples of such, using things like Akka Cluster (remoting), gRPC, HTTP.

chbatey: :+1:

cluster-bootstrap/src/test/scala/akka/discovery/MockDiscovery.scala
```diff
@@ -35,15 +35,15 @@ final class MockDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
   private val log = Logging(system, getClass)
 
-  override def lookup(name: String, meta: LookupMetadata, resolveTimeout: FiniteDuration): Future[Resolved] = {
-    MockDiscovery.data.get().get(name) match {
+  override def lookup(query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
```
ktoso: yup, thanks!

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
```diff
@@ -87,7 +89,7 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
         val unmarshalled = Unmarshal(entity).to[PodList]
 
         unmarshalled.failed.foreach { t =>
-          system.log.error("Failed to unmarshal Kubernetes API response status [{}]; check RBAC settings",
+          system.log.error(t, "Failed to unmarshal Kubernetes API response status [{}]; check RBAC settings",
```
ktoso: good

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -59,29 +59,28 @@ object SimpleServiceDiscovery {
 }
 
 /**
- * Additional information for a service lookup. It is up to each mechanism to decide
- * what to do with this.
+ * A service lookup. It is up to each mechanism to decide
+ * what to do with the optional portName and protocol fields.
+ * For example portName could be used to distinguish between
```
ktoso:
```
`portName`
```

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -59,29 +59,28 @@ object SimpleServiceDiscovery {
...
+ * akka remoting ports and http ports.
```
ktoso: Akka

ktoso: HTTP

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -59,29 +59,28 @@ object SimpleServiceDiscovery {
...
+ *
  */
 @ApiMayChange
-class LookupMetadata(portName: Option[String] = None, protocol: Option[String] = None) {
+case class Lookup(serviceName: String, portName: Option[String] = None, protocol: Option[String] = None) {
 
-  def withPortName(value: String): LookupMetadata = copy(
+  /**
+   * Which port for a service e.g. akka-remoting or http. Maps to "service" for an SRV records.
```
ktoso: HTTP

ktoso: The maps... in a new paragraph

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -59,29 +59,28 @@ object SimpleServiceDiscovery {
...
+   */
+  def withPortName(value: String): Lookup = copy(
     portName = Some(value)
   )
 
-  def withProtocol(value: String): LookupMetadata = copy(
+  /**
+   * Which protocol e.g. tcp or udp. Maps to "protocol" for SRV records.
```
ktoso: as above

PR #225, opened 2018-07-02
=boot avoid "double calling" joinSeedNodes

ktoso:
Avoid calling joinSeedNodes in bootstrap again if the same nodes were already used to issue a join call.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
@@ -122,6 +122,9 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
   private var lastContactsObservation: Option[DnsServiceContactsObservation] = None
   private var seedNodesObservations: Map[ResolvedTarget, SeedNodesObservation] = Map.empty
 
+  // set once joinSeedNodes is called; we only want to call joinSeedNodes again once the value changes;
+  // so we keep the last time used value around in this Set of addresses that we
+  private var joiningSeedNodesInProgress: Set[Address] = Set.empty
   private var decisionInProgress = false
   def startPeriodicDecisionTimer(): Unit =
     timers.startPeriodicTimer(DecideTimerKey, DecideTick, settings.contactPoint.probeInterval)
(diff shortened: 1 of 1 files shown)


ktoso:
Please review for exact wording in comments; I've yet to dig deeper into the ClusterDemon to understand this more; but we want to avoid it in any case.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -212,12 +215,23 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
+              joiningSeedNodesInProgress = addresses
+              cluster.joinSeedNodes(addresses.toList) // order doesn't matter as self node is excluded from the list
+
+              // once we issued a join bootstrapping is completed
+              replyTo ! BootstrappingCompleted
+              context.stop(self)
+            } else {
+              log.debug(
+                  "Received JoinOtherSeedNodes() decision for the same seed nodes that we are already in process of joining ({})",
+                  joiningSeedNodesInProgress)
```
patriknw: `joiningSeedNodesInProgress.mkString(", ")`

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -212,12 +215,23 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
+
+            // we avoid calling joinSeedNodes with the same set of nodes
+            // this would interrupt and start a new join seed nodes process and cause needless friction
+            // (would still work though)
+            if (addresses != joiningSeedNodesInProgress) {
+              joiningSeedNodesInProgress = addresses
+              cluster.joinSeedNodes(addresses.toList) // order doesn't matter as self node is excluded from the list
+
+              // once we issued a join bootstrapping is completed
+              replyTo ! BootstrappingCompleted
```
patriknw: `BootstrappingCompleted` is not sent back for the else case

chbatey:
I am not sure how this can happen/help as as soon as we've called joinSeedNodes context.stop(self) is called and AFAICT we never re-create a bootstrap actor?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -212,12 +215,23 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
...
+              context.stop(self)
```
ktoso: Eh I'm so blind... Yeah this means we're stopped once we call that...

patriknw: something is not matching up here. If it stopped after calling `cluster.joinSeedNodes` then then keeping track of  previous addresses will not help, it's not the same actor instance next time

Looks like `BootstrapCoordinator` actor is started once from `ClusterBootstrap.start`, so something external is retrying the `ClusterBootstrap(system).start()`

ktoso:
No, we never recreate it... that's so weird... And I'm too tired to have spotted that `stop()`, what a day :-/

I wonder why we've seen 2 bootstrap processes in a customer thing... this was the assumption but yeah it should never be able to call another one hm

ktoso:
I will aproach this again from the Akka perspective tomorrow

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
ktoso: Indeed, the PR and idea is nonsense I realised once I realised the stop there... 
Need more sleep, will approach from other angle tomorrow

Issue #236, opened 2018-07-10
Cluster bootstrap based on kubernetes-api discovery does not work


raboof:
In https://github.com/akka/akka-management/pull/217, the kubernetes-api discovery was updated to return hostnames rather than IP addresses, for compatibility with Istio.

However, the 'self' address is still an IP address. Because of this, the cluster bootstrap will never 'join itself', since it does not find itself in the discovered endpoints. There are a couple of directions possible to approach this problem:

* Ask the user to configure `akka.management.http.hostname`. This *might* be possible via [pod environment variables](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/), but seems like a big burden to put on the user.
* Somehow convert the 'self' address from IP address to hostname as well. We have all the information necessary to perform this conversion available, but it's not obvious what would be the suitable place to perform this conversion since service discovery backens and cluster bootstrap selection logic are mostly orthogonal concepts.
* Variation on the previous solution: make `ClusterBootstrap.selfContactPoint` return a list of names under which 'self' might be available, and return both the original IP and the reconstructed k8s hostname.
* Extend `ResolvedTarget` so it can contain both the hostname and the IP address. Might not make sense for all disovery mechanisms though.

This could be considered a special case of the more general issue described in https://github.com/akka/akka-management/issues/187

PR #238, opened 2018-07-10
Allow multiple 'self' addresses in cluster bootstrap

raboof:
Refs #236

### DIFF
--- a/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
+++ b/bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
@@ -70,3 +70,26 @@ spec:
   - protocol: TCP
     port: 8558
     targetPort: 8558
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pod-reader
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: read-pods
+subjects:
+# Note the `name` line below. The first default refers to the namespace. The second refers to the service account name.
+# For instance, `name: system:serviceaccount:myns:default` would refer to the default service account in namespace `myns`
+- kind: User
+  name: system:serviceaccount:default:default
+roleRef:
+  kind: Role
+  name: pod-reader
+  apiGroup: rbac.authorization.k8s.io
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -8,7 +8,6 @@ import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.duration._
 import scala.concurrent.Future
 import scala.concurrent.Promise
-import scala.collection.immutable
 
 import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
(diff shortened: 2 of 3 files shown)


chbatey:
It would be good to get someone with recent hands on k8s experience to check this

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -108,8 +109,19 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     _selfContactPointUri.success(baseUri)
 
   /** INTERNAL API */
-  @InternalApi private[akka] def selfContactPoint: Future[Uri] =
-    _selfContactPointUri.future
+  @InternalApi private[akka] def selfContactPoints: Future[immutable.Seq[(String, Int)]] =
```
patriknw: `Set` instead of `Seq`?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -108,8 +109,19 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
+    _selfContactPointUri.future.map { uri =>
+      if (system.settings.config.hasPath("akka.discovery.kubernetes-api.pod-domain") && uri.authority.host.isIPv4()) {
+        val podNamespace = system.settings.config.getString("akka.discovery.kubernetes-api.pod-namespace")
+        val podDomain = system.settings.config.getString("akka.discovery.kubernetes-api.pod-domain")
+        immutable.Seq(
+          (uri.authority.host.toString, uri.authority.port),
+          (s"${uri.authority.host.toString.replace('.', '-')}.${podNamespace}.pod.${podDomain}", uri.authority.port)
```
patriknw: which of those is used when we sort the addresses, or doesn't that matter?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -108,8 +109,19 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     _selfContactPointUri.success(baseUri)
 
   /** INTERNAL API */
-  @InternalApi private[akka] def selfContactPoint: Future[Uri] =
-    _selfContactPointUri.future
+  @InternalApi private[akka] def selfContactPoints: Future[immutable.Seq[(String, Int)]] =
+    _selfContactPointUri.future.map { uri =>
+      if (system.settings.config.hasPath("akka.discovery.kubernetes-api.pod-domain") && uri.authority.host.isIPv4()) {
```
patriknw: Somewhat ugly that we have to have special case for kubernetes here. Wonder if it could be defined or set by the kubernetes-api "plugin" instead. Perhaps we should revisit the design of how the `_selfContactPointUri` is set in a separate ticket.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
raboof: This doesn't matter: we sort the resolved targets, not the 'self contact point' (now 'self contact points')

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
raboof: I agree. I didn't like having k8s code 'slip in' here, but making this 'pluggable' as well seemed a bit over-engineered. Indeed we might find a nice way to do it later, though.

bootstrap-joining-demo/kubernetes-api/kubernetes/akka-cluster.yml
```diff
@@ -70,3 +70,26 @@ spec:
   - protocol: TCP
     port: 8558
     targetPort: 8558
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
```
ktoso: So sneaking this in right away I guess; sure we're ok with that? (I think we are but double checking)

raboof: I don't *really* know, I just copy-pasted the recommendation from the docs into the example project

jroper:
This puts a reverse dependency between cluster bootstrap and kubernetes API discovery which is incompatible with https://github.com/akka/akka-management/pull/415, causing https://github.com/akka/akka-management/issues/424. One solution to this might be for Akka discovery to offer an optional method for deriving a self host name, which would allow Kubernetes discovery to pass back its host name.

Issue #187, opened 2018-05-02
Cluster bootstrap with DNS does not work when akka management host is a host rather than IP


chbatey:
akka1 resolves to say IP 127.0.0.1

The self join check then compares akka1 with 127.0.0.1 at:

https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala#L138

Example config

```
akka {

  actor {
    provider = "cluster"
  }

  remote {
    artery {
      enabled = on
      transport = tcp
      canonical.hostname = "akka1"
      canonical.port = 25551
    }
  }

  discovery {
    method = akka-dns
  }

  management {
    http {
      port = 19999
      hostname = "akka1"
    }

    cluster.bootstrap {
      contact-point-discovery {
        service-name = "akka1"
        required-contact-point-nr = 1

      }
      contact-point {
        fallback-port = 19999
      }
    }
  }

}
```

patrick-premont:
I've observed the same referenced comparison failing for a different reason. `akka.management.hostname` was not set, leading to `InetAddress.getLocalHost.getHostAddress` to be used instead, which somehow didn't match the host of the lowest IP returned by discovery, probably because it returned `127.0.0.1`.

Also, the log messages do not reflect what is actually compared, so they are unhelpful and appear contradictory:
```
15:37:39 INFO  HeadlessServiceDnsBootstrap [akka.tcp://x@10.52.2.10:10000/system/headlessServiceDnsBootstrap] - Contact point [http://10.52.2.10:10001] exceeded stable margin with no seed-nodes in sight. Considering whether this node is allowed to JOIN itself to initiate a new cluster.
15:37:39 INFO  HeadlessServiceDnsBootstrap [akka.tcp://x@10.52.2.10:10000/system/headlessServiceDnsBootstrap] - Exceeded stable margins without locating seed-nodes, however this node is NOT the lowest address out of the discovered IPs in this deployment, thus NOT joining self. Expecting node Some(ResolvedTarget(10.52.2.10,Some(10001))) (out of List(ResolvedTarget(10.52.2.10,Some(10001)), ResolvedTarget(10.52.2.18,Some(10001)))) to perform the self-join and initiate the cluster.
```

patriknw:
What shall we do about it? Add a check that the http hostname is the same as the remoting hostname for the self node?

ktoso:
which could be fine... but let's log a warning then?
[...]

PR #242, opened 2018-07-17
Allow an IP address to be used for the lowest "address"

eed3si9n:
Ref #187

#217 started returning “pod-ip-address.my-namespace.pod.cluster.local” in `ResolvedTarget` for Kubernetes API Discovery.

#236 reports that this is in conflict with Cluster Bootstrap when the "self" address is still based on an IP address.

This attempts to provide a fix for the situation by returning both the host name and an optional IP address in `ResolvedTarget`, and letting `LowestAddressJoinDecider` match on either the host or the address.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
@@ -135,12 +135,16 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
               "Bootstrap.selfContactPoint was NOT set! This is required for the bootstrap to work! " +
               "If binding bootstrap routes manually and not via akka-management"))
 
+      // some discovery mechanism can return both host name and IP address. this checks for both.
+      def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
+        (Some(host) == lowest.address) || (host == lowest.host)
+
       // we check if a contact point is "us", by comparing host and port that we've bound to
       def lowestContactPointIsSelfManagement(lowest: ResolvedTarget): Boolean = lowest.port match {
         case None =>
-          selfContactPoints.exists { case (host, _) => host == lowest.host }
+          selfContactPoints.exists { case (host, _) => hostMatches(host, lowest) }
         case Some(lowestPort) =>
-          selfContactPoints.exists { case (host, port) => host == lowest.host && port == lowestPort }
+          selfContactPoints.exists { case (host, port) => hostMatches(host, lowest) && port == lowestPort }
       }
 
       lowestAddressContactPoint(info) match {
--- a/cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/LowestAddressJoinDeciderSpec.scala
+++ b/cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/LowestAddressJoinDeciderSpec.scala
@@ -54,19 +54,24 @@ class LowestAddressJoinDeciderSpec extends WordSpecLike with Matchers with Scala
     val system = ActorSystem("sys", config)
     val settings = ClusterBootstrapSettings(system.settings.config)
 
-    val contactA = ResolvedTarget("127.0.0.1", None)
-    val contactB = ResolvedTarget("b", None)
-    val contactC = ResolvedTarget("c", None)
+    val contactA = ResolvedTarget(host = "127.0.0.1", port = None, address = Some("127.0.0.1"))
+    val contactB = ResolvedTarget(host = "b", port = None, address = None)
+    val contactC = ResolvedTarget(host = "c", port = None, address = None)
 
     "sort ResolvedTarget by lowest hostname:port" in {
-      List(ResolvedTarget("c", None), ResolvedTarget("a", None), ResolvedTarget("b", None)).sorted should ===(
-        List(ResolvedTarget("a", None), ResolvedTarget("b", None), ResolvedTarget("c", None))
+      List(ResolvedTarget("c", None, None), ResolvedTarget("a", None, None),
+        ResolvedTarget("b", None, None)).sorted should ===(
+        List(ResolvedTarget("a", None, None), ResolvedTarget("b", None, None), ResolvedTarget("c", None, None))
       )
-      List(ResolvedTarget("c", Some(1)), ResolvedTarget("a", Some(3)), ResolvedTarget("b", Some(2))).sorted should ===(
-        List(ResolvedTarget("a", Some(3)), ResolvedTarget("b", Some(2)), ResolvedTarget("c", Some(1)))
+      List(ResolvedTarget("c", Some(1), None), ResolvedTarget("a", Some(3), None),
+        ResolvedTarget("b", Some(2), None)).sorted should ===(
+        List(ResolvedTarget("a", Some(3), None), ResolvedTarget("b", Some(2), None),
+          ResolvedTarget("c", Some(1), None))
       )
-      List(ResolvedTarget("a", Some(2)), ResolvedTarget("a", Some(1)), ResolvedTarget("a", Some(3))).sorted should ===(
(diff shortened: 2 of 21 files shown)


ktoso:
Thanks Eugene, reviewing -- have you had the chance to test that this works in a real env you're targeting?

eed3si9n:
@ktoso No, that's why it's (wip) ;)

ktoso:
People use WIP  to mean various things, thus the question to confirm your meaning of it.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -135,12 +135,16 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
               "Bootstrap.selfContactPoint was NOT set! This is required for the bootstrap to work! " +
               "If binding bootstrap routes manually and not via akka-management"))
 
+      // some discovery mechanism can return both host name and IP address. this checks for both.
+      def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
+        (Some(host) == lowest.address) || (host == lowest.host)
```
ktoso: ok 👍

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -41,13 +41,30 @@ object SimpleServiceDiscovery {
     }
   }
 
-  /** Resolved target host, with optional port and protocol spoken */
-  final case class ResolvedTarget(host: String, port: Option[Int]) {
+  /**
+   * Resolved target host, with optional port and the IP address.
+   * The address may be used for cluster bootstrap.
```
ktoso: Should we explain that the address is expected to be an IP while the host may be a hostname or ip?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -41,13 +41,30 @@ object SimpleServiceDiscovery {
...
 
+    def getAddress: Optional[String] = {
+      import scala.compat.java8.OptionConverters._
+      address.asJava
+    }
+
+    override def equals(o: Any): Boolean = o match {
+      case x: ResolvedTarget => (this.host == x.host) && (this.port == x.port)
+      case _ => false
+    }
```
ktoso: Hm... not entirely sure on this, port is part of the sorting after all, so lowest address depends on the port as well:

```
    implicit val addressOrdering: Ordering[ResolvedTarget] = Ordering.fromLessThan[ResolvedTarget] { (a, b) ⇒
      if (a eq b) false
      else if (a.host != b.host) a.host.compareTo(b.host) < 0
      else if (a.port != b.port) a.port.getOrElse(0) < b.port.getOrElse(0)
      else false
    }

```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/LowestAddressJoinDeciderSpec.scala
```diff
@@ -78,9 +83,9 @@ class LowestAddressJoinDeciderSpec extends WordSpecLike with Matchers with Scala
         contactPointsChangedAt = now.minusSeconds(2),
         contactPoints = Set(contactA, contactB, contactC),
         seedNodesObservations = Set(new SeedNodesObservation(now.minusSeconds(1), contactA,
-            Address("akka", "sys", "127.0.0.1", 2552), Set(Address("akka", "sys", "127.0.0.1", 2552))))
+            Address("akka", "sys", "10.0.0.2", 2552), Set(Address("akka", "sys", "10.0.0.2", 2552))))
```
ktoso: Not sure if the change from 127 to 10 is important in this changeset?

eed3si9n:
Yes! I confirmed that chirper friendservice can finally form a cluster on GKE.

pod-m2

### akka.tcp://friendservice@10.0.8.7:10001

```
2018-07-17T04:26:02.132Z [[37minfo[0m] akka.management.cluster.bootstrap.internal.BootstrapCoordinator [sourceThread=friendservice-akka.actor.default-dispatcher-6, akkaTimestamp=04:26:02.132UTC, akkaSource=akka.tcp://friendservice@10.0.8.7:10001/system/bootstrapCoordinator, sourceActorSystem=friendservice] - 
Initiating new cluster, self-joining [akka.tcp://friendservice@10.0.8.7:10001]. Other nodes are expected to locate this cluster via continued contact-point probing.
```

### akka.tcp://friendservice@10.0.9.5:10001

```
2018-07-17T04:26:02.901Z [[37minfo[0m] akka.management.cluster.bootstrap.internal.BootstrapCoordinator [sourceThread=friendservice-akka.actor.default-dispatcher-19, akkaTimestamp=04:26:02.899UTC, akkaSource=akka.tcp://friendservice@10.0.9.5:10001/system/bootstrapCoordinator, sourceActorSystem=friendservice] - 
Joining [akka.tcp://friendservice@10.0.9.5:10001] to existing cluster [akka.tcp://friendservice@10.0.8.7:10001]
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/LowestAddressJoinDeciderSpec.scala
eed3si9n: The test does not hinge on the change of IP address to `10.0.0.2`, but there's a discussion around `127.0.0.1` in #187, so I wanted to make it clear that this PR does NOT address that part of the story.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
eed3si9n: `port` is part of the comparison. `equals` is overridden specifically to exclude `address` out of the comparison.

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/LowestAddressJoinDeciderSpec.scala
ktoso: I'm fine with keeping the change, just that it did not seem directly related to the issue at hand... We try to keep changes minimal and focused.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
ktoso: Hmm right, this should be fine

ktoso:
Excellent, thanks Eugene :)

ktoso:
Do you have some ideas how we could make checking this an automated test somehow? 
I'm always worried with the bootstrap changes that we don't have built-in but rely on manual tests of those..

ktoso:
Resolves https://github.com/akka/akka-management/issues/187

eed3si9n:
> Do you have some ideas how we could make checking this an automated test somehow?
I'm always worried with the bootstrap changes that we don't have built-in but rely on manual tests of those..

Not sure how tangled the dependencies will be, but some kind of Jenkins job that deploys chirper using the freshly-baked akka-management and/or rp-cli might be doable?

ktoso:
Yes though the question is where to it could do that - seems you have some account on GCloud now?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
dwijnand: "expected"? 😕 It would be better if this were "enforced", i.e with a specific type.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
dwijnand: Why is `address` specifically excluded? If two resolved target hosts for the same hostname resolve to different ip addresses, are they equal?

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
ktoso: Good point, could you follow up on this @eed3si9n and do the comment rewording as well as I requested there?

PR #243, opened 2018-07-17
Use InetAddress type to denote IPv4 address

eed3si9n:
This is a refinement on #242 a part of a fix towards #187

This brings `apply(String, Option[Int])` back as a factory method for `ResolvedTarget`, and changes the type of `address` to `java.net.InetAddress`.

Since various orchestration engines just return a String for "host", I am doing a quick regex check to make sure it looks like an IPv4 address, and then parsing it into an
`InetAddress`.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
@@ -137,7 +137,7 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
 
       // some discovery mechanism can return both host name and IP address. this checks for both.
       def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
-        (Some(host) == lowest.address) || (host == lowest.host)
+        (Some(host) == lowest.address.map(_.getHostAddress)) || (host == lowest.host)
 
       // we check if a contact point is "us", by comparing host and port that we've bound to
       def lowestContactPointIsSelfManagement(lowest: ResolvedTarget): Boolean = lowest.port match {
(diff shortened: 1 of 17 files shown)


discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -39,37 +40,40 @@ object SimpleServiceDiscovery {
...
+      }
+    }
   }
 
   /**
    * Resolved target host, with optional port and the IP address.
-   * The address may be used for cluster bootstrap.
+   * @param host the hostname or the IP address of the target
+   * @param port optional port number
+   * @param address optional IP address of the target. This is used during cluster bootstap when available.
```
eed3si9n: @ktoso @dwijnand Updated the Scaladoc.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -137,7 +137,7 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
 
       // some discovery mechanism can return both host name and IP address. this checks for both.
       def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
-        (Some(host) == lowest.address) || (host == lowest.host)
+        (Some(host) == lowest.address.map(_.getHostAddress)) || (host == lowest.host)
```
ktoso: since now this has an op let's inverse them

`(host == lowest.host) || ...`

eed3si9n: ok.

eed3si9n: done c5b633036f229e172c064fb370e35b2a7762cf7b

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
ktoso: thanks, good :)

PR #252, opened 2018-07-25
Clarify the way to start ClusterHTTPManagement

### LINKED ISSUE #232: better documentation of how to use/enable Cluster Http Management
https://developer.lightbend.com/docs/akka-management/current/cluster-http-management.html
doesn't really tell what needs to be done to start it

It's probably via the `AkkaManagement(system).start()` that is described in another page
https://developer.lightbend.com/docs/akka-management/current/akka-management.html
but we should describe this more explicitly.

Also, I can't see any mention of `ClusterHttpManagementRoutes` if you want to include it in an existing http server.

hythloday:
I got sufficiently stuck on this today that I thought it was worth writing docs for.

(Partially?) fixes #232

### DIFF
--- a/cluster-http/src/test/scala/doc/akka/cluster/http/management/CompileOnlySpec.scala
+++ b/cluster-http/src/test/scala/doc/akka/cluster/http/management/CompileOnlySpec.scala
@@ -1,0 +1,15 @@
+/*
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package doc.akka.cluster.http.management
+
+import akka.actor.ActorSystem
+import akka.management.cluster.ClusterHttpManagement
+
+object CompileOnlySpec {
+
+  //#loading
+  val system = ActorSystem()
+  val httpMgmt = ClusterHttpManagement(system)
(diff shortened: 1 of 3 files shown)


cluster-http/src/test/scala/doc/akka/cluster/http/management/CompileOnlySpec.scala
```diff
@@ -0,0 +1,15 @@
...
+package doc.akka.cluster.http.management
+
+import akka.actor.ActorSystem
+import akka.management.cluster.ClusterHttpManagement
+
+object CompileOnlySpec {
+
+  //#loading
+  val system = ActorSystem()
+  val httpMgmt = ClusterHttpManagement(system)
```
patriknw: Is this really correct? It might work, but I thought the intended usage would be 
```
AkkaManagement(system).start()
```

which would load from reference.conf
```
```

hythloday: That makes a great deal more sense. Would you like me to submit a PR?

PR #253, opened 2018-07-28
EC2 Tag Based Discovery : allow running multiple Akka nodes per EC2 instance

sebastianharko:
This simply removes an existing limitation (it now allows running multiple Akka nodes (i.e. JVMs) per EC2 instance).

**Note**: this was tested via the integration script and I'm reasonable sure that it doesn't break existing functionality.


docs/src/main/paradox/discovery.md
```diff
@@ -460,8 +460,20 @@ character. For example:
...
+    akka {
+      discovery {
+        aws-api-ec2-tag-based {
+          ports = [8557, 8558, 8559] # 3 Akka nodes per EC2 instance
+          # note that the above need to be the ports associated with the *Akka Management* extension
+        }
+      }
+    }
+    ```
+    Note: this comes with the limitation that each EC2 instance has to have the same number of Akka nodes.
```
sebastianharko: Most probably still works even in that case actually... but you may see some error/warn msgs in the logs ...

sebastianharko:
anyone interested in reviewing this PR ? @ktoso @raboof

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -22,15 +23,15 @@ import scala.collection.immutable.Seq
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ ExecutionContext, Future }
 
-object Ec2TagBasedSimpleServiceDiscovery {
+@InternalApi private[ec2] object Ec2TagBasedSimpleServiceDiscovery {
```
ktoso: also a comment `/** INTERNAL API */`

sebastianharko:
@ktoso Thanks. It's definitely a more exotic use case though it has value . I can't add a test right now but definitely in the following iteration :-) There's an existing integration test that ensures that this addition does not break down the existing functionality BTW. Let me know if this can be merged. Cheers!

Issue #156, opened 2018-04-02
Possible split-brain of Akka Cluster using K8s Deployment


notxcain:
I want to describe a possible split-brain scenario (even with SBR enabled) when using K8s Deployment for Akka Cluster.

## How `Deployment` works
1. `Deployment` assumes it's replicas are stateless and that it's safe to have more of them than configured by `replicas` parameter
2. Kubernetes controller manager will try to check `nodeStatusUpdateRetry` times every `--node-monitor-period` of time. After `--node-monitor-grace-period` it will consider node unhealthy. It will remove its pods after `--pod-eviction-timeout`

notxcain:
Ooops, accidentally hit `Ctrl+Enter` - WIP

ktoso:
Hi @notxcain, your writeup is a bit weird... specifically:

> Eventually K8s starts 3 containers on reachable nodes, they form cluster. Now we have 2 clusters.

Why 3 new ones? 
Why would they form a cluster?

>What about StatefulSet?

Why not stateful set is documented here: https://github.com/akka/akka-management/issues/134
Please give it a look and correct us if you think that analysis is incorrect.

notxcain:
> Why 3 new ones?
One member left on reachable worker will kill itself. K8s will restart it. Also k8s will start evicted pods on reachable nodes. At this moment there will be 5 running members.

> Why would they form a cluster?
They would form a cluster because of how cluster-bootstrap works. Members left on unreachable nodes won't be listed in headless service DNS records, so there gonna be no evidence of their existence.

> Please give it a look and correct us if you think that analysis is incorrect.
I'll take a closer look later.

ktoso:
> One member left on reachable worker will kill itself. K8s will restart it. Also k8s will start evicted pods on reachable nodes. At this moment there will be 5 running members.

I have a feeling wording is not quite precise here... Could you please explain the exact steps we are worried about and how it relates to "akka nodes" "pods" "physical nodes" etc?

>      They would form a cluster because of how cluster-bootstrap works. 

Again I feel you're making a huge mental leap over all the details between. Which nodes and for what time are visible in discovery via k8s APIs etc plays a huge role in these things. I don't see these taken into consideration in this ticket so far?

notxcain:
> I have a feeling wording is not quite precise here... Could you please explain the exact steps we are worried about and how it relates to "akka nodes" "pods" "physical nodes" etc?

akka member == pod
worker == physical node

[...]

PR #254, opened 2018-07-29
Bootstrap: should be done only with reachable nodes that are "up", "weaklyUp" or "joining"

thomschke:
Akka Cluster Bootstrap should response only reachable nodes that are in status "up".


ktoso:
I think that's a good idea but we should add a test for this, would you be up for this?

raboof:
wouldn't that prevent `min-nr-of-members` from working?

The initial seed node would join itself, but not move itself to 'Up' because `min-nr-of-members` would not be met. However, that would then prevent any other nodes from joining, so `min-nr-of-members` would never be reached?

patriknw:
Why would only Up nodes be used?
I agree with the concerns raised by @raboof ?

markusjura:
I think @thomschke created the PR based on the issue https://github.com/akka/akka-management/issues/156

thomschke:
Yes, 

- joining unreachable seed-nodes during k8s rolling updates was the intention for this PR
- [epidemic joining](https://developer.lightbend.com/docs/akka-management/current/bootstrap.html#case-1-bootstrap-process) can be prevented if contact point probing returns only "up" members. That's why contact point probing should return "weaklyUp" members too.

thomschke:
Due to `akka.cluster.allow-weakly-up-members = off` contact point probing should return members whose status is "up", "weaklyUp" or "joining".

PR #258, opened 2018-08-03
Update akka-http to 10.1.3

snovgorodtsev:


snovgorodtsev:
@ktoso 
Thank you! Just in case I'll post a problem I experienced with `akka-http 10.0.13` here.
When I call `akka.management.AkkaManagement#stop` in my project it always fails with `java.lang.ClassCastException: akka.Done$ cannot be cast to scala.runtime.BoxedUnit`. In the project I use `10.1.1` and I can not downgrade to `10.0.x`, but I did it just for the sake of experiment and off course it helped me out.

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -165,7 +165,7 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
     if (binding == null) {
       Future.successful(Done)
     } else if (bindingFuture.compareAndSet(binding, null)) {
-      val stopFuture = binding.flatMap(_.unbind()).map(_ => Done)
+      val stopFuture = binding.flatMap(_.unbind())
```
raboof: Is this the place that triggered the `java.lang.ClassCastException: akka.Done$ cannot be cast to scala.runtime.BoxedUnit`? That's a bit of a bummer... perhaps we should change it to `.map{ (_: Object) => Done }` to be on the safe side?

snovgorodtsev: >Is this the place that triggered the `java.lang.ClassCastException: akka.Done$ cannot be cast to scala.runtime.BoxedUnit`?

Yes it is!

>perhaps we should change it to .map{ (_: Object) => Done } to be on the safe side?

Yeah I hope `(_ : Object)` potentially could solve the issue for `10.0.x` and for `10.1.x` we can just remove the whole `.map` part as I did.

project/Dependencies.scala
```diff
@@ -4,7 +4,7 @@ import Keys._
 object Dependencies {
 
   val AkkaVersion = "2.5.14"
-  val AkkaHttpVersion = "10.0.13" // TODO update to 10.1.x as soon as possible
+  val AkkaHttpVersion = "10.1.3"
```
patriknw: This means that we will not support using Akka Management with applications that are still using Akka HTTP 10.0.x. I think that is fine since it's time to more aggressively phase out 10.0 anyway, but wanted to highlight that so it doesn't come as surprise after the fact for anyone.

TimMoore: Keep in mind that the current stable versions of Play and Lagom still use 10.0. A lot of Lagom users especially will have this dependency brought in by Lightbend Orchestration or Cluster Bootstrap.

I'm not sure if it would cause problems to evict 10.0 in these situations (I don't think we've seen any evidence that it would) but it won't be tested thoroughly in this combination, and users might not even be aware that it's happening.

management/src/main/scala/akka/management/AkkaManagement.scala
patriknw: This is because of the binary incompatible change in https://github.com/akka/akka-http/pull/1908

project/Dependencies.scala
patriknw: Then we should stay on 10.0.13 and try to work around then bincompat issue

raboof:
To make @TimMoore's comment above more prominent: the current stable versions of Play and Lagom still use 10.0. A lot of Lagom users especially will have the akka-management dependency brought in by Lightbend Orchestration or Cluster Bootstrap. For that reason it seems prudent to stay on 10.0.x until those projects have made the switch.

ktoso:
I agree, let's stay on 10.0 here for now.

TimMoore:
Thanks, @raboof. As additional clarification, Play 2.7 and Lagom 1.5 are the versions currently in development on `master` of each project, and have already updated Akka HTTP to 10.1.x. Once those are released and become the current stable versions, I think it's OK to update here too.

dwijnand:
@TimMoore When those versions are released are they instantly the "current stable versions" or does that happen some time later? Also, when are they individually scheduled for release? (here or privately 🙂)

TimMoore:
Yeah, this is probably not the right place for the discussion… :smile:

management/src/main/scala/akka/management/AkkaManagement.scala
jrudolph: That's indeed a bummer and we should make the change as hinted above to make it compatible with 10.0 if we make another release before updating to akka-http 10.1.x.

dwijnand: Done: https://github.com/akka/akka-management/pull/280

jrudolph: Ah, cool. What we can also do it akka-http is just returning `null` instead of `Done` to rule out problems with third-party code.

PR #274, opened 2018-08-22
Bootstrap support for portName and protocol

chbatey:
Required for SRV records

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
@@ -17,7 +17,7 @@ import akka.actor.Status.Failure
 import akka.actor.Timers
 import akka.annotation.InternalApi
 import akka.cluster.Cluster
-import akka.discovery.SimpleServiceDiscovery
+import akka.discovery.{ Lookup, SimpleServiceDiscovery }
 import akka.discovery.SimpleServiceDiscovery.ResolvedTarget
 import akka.http.scaladsl.model.Uri
 import akka.management.cluster.bootstrap.ClusterBootstrapSettings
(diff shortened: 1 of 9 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -274,7 +275,7 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
     val baseUri = settings.managementBasePath.fold(rawBaseUri)(prefix => rawBaseUri.withPath(Uri.Path(s"/$prefix")))
 
     val childActorName = s"contactPointProbe-${baseUri.authority.host}-${baseUri.authority.port}"
-    log.info("Ensuring probing actor: " + childActorName)
+    log.debug("Ensuring probing actor: " + childActorName)
```
chbatey: This log message isn't useful at Info to users

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -171,10 +172,8 @@ private[akka] final class BootstrapCoordinator(discovery: SimpleServiceDiscovery
       // the next round of discovery will be performed once this one returns
       discoverContactPoints()
 
-    case SimpleServiceDiscovery.Resolved(name, contactPoints) ⇒
-      serviceName = name
-
-      log.info("Located service members with name: [{}]: [{}]", serviceName, contactPoints.mkString(", "))
+    case SimpleServiceDiscovery.Resolved(_, contactPoints) ⇒
+      log.info("Located service members with name: [{}]: [{}]", lookup, contactPoints.mkString(", "))
```
raboof: Perhaps make the message something like "Looked up service members based on [{}]: [{}]"?

PR #275, opened 2018-08-22
Additional records

chbatey:
Also includes commits from https://github.com/akka/akka-management/pull/274#pullrequestreview-148442323

### DIFF
--- a/discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
+++ b/discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
@@ -3,6 +3,8 @@
  */
 package akka.discovery.dns
 
+import java.net.InetAddress
+
 import akka.AkkaVersion
 import akka.actor.ActorSystem
 import akka.event.Logging
@@ -20,8 +22,8 @@ import akka.io.dns.DnsProtocol.{ Ip, Srv }
  */
 class DnsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
-  // Required for async dns
-  AkkaVersion.require("discovery-dns", "2.5.14")
+  // Required for async dns, 15 for additional records
+  AkkaVersion.require("discovery-dns", "2.5.15")
   require(system.settings.config.getString("akka.io.dns.resolver") == "async-dns",
     "Akka discovery DNS requires akka.io.dns.resolver to be set to async-dns")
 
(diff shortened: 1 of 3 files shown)


chbatey:
2.5.15 released now :)

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -3,25 +3,60 @@
...
  * Looks for A records for a given service.
  */
 class DnsSimpleServiceDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
 
-  // Required for async dns
-  AkkaVersion.require("discovery-dns", "2.5.14")
+  import DnsSimpleServiceDiscovery._
+
+  // Required for async dns, 15 for additional records
+  AkkaVersion.require("discovery-dns", "2.5.15")
```
ktoso: 👍

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -3,25 +3,60 @@
...
+        } else {
+          addresses
+        }
+      case other => im.Seq.empty[ResolvedTarget]
+    }
+
+    Resolved(srvRequest, addresses)
+  }
+
+}
```
ktoso: 👍

project/Dependencies.scala
```diff
@@ -3,9 +3,8 @@ import Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.5.14"
-  // TODO #281 update to 10.1.x when play and lagom are released
-  val AkkaHttpVersion = "10.0.13"
+  val AkkaVersion = "2.5.15"
+  val AkkaHttpVersion = "10.0.13" // TODO update to 10.1.x as soon as possible
```
ktoso: please keep the other note: `   // TODO #281 update to 10.1.x when play and lagom are released `

chbatey: Merge mistake, will add back

discovery-dns/src/main/scala/akka/discovery/dns/DnsSimpleServiceDiscovery.scala
```diff
@@ -3,25 +3,44 @@
...
-import akka.io.dns.{ AAAARecord, ARecord, DnsProtocol, SRVRecord }
 import akka.io.dns.DnsProtocol.{ Ip, Srv }
+import akka.io.dns.{ AAAARecord, ARecord, DnsProtocol, SRVRecord }
+
+object DnsSimpleServiceDiscovery {
+  def srvRecordsToResolved(srvRequest: String, resolved: DnsProtocol.Resolved): Resolved = {
+    val ips: Map[String, InetAddress] = resolved.additionalRecords.collect {
+      case a: ARecord => a.name -> a.ip
+      case aaaa: AAAARecord => aaaa.name -> aaaa.ip
+    }.toMap
```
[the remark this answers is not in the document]
raboof: Seems elegant :+1: - so this way it's up to the consumer of the API what to do when the IP address is `None`. Makes sense.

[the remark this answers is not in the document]
chbatey: Yeah which could be to resolve the target some other way. Which is what most db/http clients will do if they pass that into them

PR #278, opened 2018-08-22
K8s bootstrap docs

chbatey:
I'll update the demon as well but thought it is worth getting @TimMoore @raboof @ktoso thoughts on this so I don't need to change it again :)


docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+      protocol: TCP
+      targetPort: 8558
+    - name: remoting
+      port: 2552
+      protocol: TCP
+      targetPort: 2552
+  selector:
+    appName: "akka-cluster-kubernetes"
+  clusterIP: None
+  publishNotReadyAddresses: true
```
chbatey: @TimMoore this is the property for publishing DNS entries for not ready

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
+  name: "akka-cluster-kubernetes"
+spec:
+  ports:
+    - name: management
+      port: 8558
+      protocol: TCP
+      targetPort: 8558
+    - name: remoting
```
chbatey: this doesn't actually need to be on here as the remoting port is discovered via bootstrap

docs/src/main/paradox/bootstrap/kubernetes.md
ktoso: I see, nice! Where's that in docs?

ktoso: Thought that's an url but isn't hm http://service.alpha.kubernetes.io/tolerate-unready-endpoints

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+      protocol: TCP
+      targetPort: 2552
+  selector:
+    appName: "akka-cluster-kubernetes"
+  clusterIP: None
+  publishNotReadyAddresses: true
+```
+
+Note there are currently two ways to set that addressed should be published if not ready, the initial way via an annotation 
+`service.alpha.kubernetes.io/tolerate-unready-endpoints` and via the new officially supported way as the property `publishNotReadyAddresses`.
```
ktoso: link to its docs?

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+```
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    appName: "akka-cluster-kubernetes"
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
```
ktoso: really "true" or true would work too? for my info

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+  private val aliveStates: Set[MemberStatus] = Set(MemberStatus.Joining, MemberStatus.WeaklyUp, MemberStatus.Up, MemberStatus.Leaving, MemberStatus.Exiting)
+
+  val k8sHealthChecks: Route =
+    path("ready") {
+      get {
+        val selfState = cluster.selfMember.status
+        if (readyStates.contains(selfState)) complete(StatusCodes.OK)
+        else complete(StatusCodes.InternalServerError)
+      }
+    } ~
```
ktoso: please use `concat()`

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
+# Kubernetes
+
+As of Akka 2.5.15 and Akka-Management 0.18 the recommended way to run Akka Cluster in Kubernetes is to:
+
+* Use Akka Bootstrap with `akka-dns` with cluster formation via DNS SRV records
+* Use a headless service for internal for Akka management/bootstrap so that readiness probes for prod traffic don't interfere with bootstrap 
+* If required use a separate service and/or ingress for user facing endpoints e.g. HTTP gRPC
```
raboof: Maybe add some more punctiation, like "ingress for user-facing endpoints, for example HTTP or gRPC"? Maybe add links to akka-http and akka-grpc? Do we know if ingress works for gRPC?

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+      port: 2552
+      protocol: TCP
+      targetPort: 2552
+  selector:
+    appName: "akka-cluster-kubernetes"
+  clusterIP: None
+  publishNotReadyAddresses: true
+```
+
+Note there are currently two ways to set that addressed should be published if not ready, the initial way via an annotation
```
raboof: maybe 'specify' rather than 'set'?

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+      targetPort: 2552
+  selector:
+    appName: "akka-cluster-kubernetes"
+  clusterIP: None
+  publishNotReadyAddresses: true
+```
+
+Note there are currently two ways to set that addressed should be published if not ready, the initial way via an annotation 
+`service.alpha.kubernetes.io/tolerate-unready-endpoints` and via the new officially supported way as the property `publishNotReadyAddresses`.
+Set both as depending on your DNS solution it may have not migrated from the annotation to the property.
```
raboof: Maybe record which versions of common DNS solutions are updated to understand the 'new' way?

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+
+  discovery {
+    method = akka-dns
+  }
+}
+```
+
+The same configuration will work for any environment that has an SRV record for your Akka Clustered application. 
+
+For prod traffic e.g. HTTP use a regular service or an alternative ingress mechanism.
```
raboof: Maybe add 2 subsection headers in this section ('Headless service' and 'Exposed service'?)

docs/src/main/paradox/bootstrap/kubernetes.md
```diff
@@ -0,0 +1,149 @@
...
+  discovery {
+    method = akka-dns
+  }
+}
+```
+
+The same configuration will work for any environment that has an SRV record for your Akka Clustered application. 
+
+For prod traffic e.g. HTTP use a regular service or an alternative ingress mechanism. 
+With an appropriate readiness check this results in traffic not being routed until bootstrap has finished.
```
raboof: :+1:

docs/src/main/paradox/bootstrap/kubernetes.md
chbatey: Only in the swagger for the api server from what I can see. 

I found it via its github issue.

docs/src/main/paradox/bootstrap/kubernetes.md
chbatey: Ah but that would require it to be documented :)

docs/src/main/paradox/bootstrap/kubernetes.md
chbatey: Annotations have to be strings :(

If you have it as true, `kubectl create` tells you it is invalid and `kubectl apply` ignores it so it doesn't get set (reported bug)

docs/src/main/paradox/bootstrap/kubernetes.md
chbatey: I haven't tried ingress for gRPC only internal comms. It'll depend on what is your ingress controller, most of them are nginx

docs/src/main/paradox/bootstrap/kubernetes.md
chbatey: :+1:

chbatey:
This won't build until 2.5.15 is released.

I've updated the docs to take the yaml / config from a working demo that i've added to this project.

I've also restructured the demos so they don't get up with such long names when building docker images for them

project/Dependencies.scala
```diff
@@ -3,7 +3,7 @@ import Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.5.14"
+  val AkkaVersion = "2.5-SNAPSHOT"
```
chbatey: Will update this to 2.5.15 when released

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
```diff
@@ -0,0 +1,82 @@
...
+  publishNotReadyAddresses: true
+#headless
+---
+#public
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    appName: "bootstrap-demo-kubernetes-dns"
+  name: "bootstrap-demo-kubernetes-dns-public"
```
raboof: maybe remove the `-public` suffix here and add a `-headless` or `-akka-internal` or something similar to the headless service?

markusjura:
Very nice solution and writeup 👍

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
chbatey: Yep good idea

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+Health checks can be used check a node is part of a cluster e.g.
+
+@@snip [health-checks]($management$/bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala)  { #health }
+
+This will mean that a pod won't get traffic until it is part of a cluster which is important
+if either `ClusterSharding` or `ClusterSingleton` are used.
+
+## Services
+
+## Internal headless service
```
raboof: Should be a `###` heading?

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+* Use a headless service for internal and Akka management/bootstrap so that readiness probes for prod traffic don't interfere with bootstrap 
+* If required use a separate service and/or ingress for user-facing endpoints, for example [HTTP](https://doc.akka.io/docs/akka-http/current/) or [gRPC](https://developer.lightbend.com/docs/akka-grpc/current/)
+
+## Deployments
+
+Use a regular deployment (not a StatefulSet). 
+
+### Update strategy
+
+For small clusters it may make sense to set `maxUnabailable` to 0 and `maxSurge` to 1.
```
raboof: typo

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+
+As of Akka 2.5.15 and Akka-Management 0.18 the recommended way to run Akka Cluster in Kubernetes is to:
+
+* Use Akka Bootstrap with `akka-dns` with cluster formation via DNS SRV records
+* Use a headless service for internal and Akka management/bootstrap so that readiness probes for prod traffic don't interfere with bootstrap 
+* If required use a separate service and/or ingress for user-facing endpoints, for example [HTTP](https://doc.akka.io/docs/akka-http/current/) or [gRPC](https://developer.lightbend.com/docs/akka-grpc/current/)
+
+## Deployments
+
+Use a regular deployment (not a StatefulSet).
```
raboof: Perhaps explain why?

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+with cluster singletons is typically removed last and then the cluster singletons move the the oldest new pod.  
+
+
+### Health checks
+
+Health checks can be used check a node is part of a cluster e.g.
+
+@@snip [health-checks]($management$/bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala)  { #health }
+
+This will mean that a pod won't get traffic until it is part of a cluster which is important
```
raboof: Maybe add a comma, 'until it is part of a cluster, which is ...'

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
+# Kubernetes
+
+A full working example that can be deployed to `minikube` is in `bootstrap-demo/kuvernetes-dns`.
```
raboof: kuvernetes, va?

bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala
```diff
@@ -0,0 +1,42 @@
...
+
+class KubernetesHealthChecks(system: ActorSystem) {
+
+  val log: LoggingAdapter = Logging(system, getClass)
+  val cluster = Cluster(system)
+
+  //#health
+  private val readyStates: Set[MemberStatus] = Set(MemberStatus.Up, MemberStatus.WeaklyUp)
+  private val aliveStates: Set[MemberStatus] =
+    Set(MemberStatus.Joining, MemberStatus.WeaklyUp, MemberStatus.Up, MemberStatus.Leaving, MemberStatus.Exiting)
```
raboof: Hm, indeed we don't want K8s to kill the container while it is leaving/exiting, OTOH we want K8s to start removing the node from the loadbalancer, since otherwise there could be a window in which it still might get traffic while it has been moved to `Down`, right?

When a node is no longer 'alive', shouldn't K8s keep in account a grace period, and even when it decides to kill the node, keep a healthy delay between `SIGTERM` and `SIGKILL`?

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
```diff
@@ -0,0 +1,88 @@
...
+    appName: "bootstrap-demo-kubernetes-dns"
+  name: "bootstrap-demo-kubernetes-dns"
+spec:
+  ports:
+  - name: http
+    port: 8080
+    protocol: TCP
+    targetPort: 8080
+  selector:
+    appName: "bootstrap-demo-kubernetes-dns"
```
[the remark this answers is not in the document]
chbatey: :+1:

bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala
chbatey: I believe that is what readiness is for, k8s keeps hitting the readiness probe which will only succeeds in Up and WeeklyUp. So we can be alive but not ready when shutting down

raboof: Ah yes I incorrectly assumed 'ready' would stop polling as soon as the pod has become ready. I agree this is :+1: then - we'd only need to make sure nodes aren't downed faster than the `readinessProbe` `periodSeconds` plus some buffer.

TimMoore:
Sorry for the delay. Looks good to me, too. I think it's OK to go ahead and merge.

I'll also summon @eed3si9n, @cunei, @renatocaval and @dwijnand so they're aware of it. It's relevant to what they're working on.

docs/src/main/paradox/bootstrap/demos/kubernetes-api.md
```diff
@@ -54,7 +55,7 @@ $ sbt shell
 
 Once the image is published, you can deploy it onto the kubernetes cluster by calling:
 
-@@snip [kube-create.sh](../../../../../bootstrap-joining-demo/kubernetes-api/kube-create.sh) 
+@@snip [kube-create.sh]($management$/bootstrap-demo/kubernetes-api/kube-create.sh) 
 
 This will create and start running a number of Pods hosting the application. The application nodes will proceed with 
 forming the cluster using the DNS bootstrap method. In order to observe the logs during the cluster formation you can
```
TimMoore: Just noticed that this page talks about the DNS bootstrap method, but it's supposed to be the Kubernetes API demo.

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
+# Kubernetes
+
+A full working example that can be deployed to kubernetes via `minikube` is in `bootstrap-demo/kuvernetes-dns`.
```
TimMoore: small typo: kuvernetes

docs/src/main/paradox/bootstrap/demos/index.md
```diff
@@ -0,0 +1,4 @@
+# Bootstrap examples
```
TimMoore: This page could be hard to find, because it's not listed at the top-level menu, only linked from the bottom of the bootstrap page.

docs/src/main/paradox/bootstrap/demos/kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+
+As of Akka 2.5.15 and Akka-Management 0.18 the recommended way to run Akka Cluster in Kubernetes is to:
```
TimMoore: Maybe update to 2.5.16 now

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
```diff
@@ -0,0 +1,89 @@
...
+      volumes: []
+---
+#headless
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    appName: "bootstrap-demo-kubernetes-dns"
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
```
longmuir: It looks like this annotation will be removed in Kubernetes 1.13 (https://github.com/kubernetes/kubernetes/issues/63741) to be replaced by `Service.spec.publishNotReadyAddresses` (https://github.com/kubernetes/kubernetes/commit/944e07480fa4066fb1345cd1a81d74b3bbb78440). As a general rule, I don't think we should be telling people to use Alpha K8s features. Should we wait until this is sorted out in K8s before releasing?

longmuir:
See note about `service.alpha.kubernetes.io/tolerate-unready-endpoints annotation` being deprecated and shortly removed. I'm wondering if we should hold off on this until Kubernetes 1.13 is released with https://github.com/kubernetes/kubernetes/commit/944e07480fa4066fb1345cd1a81d74b3bbb78440. The challenge with recommending this, is that when the change happens each individual user/customer will have to adjust their K8s configs, and it may not be clear why failures are happening. I'm also not clear if this annotation is supported on OpenShift.

There may have to be different recommendations depending on the version or packaging of K8s that people are using.

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
raboof: In the docs at https://github.com/akka/akka-management/pull/278/files#diff-e737933b9744b8db3a71650e02a1e627R50 we recommend adding both to avoid problems with different versions/assumptions - that seems reasonably safe?

chbatey:
It’s an alpha feature which has made it into main line

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
ktoso: I don't think we know; would have to be looked into. I'll mark it as an action item though.
@chbatey is now laser focused on shipping another thing for a customer, we'll get his attention later on to this.

raboof:
I think @longmuir makes some great points above and we should definitely follow up and test some of our assumptions.

On the other hand I think the content in the PR could already be extremely valuable for those situations where this annotation/setting is available.

Should we perhaps rephrase some of the introduction here, warn that it is not widely tested and that this relatively-new feature is required, and merge this PR relatively quickly but without immediately elevating this approach to 'The Recommended Way'?

chbatey:
+1 I’ll update this next week as saying this is one way

chbatey:
Updated saying they are alternatives. I think the kubernetes api demo needs some work as it mentions dns created https://github.com/akka/akka-management/issues/292 to follow up

docs/src/main/paradox/bootstrap/demos/kubernetes-api.md
chbatey: Raised  https://github.com/akka/akka-management/issues/292 to track making that example purely k8s API

PR #285, opened 2018-09-03
Bump akka-http to 10.1.4

dsebban:


lightbend-cla-validator:
Hi @dsebban,

ktoso:
Thanks for the PR but we won't bump like this in a minor version since it would break other projects.
See https://github.com/akka/akka-management/pull/258#discussion_r210177457 for details.

If there's any issues about doing so, these should be addressed, not the other way around.

PR #288, opened 2018-09-05
Update AkkaHttpVersion to 10.1.4

dmi3zkm:


ktoso:
Thanks for the PR but we won't bump like this in a minor version since it would break other projects.
See https://github.com/akka/akka-management/pull/258#discussion_r210177457 for details.

ktoso:
You should be able to use bootstrap etc by bumping the dependency in _your_ projects of akka http to the 10.1 series. There's one issue about it currently and it will be resolved.

PR #304, opened 2018-09-19
Improve Marathon API logging to see where and how seed nodes have been requested.

ygree:
This is driven by the real use case that took lots of time to debug. This setting was set during automatic deployment process and was pointing to the wrong Marathon master. It led to the situation when the app was joining different Akka Cluster from completely different environment.


discovery-marathon-api/src/main/scala/akka/discovery/marathon/MarathonApiSimpleServiceDiscovery.scala
```diff
@@ -79,6 +79,8 @@ class MarathonApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
 
     val request = HttpRequest(uri = uri)
 
+    system.log.info("Requesting seed nodes by: {}", request.uri)
```
ktoso: Good however please using a logger made like this:

`val log = Logging(getClass, system)` since we should use the "right logger"

ygree: Fixed. Thanks, @ktoso.

PR #308, opened 2018-09-21
added some buffer for k8s health check example and dependency info fo…

alanngai:
added some padding to k8s health check examples.  I've found that in some cases when trying to spin up an akka cluster, the live and readiness checks fail too quickly because they kick in as soon as the pod is scheduled.  Since the jvm takes some time to start up, i found much better results in delaying the checks for a little bit.  I'm recommending including this in the example.  Here's a before and after comparison

image

image

Also a couple more minor changes
- annotated documentation to clarify that bootstrap w/dns requires akka 2.5.14 and above
- added k8s service discovery example in the dependency doc since dns isn't an option yet for our Lagom users (according to @TimMoore  an update to reactivelib is needed before that can be used)

### DIFF
--- a/bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
+++ b/bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
@@ -32,8 +32,6 @@ spec:
           httpGet:
             path: /ready
             port: 8080
-          initialDelaySeconds: 30
-          periodSeconds: 15
         livenessProbe:
           httpGet:
             path: /alive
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @alanngai,

bootstrap-demo/kubernetes-dns/kubernetes/akka-cluster.yml
```diff
@@ -32,10 +32,14 @@ spec:
           httpGet:
             path: /ready
             port: 8080
+          initialDelaySeconds: 30
```
chbatey: Is it needed for readiness? That doesn't cause the node to be killed so no harm in probing right away

alanngai: I suppose it's not strictly necessary, but practically speaking if the cluster isn't stable enough where it can't say "I am alive" successfully, it doesn't seem like a good idea for it to say "I am ready for traffic."  For my own applications, I'd keep it in there.  But I can remove it if you wish for this example

raboof:
I suspect this is fixed by #332, please re-open if not

PR #310, opened 2018-09-21
Warn when 'self' is not found in contact points

### LINKED ISSUE #189: Fail when 'self' is not included in the service discovery
When service discovery does not discover the 'current' node (but none at all or only 'peers'), this seems to point to a misconfiguration (#187, https://github.com/akka/akka-management/issues/183#issuecomment-385932241).

Should we (retry and) fail in such a situation?

raboof:
fixes #189

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
@@ -139,17 +139,25 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
       def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
         (host == lowest.host) || (Some(host) == lowest.address.map(_.getHostAddress))
 
-      // we check if a contact point is "us", by comparing host and port that we've bound to
-      def lowestContactPointIsSelfManagement(lowest: ResolvedTarget): Boolean = lowest.port match {
+      def matchesSelf(target: ResolvedTarget): Boolean = target.port match {
         case None =>
-          selfContactPoints.exists { case (host, _) => hostMatches(host, lowest) }
+          selfContactPoints.exists { case (host, _) => hostMatches(host, target) }
         case Some(lowestPort) =>
-          selfContactPoints.exists { case (host, port) => hostMatches(host, lowest) && port == lowestPort }
+          selfContactPoints.exists { case (host, port) => hostMatches(host, target) && port == lowestPort }
       }
 
       lowestAddressContactPoint(info) match {
-        case Some(lowest) => lowestContactPointIsSelfManagement(lowest)
-        case None => false
+        case Some(lowest) =>
+          if (matchesSelf(lowest)) true
+          else {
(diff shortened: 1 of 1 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -139,17 +139,25 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
       def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
         (host == lowest.host) || (Some(host) == lowest.address.map(_.getHostAddress))
 
-      // we check if a contact point is "us", by comparing host and port that we've bound to
-      def lowestContactPointIsSelfManagement(lowest: ResolvedTarget): Boolean = lowest.port match {
+      def matchesSelf(target: ResolvedTarget): Boolean = target.port match {
         case None =>
-          selfContactPoints.exists { case (host, _) => hostMatches(host, lowest) }
+          selfContactPoints.exists { case (host, _) => hostMatches(host, target) }
```
chbatey: Not this PR but very odd that selfContactPoints refers directly to Kubernetes API configuration. Users may have that but not be using kubernetes api for boostrap

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -139,17 +139,25 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
       def hostMatches(host: String, lowest: ResolvedTarget): Boolean =
         (host == lowest.host) || (Some(host) == lowest.address.map(_.getHostAddress))
```
chbatey: could be a contains

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -139,17 +139,25 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
...
       }
 
       lowestAddressContactPoint(info) match {
-        case Some(lowest) => lowestContactPointIsSelfManagement(lowest)
-        case None => false
+        case Some(lowest) =>
+          if (matchesSelf(lowest)) true
+          else {
+            if (!info.contactPoints.exists(matchesSelf)) {
+              log.warning("Self contact point [{}] not found in targets [{}]", selfContactPoints.mkString(", "),
```
chbatey: We could be a bit more explicit that this is likely a misconfiguration, up to you

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
raboof: Yes, I agree but didn't find another way - (https://github.com/akka/akka-management/pull/238#discussion_r201627109)

PR #315, opened 2018-09-21
Consider 'Removed' an 'alive' state

### LINKED ISSUE #312: Consider 'Removed' as an 'alive' state for K8s liveness checking
Since the cluster state is `Removed` while in the process of discovering other nodes (https://github.com/akka/akka/issues/25663), for now I think it would make sense to present ourselves as 'alive' in that state as well, since we don't want Kubernetes to kill us while looking for peers.

raboof:
fixes #312


bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala
```diff
@@ -18,7 +18,16 @@ class KubernetesHealthChecks(system: ActorSystem) {
   //#health
   private val readyStates: Set[MemberStatus] = Set(MemberStatus.Up, MemberStatus.WeaklyUp)
   private val aliveStates: Set[MemberStatus] =
-    Set(MemberStatus.Joining, MemberStatus.WeaklyUp, MemberStatus.Up, MemberStatus.Leaving, MemberStatus.Exiting)
+    Set(
+      // 'Removed' might be counter-intuitive here, but the cluster status is also 'removed' during pre-join
+      // initialization (https://github.com/akka/akka/issues/25663)
+      MemberStatus.Removed,
```
TimMoore: This results in the counter-intuitive consequence that a "Down" node is considered dead, but after it has transitioned to "Removed" it is considered alive.

patriknw: I commented on the ticket 
https://github.com/akka/akka-management/issues/312#issuecomment-423885488

bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala
```diff
@@ -26,7 +26,10 @@ class KubernetesHealthChecks(system: ActorSystem) {
       MemberStatus.WeaklyUp,
       MemberStatus.Up,
       MemberStatus.Leaving,
-      MemberStatus.Exiting
+      MemberStatus.Exiting,
+      // We do not want to be killed when we are in Down state, because we want
+      // this to propagate to the leader so it can remove us.
+      MemberStatus.Down,
```
patriknw: Doesn't that cover all states?

chbatey:
Other failure was whiresource, gave it a kick

patriknw:
ok, so this is an intermediate fix until we have 
https://github.com/akka/akka/issues/25663 in place

I'm not convinced that the member status should be considered at all in the liveness check.
If all we want is to kill the process when the member has been removed from the cluster (unless it exited the jvm process itself) we can use `Cluster(system).isTerminated` instead of looking at members status.

chbatey:
yeah perhaps we should just make the liveness standard 200 or isTerminated :+1:

PR #319, opened 2018-09-25
[PoC] UDP broadcast discovery

jrudolph:
This is in an early stage but I wanted to register it here anyway.

I tested it with 3 nodes on vultr.com that share a private network.

I'll add some comments for explanation.

### DIFF
--- a/bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
+++ b/bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
@@ -6,10 +6,28 @@ akka {
 
   discovery {
     # pick the discovery method you'd like to use:
-    method = kubernetes-api
+    method = udp-broadcast
+    #method = config
+    config {
+      services {
+        my-cluster {
+          endpoints = [
+            {
+              host = "127.0.0.1"
+            },
+            {
+              host = "127.0.0.2"
+            }
+            {
+              host = "127.0.0.3"
+            }
+          ]
+        }
+      }
+    }
 
-    kubernetes-api {
-      pod-label-selector = "actorSystemName=appka"
+    udp-broadcast = {
+      class = "akka.discovery.udp.UdpBroadcastDiscovery"
     }
   }
 
@@ -19,10 +37,9 @@ akka {
     }
 
     cluster.bootstrap {
-
       contact-point-discovery {
-        service-name = "appka-service"
-        service-namespace = "default.svc.cluster.local"
+        service-name = "my-cluster"
+        service-namespace = ""
         stable-margin = 5 seconds
       }
 
--- a/bootstrap-joining-demo/kubernetes-api/src/main/scala/akka/discovery/udp/UdpBroadcastDiscovery.scala
+++ b/bootstrap-joining-demo/kubernetes-api/src/main/scala/akka/discovery/udp/UdpBroadcastDiscovery.scala
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.discovery.udp
+
+import java.net.InetSocketAddress
+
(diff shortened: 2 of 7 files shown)


bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
```diff
@@ -6,10 +6,28 @@ akka {
 
   discovery {
     # pick the discovery method you'd like to use:
-    method = kubernetes-api
```
jrudolph: I just repurposed the kubernetes-api demo for my purposes.

bootstrap-joining-demo/kubernetes-api/src/main/scala/akka/discovery/udp/UdpBroadcastDiscovery.scala
```diff
@@ -0,0 +1,97 @@
...
+        } else
+          log.info(s"Got unexpected probing command: [$cmd] [$value]") // TODO: show more of the message
+
+      /*case r @ Udp.Received(data, ret) =>
+        log.info(s"Received unhandled UDP: $r")*/
+
+      case SendProbe =>
+        //log.info(s"Broadcasting our address")
+        // TODO: we are cheating here, we shouldn't query for `thisServiceName`
+        listener ! Udp.Send(ByteString(s"QRY$thisServiceName"), new InetSocketAddress("10.0.0.0", wellKnownPort))
```
jrudolph: Subnet broadcast address should be configurable obviously. We could default to 0.0.0.0 (which depending on network configuration might interfere with something).

bootstrap-joining-demo/kubernetes-api/src/main/scala/akka/discovery/udp/UdpBroadcastDiscovery.scala
```diff
@@ -0,0 +1,97 @@
...
+import akka.discovery.SimpleServiceDiscovery.Resolved
+import akka.discovery.SimpleServiceDiscovery.ResolvedTarget
+import akka.io.Udp
+import akka.util.ByteString
+import akka.util.Timeout
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class UdpBroadcastDiscovery(system: ActorSystem) extends SimpleServiceDiscovery {
```
jrudolph: The whole lifecycle of this discovery mechanism is unclear right now. The reason is probably that it mixes up the server and client side of service discovery. 

For running an active discovery (client side) we need to broadcast discovery probes (one or several ones?) and wait for a certain time to wait for responses from peers. We don't know the number of peers so we cannot easily shortcut the lookup.

We could probably leave the server side running for nodes wanting to join later without having to know explicit seed nodes.

More ideas:
 * keep track of peers and also return those on queries
 * allow to use this service discovery mechanism for any kind of service discovery

Alternatives:
 * instead of the query/response protocol (joining node is actively searching for peers), during an initial interval broadcast beacons with the own address, other peers would listen to those during the initial phase (joining node is passively waiting for node beacons). This wouldn't fit the ServiceDiscovery pattern well and also wouldn't work as well for nodes that want to join later because you probably wouldn't want to publish beacons all the time

bootstrap-joining-demo/kubernetes-api/src/main/resources/application.conf
chbatey: Thief!

bootstrap-joining-demo/kubernetes-api/src/main/scala/akka/discovery/udp/UdpBroadcastDiscovery.scala
chbatey: why is this cheating? As it looks for thisServiceName even before a lookup has come?

patriknw:
Closing due to inactivity

Issue #326, opened 2018-10-03
Test kubernetes-api cluster bootstrap example on Travis


raboof:
Taking inspiration from https://github.com/akka/akka-management/tree/master/bootstrap-demo/kubernetes-api

Issue #287, opened 2018-09-05
Run whitesource in a separate stage


raboof:
This appears to reduce the risk of JPA locking problems on the server side.

Issue #256, opened 2018-08-02
Extend PR validation to cover Akka cluster formation


dwijnand:

ktoso:
Let me know however I can help out here!

chbatey:
We could add tests for `config` service discovery mechanism now. Were you also thinking in minikube?

dwijnand:
I guess you're asking @ktoso: I created this issue from his comment https://github.com/lightbend/orchestration-team/issues/77#issuecomment-409785094.

raboof:
I think both would be worthwhile, perhaps split the issue?

ktoso:
As the bootstrap dance is kind of important for us it would be really good to have it in validation. 
Splitting into separate things sounds okey

PR #329, opened 2018-10-03
test kubernetes-api on travis

raboof:
Refs #326, #256, also happens to implement #287


chbatey:
Nice, and only 4 minutes :)

PR #330, opened 2018-10-04
Correctly provide autodetermined binding port.

coreyoconnor:
When using auto selection for the HTTP binding port the resulting address will be `serverBinding.localAddress`. Derive the Uri from this.

### DIFF
--- a/management/src/main/scala/akka/management/AkkaManagement.scala
+++ b/management/src/main/scala/akka/management/AkkaManagement.scala
@@ -74,7 +74,7 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
   private[this] var _connectionContext: akka.http.scaladsl.ConnectionContext = Http().defaultServerHttpContext
 
   private val bindingFuture = new AtomicReference[Future[Http.ServerBinding]]()
-  private val selfUriPromise = Promise[Uri]() // TODO has to keep config as well as the Uri, so we can reject 2nd calls with diff uri
+  private val serverUriPromise = Promise[Uri]() // TODO has to keep config as well as the Uri, so we can reject 2nd calls with diff uri
 
   /**
    * Start the HTTP management endpoint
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @coreyoconnor,

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -116,16 +116,20 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
...
             )
 
-          serverBindingPromise.completeWith(serverFutureBinding).future.flatMap { _ =>
-            log.info("Bound Akka Management (HTTP) endpoint to: {}:{}", effectiveBindHostname, effectiveBindPort)
-            selfUriPromise.success(selfBaseUri).future
+          serverBindingPromise.completeWith(serverFutureBinding).future.flatMap { serverBinding =>
+            val serverAddress = serverBinding.localAddress
+            val path = settings.Http.BasePath.fold("")("/" + _)
+            val serverUri =
+              Uri(s"$protocol://${serverAddress.getHostName}:${serverAddress.getPort}${path}")
```
raboof: This appears to ignore the fact that the IP and port we bind to 'internal' to a container can be different from the 'externally visible' IP and port. It looks like we want to resolve the promise with the 'externally visible' IP and port.

I guess it doesn't make much sense to specify the port as '0' when there is a difference between the 'bound' and the 'externally visible' port, so perhaps a good strategy would be to:
* report the 'externally visible' IP, or if that is `0.0.0.0`, the actually-bound IP
* report the 'externally visible' port, or if that is `0`, the actually-bound port

Would that make sense in your situation as well?

coreyoconnor: Thanks for the feedback! Definitely was not sure about this patch :)

I'm unfamiliar with the externally visible port aspects. Reading through it. I think your proposal would work tho. If there is the requirement for the port to be externally visible there is unlikely to be the use of automatic port assignment. Doesn't make sense in that context.

For me, the use case is a local test cluster. For deployments a fixed port is used.

Will look through the code a bit more and post an update.

coreyoconnor:
Hi. I'm still committed to fixing this; just busy with other tasks right now :) Still on my queue tho

raboof:
> Hi. I'm still committed to fixing this; just busy with other tasks right now :) Still on my queue tho

That's fine! I'll close for now due to inactivity, but feel free to re-open when you get back to it!

PR #332, opened 2018-10-04
Simpler liveness check

raboof:


bootstrap-demo/kubernetes-dns/src/main/scala/akka/cluster/bootstrap/KubernetesHealthChecks.scala
```diff
@@ -35,10 +30,11 @@ class KubernetesHealthChecks(system: ActorSystem) {
...
           get {
-            val selfState = cluster.selfMember.status
-            log.debug("alive? clusterState {}", selfState)
-            if (aliveStates.contains(selfState)) complete(StatusCodes.OK)
-            else complete(StatusCodes.InternalServerError)
+            // When Akka HTTP can respond to requests, that is sufficient
+            // to consider ourselves 'live': we don't want K8s to kill us even
+            // when we're in the process of shutting down (only stop sending
+            // us traffic, which is done due to the readyness check then failing)
+            complete(StatusCodes.OK)
```
patriknw: good

I was initially thinking that we should check cluster.isTerminated, to catch cases when user doesn't exit the jvm when leaving/removing from cluster, but this endpoint will not be alive when the ActorSystem has been terminated so that should cover it (we terminated ActorSystem by default in coordinated-shutdown, but don't exit jvm by default)

raboof:
(build failure is #287)

PR #339, opened 2018-10-08
Extract sample code for docs

raboof:
Refs #4


bootstrap-demo/kubernetes-api-java/src/main/java/akka/cluster/bootstrap/demo/DemoApp.java
```diff
@@ -0,0 +1,47 @@
+/*
```
chbatey: Could we just put this in the same project to avoid the non code related duplication?

chbatey:
Given that the other PR switches this to use Maven will merge this, disregard my comment

PR #340, opened 2018-10-09
Allow connecting to the cluster bootstrap probe endpoint by IP

raboof:
I suspect this might be necessary when the DNS server is configured not to
issue DNS records for pods, which is possible with 'pods disabled'
https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#coredns-configmap-options

Not yet tested outside of the unit test

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -100,6 +100,12 @@ akka.management {
     # Configured how we communicate with the contact point once it is discovered
     contact-point {
 
+      # connect to the contact point by IP address instead of by hostname
+      #
+      # This is useful when coredns is configured with "pods disabled"
+      # Disabled by default because this is blocked when using Istio (#209)
+      connect-by-ip = no
+
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
       # Also, when no port-name is used and multiple results are returned for a given service with at least one
       # port defined, this port is used to disambiguate. When set to <fallback-port>, defaults to the value of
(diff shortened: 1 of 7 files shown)


raboof:
(tested locally with the kubernetes-api bootstrap demo, appears to work, but haven't checked actually disabling pod DNS)

chbatey:
If you only want access via the VIP or some service mesh this seems a valid setup.

chbatey:
Not sure how I closed this

chbatey:
Ah was mentioned in #412

chbatey:
@raboof have you tested this in a real cluster? Is it something we want? Seems useful

raboof:
I have tested it on a local cluster without `pods disabled` - it seemed to work, but the real test would be in a cluster with `pods disabled` of course. That might be tricky to set up, though.

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -95,6 +95,12 @@ akka.management {
     # Configured how we communicate with the contact point once it is discovered
     contact-point {
 
+      # connect to the contact point by IP address instead of by hostname
+      #
+      # This is useful when coredns is configured with "pods disabled"
+      # Disabled by default because this is blocked when using Istio (#209)
+      connect-by-ip = no
```
jrudolph: It might be my ignorance about k8s details, it's not clear to me from this description why it is needed.

So, with "pods disabled", the node cannot resolve names? Where does the IP address then come from?

On the other had, why doesn't it work with Istio? Does it add another layer of HTTP routing/reverse proxying that doesn't work without a proper `Host` header?

raboof: Indeed when coredns is configured with `pods disabled`, no DNS names will be generated for individual pods. Nodes must then always go through other 'more high-level' names to find other  services. I'm not sure if they will use more abstract names and still get the direct IP's, or that they also use more high-level IP's and the routing will be done in the infrastructure. Perhaps both are possible.

Indeed Istio (if I understand correctly) inserts itself by making all pod DNS names resolve to the 'sidecar' and doing the routing there based on the `Host` headers.

This is relatively unfamiliar territory for me as well, though.

raboof:
This PR has some conflicts and it would be good to test with a k8s (minikube?) that actually has `pods disabled` set.

We haven't really heard any demand for this yet though... let's close it until that time.

PR #342, opened 2018-10-11
Show and test how to use cluster bootstrap from maven

raboof:


bootstrap-demo/kubernetes-api-java/pom.xml
```diff
@@ -0,0 +1,145 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
chbatey: Aha maybe disregard my comment from the previous PR if we're doing to use maven for this

bootstrap-demo/kubernetes-api-java/build.sbt
```diff
@@ -0,0 +1,18 @@
+import com.typesafe.sbt.packager.docker._
```
chbatey: should we just remove this and have the scala one use sbt and the Java one use maven?

raboof:
The build failure is unrelated, #344. Please don't hit 'rebuild' though, I'd like to keep those logs.

Issue #344, opened 2018-10-12
Minikube SRV records yield inconsistent A names


raboof:
In the minikube test at https://travis-ci.org/akka/akka-management/jobs/440635804, the SRV request resolved to differently-named A records on different nodes:

```
[DEBUG] [10/12/2018 13:42:00.150] [default-akka.actor.default-dispatcher-15] [akka.tcp://default@172.17.0.7:2552/system/IO-DNS/async-dns/$a/$a] Decoded: Message(13,<AN,QUERY,AA,RD,RA,SUCCESS>,Vector(Question(_management._tcp.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,RecordType(33,SRV),RecordClass(1,IN))),Vector(SRVRecord(_management._tcp.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,10,33,8558,36393633363232.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local), SRVRecord(_management._tcp.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,10,33,8558,66646264346164.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local), SRVRecord(_management._tcp.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,10,33,8558,3337623737363734.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local)),Vector(),Vector(ARecord(36393633363232.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,/172.17.0.7), ARecord(66646264346164.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,/172.17.0.8), ARecord(3337623737363734.bootstrap-demo-kubernetes-dns-internal.default.svc.cluster.local,30,/172.17.0.9)))
```
vs
```
[...]

PR #347, opened 2018-10-15
Order addresses by address, then name

raboof:
Workaround for #344

This might not have the desired effect when there are multiple routers with
different IP addresses doing name-based routing, and the service discovery
returns those with unique names but also addresses that point to those routers.
Perhaps we should cross that bridge when we know it appears in practice though.


raboof:
(travis is green but apparently github missed the notification)

Issue #165, opened 2018-04-05
Better logging when discovery mechanism fails to load


lmlynik:
Continuing https://github.com/akka/akka-management/issues/163, there should be better logging and errors for that scenario

ktoso:
Łukasz, please when opening a new ticket like this make it self-explanatory so people don't have to do detective work what better logging and "that scenario" means. Could you copy the relevant information here? Help us help you :-)

chbatey:
E.g. `Illegal `akka.discovery.method` value 'kubernetes-api' or incompatible class! The implementation class MUST extend akka.discovery.SimpleServiceDiscovery and take an ExtendedActorSystem as constructor argument.` is not very helpful

teroxik:
Going to create a PR for this, as it misguided me as well, just a missing AWS_REGION variable for local testing in a docker container, but because of an update for couple of minor versions though that the plugin is wrongly configured.

PR #372, opened 2018-11-03
Discovery Service Additional Logging

teroxik:
- the underlying discovery mechanisms can throw exception related to their configuration, the exception is swallowed in recover with, causing confusion

Addresses the issue: #165


discovery/src/main/scala/akka/discovery/ServiceDiscovery.scala
```diff
@@ -58,11 +60,13 @@ final class ServiceDiscovery(implicit system: ExtendedActorSystem) extends Exten
       dynamic
         .createInstanceFor[SimpleServiceDiscovery](clazzName, (classOf[ExtendedActorSystem] → system) :: Nil)
         .recoverWith {
-          case _ ⇒
+          case e: Throwable ⇒
```
chbatey: I think we should probably only catch things related to the constructor not being available e.g. NoSuchMethod and then NonFatail(e) for the the other exceptions. Up to you if you want to do this now, happy to merge as is as this is definitely an improvement

chbatey: Sounds good :+1:

PR #375, opened 2018-11-05
Use Servicedisocvery.load in bootstrap

### LINKED ISSUE #368: Cluster bootstrap specific  service discovery doesn't work for all discovery types
When specifying a service discovery mechanism other than the globally `akka.discovery` mechanism bootstrap has its own way of loading a service discovery mechanism which doesn't work for all of the implementations. It should use ServiceDiscovery.lookup instead

chbatey:
Rather than duplicating the logic of loading a discovery mechanism

Fixes #368

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -48,11 +48,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
         discovery
 
       case otherDiscoveryMechanism ⇒
-        val implClazz = system.settings.config.getString(otherDiscoveryMechanism + ".class")
-        log.info("Bootstrap using [{}] discovery mechanism, instantiating [{}]", otherDiscoveryMechanism, implClazz)
-        system.dynamicAccess
-          .createInstanceFor[SimpleServiceDiscovery](implClazz, List(classOf[ActorSystem] → system))
-          .get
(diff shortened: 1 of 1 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -48,11 +48,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
         discovery
 
       case otherDiscoveryMechanism ⇒
-        val implClazz = system.settings.config.getString(otherDiscoveryMechanism + ".class")
-        log.info("Bootstrap using [{}] discovery mechanism, instantiating [{}]", otherDiscoveryMechanism, implClazz)
-        system.dynamicAccess
-          .createInstanceFor[SimpleServiceDiscovery](implClazz, List(classOf[ActorSystem] → system))
-          .get
+        log.info("Bootstrap using `akka.discovery` mechanism: {}", otherDiscoveryMechanism)
+        ServiceDiscovery(system).loadServiceDiscovery(otherDiscoveryMechanism)
```
raboof: This now also 'loads' the discovery mechanism - does this mean it becomes available when using akka-discovery 'directly' as well? Would that lead to confusion? Or was that already the case?

chbatey: It won't be used by the globally configured discovery but will already be cached if another area calls loadServiceDiscovery. So I don't think users will see any change.

PR #380, opened 2018-11-07
Note support status in docs

### LINKED ISSUE #379: Clarify support status
Based on akka/akka-grpc#454 when that is done

johanandren:
Fixes #379 

Started out with only the support status, ideas where to add recommended service discovery mechanisms welcome.

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -21,4 +21,4 @@ Support for some protocols is provided out of the box and others are provided as
 This project does not yet have to obey the rule of staying binary compatible between releases that is common 
 for Akka libraries. Breaking API changes may be introduced without notice as we refine and simplify based on your feedback.
 
-Akka gRPC is currently *Incubating*. The Lightbend subscription does not yet cover support for this project.
+Akka Management is currently *Incubating*. The Lightbend subscription does not yet cover support for this project.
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/index.md
```diff
@@ -14,3 +14,11 @@ Support for some protocols is provided out of the box and others are provided as
...
 
 @@@
+
+
+## Compatibility & support
+
+This project does not yet have to obey the rule of staying binary compatible between releases that is common 
+for Akka libraries. Breaking API changes may be introduced without notice as we refine and simplify based on your feedback.
+
+Akka gRPC is currently *Incubating*. The Lightbend subscription does not yet cover support for this project.
```
patriknw: gRPC

patriknw:
fyi @TimMoore

docs/src/main/paradox/index.md
johanandren: Mmmm that copy pasta

TimMoore:
Maybe https://developer.lightbend.com/docs/akka-management/current/discovery/index.html could be split into "core discovery methods" and "contributed discovery methods". Or we could list the recommended ones in the ["Discovery Method trade-offs"](https://developer.lightbend.com/docs/akka-management/current/discovery/index.html#discovery-method-trade-offs) section. That page is a bit long, to be honest, so maybe it would be a good idea to break it up into sub-pages with an index that designates which ones are recommended/Lightbend-maintained.

chbatey:
Yep, we can do that in another PR and get this one in

PR #382, opened 2018-11-07
Adds a parser to generate SRV Lookups from SRV strings

renatocaval:
This will be handy when creating a Lagom `ServiceLocator` backed by Akka `ServiceDiscovery`. 

Lagom's `ServiceLocator` doesn't support port name and protocol. An easy way to work this around is to configure Lagom to lookup for a full SRV string, eg: `_cassandra._tpc.my-cassandra-instance`.

Before calling the underlying `ServiceDiscovery`, Lagom's `ServiceLocator` can build a SRV Lookup using that string. 

This can obviously be used in other situations as well (not included in this PR). For instances, `SimpleServiceDiscovery.lookup(serviceName: String, resolveTimeout: FiniteDuration)` could use the parser automatically. 

Current impl: 

```scala 
def lookup(serviceName: String, resolveTimeout: FiniteDuration): Future[Resolved] =
    lookup(Lookup(serviceName), resolveTimeout)
```
 
Alternatively:

```scala 
def lookup(serviceName: String, resolveTimeout: FiniteDuration): Future[Resolved] =
    lookup(Lookup.parseSrvString(serviceName), resolveTimeout)
```

Though, that's debatable. SRV format is used by DNS and Kubernetes, but not necessarily makes sense for all ServiceDiscovery implementations.

### DIFF
--- a/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
+++ b/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
@@ -123,32 +123,20 @@ case object Lookup {
    * (as specified by https://www.ietf.org/rfc/rfc2782.txt)
    *
    * If the passed string conforms with this format, a SRV Lookup is returned.
+   * The serviceName part must be a valid domain name.
    *
    * The string is parsed and dismembered to build a Lookup as following:
    * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
    *
-   * The serviceName part must be a valid domain name.
-   *
    * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned using the whole string as service name.
-   *
-   * @throws IllegalArgumentException if the service name extracted from the SRV string is not a valid domain name.
    */
-  def fromString(str: String): Lookup = {
-
-    def fromDomainName(name: String): Lookup = {
-      if (validDomainName(name))
-        Lookup(name)
-      else
-        throw new IllegalArgumentException(s"Illegal domain name lookup: $name")
-    }
-
+  def fromString(str: String): Lookup =
     str match {
-      case SrvQuery(portName, protocol, serviceName) =>
-        fromDomainName(serviceName).withPortName(portName).withProtocol(protocol)
+      case SrvQuery(portName, protocol, serviceName) if validDomainName(serviceName) =>
+        Lookup(serviceName).withPortName(portName).withProtocol(protocol)
 
       case _ => Lookup(str)
     }
-  }
 
   /**
    * Returns true if passed string conforms with SRV format. Otherwise returns false.
--- a/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
+++ b/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
@@ -124,7 +124,7 @@ case object Lookup {
    * The string is parsed and dismembered to build a Lookup as following:
    * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
    *
-   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned.
+   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned using the whole string as service name.
    */
   def parseSrvString(srv: String): Lookup =
     srv match {
--- a/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
+++ b/discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
@@ -115,9 +115,12 @@ case object Lookup {
 
   private val SrvQuery = """^_(.+?)\._(.+?)\.(.+?)$""".r
 
+  private val DomainName = "^((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,6}$".r
+
   /**
    * Create a service Lookup from a string with format:
    * _portName._protocol.serviceName.
+   * (as specified by https://www.ietf.org/rfc/rfc2782.txt)
    *
    * If the passed string conforms with this format, a SRV Lookup is returned.
    *
@@ -125,19 +128,38 @@ case object Lookup {
    * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
    *
    * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned using the whole string as service name.
+   *
+   * @throws This method throws IllegalArgumentException if the service name is not a valid domain name.
    */
-  def fromString(str: String): Lookup =
-    srv match {
-      case SrvQuery(portName, protocol, serviceName) =>
-        Lookup(serviceName).withPortName(portName).withProtocol(protocol)
-      case _ => Lookup(srv)
+  def fromString(str: String): Lookup = {
+
+    def fromDomainName(name: String): Lookup = {
+      if (validDomainName(name))
+        Lookup(name)
+      else
+        throw new IllegalArgumentException(s"Illegal domain name lookup: $name")
     }
+
+    str match {
+      case SrvQuery(portName, protocol, serviceName) =>
+        fromDomainName(serviceName).withPortName(portName).withProtocol(protocol)
+
+      case _ => fromDomainName(str)
+    }
+  }
 
   /**
    * Returns true if passed string conforms with SRV format. Otherwise returns false.
    */
-  def isSrvString(srv: String): Boolean =
-    SrvQuery.pattern.asPredicate().test(srv)
+  def isValidSrv(srv: String): Boolean =
+    srv match {
+      case SrvQuery(_, _, serviceName) if validDomainName(serviceName) => true
+      case _ => false
+    }
+
+  private def validDomainName(name: String): Boolean =
+    DomainName.pattern.asPredicate().test(name)
+
 }
 
 /**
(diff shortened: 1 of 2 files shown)


dwijnand:
Is this a standard string format or an invention?

chbatey:
https://www.ietf.org/rfc/rfc2782.txt

renatocaval:
Maybe we should make the regex pattern more strict. I was only making sure that I could dismember the string in its three components, but we should also make sure that the last bit is well-formed.

Don't merge for now. Let me first have a second look on that.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -112,6 +112,32 @@ case object Lookup {
...
+   * _portName._protocol.serviceName.
+   *
+   * If the passed string conforms with this format, a SRV Lookup is returned.
+   *
+   * The string is parsed and dismembered to build a Lookup as following:
+   * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
+   *
+   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned.
+   */
+  def parseSrvString(srv: String): Lookup =
```
renatocaval: I was thinking about the method name. Because we can pass a non-compliant String and still get a valid `Lookup` back, we should probably not call it `parse`. 

Falling to parse should be an error and that's not what we are doing here. 

Some suggestions:

```suggestion
  def fromString(str: String): Lookup =
```
or 
```suggestion
  def fromSrvString(srv: String): Lookup =
```

I prefer the first one because we can effectively pass any string and get a valid `Lookup`.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -112,6 +112,32 @@ case object Lookup {
...
+  /**
+   * Create a service Lookup from a string with format:
+   * _portName._protocol.serviceName.
+   *
+   * If the passed string conforms with this format, a SRV Lookup is returned.
+   *
+   * The string is parsed and dismembered to build a Lookup as following:
+   * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
+   *
+   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned.
```
renatocaval:
```suggestion
   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned using the whole string as service name.
```

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -112,6 +112,32 @@ case object Lookup {
...
+      case SrvQuery(portName, protocol, serviceName) =>
+        Lookup(serviceName).withPortName(portName).withProtocol(protocol)
+      case _ => Lookup(srv)
+    }
+
+  /**
+   * Returns true if passed string conforms with SRV format. Otherwise returns false.
+   */
+  def isSrvString(srv: String): Boolean =
+    SrvQuery.pattern.asPredicate().test(srv)
```
renatocaval: FYI, this method will be useful to Lagom to check if it must add defaults port names and protocol.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -112,6 +112,54 @@ case object Lookup {
...
+   * (as specified by https://www.ietf.org/rfc/rfc2782.txt)
+   *
+   * If the passed string conforms with this format, a SRV Lookup is returned.
+   *
+   * The string is parsed and dismembered to build a Lookup as following:
+   * Lookup(serviceName).withPortName(portName).withProtocol(protocol)
+   *
+   * If the string doesn't not conform with the SRV format, a simple A/AAAA Lookup is returned using the whole string as service name.
+   *
+   * @throws This method throws IllegalArgumentException if the service name is not a valid domain name.
```
dwijnand:
```suggestion
   * @throws IllegalArgumentException if the service name is not a valid domain name.
```

You have to put the exception class first.

renatocaval: correct

renatocaval:
@chbatey / @dwijnand, please give it a second review. I did some modifications.

* changed method names to `fromString` and `isValidSrv`
* strict check that enforces domain name correctness. The service name part of an SRV string must be valid domain

dwijnand:
Now sometimes you get an exception for a misformed domain name, sometimes you don't, depending on which factory method you use.  That's not idea, I think.

dwijnand:
I think just drop the domain name checking, for now, and get the port and protocol parts in.

renatocaval:
> Now sometimes you get an exception for a misformed domain name, sometimes you don't, depending on which factory method you use. That's not idea, I think.

Yes indeed. I was in doubt if that's a good thing to do. Although it makes for SRV records, some service discovery implementation may consider any string as a valid identifier.

dwijnand:
Hmm, ok I buy that argument.

dwijnand:
But then drop the `fromDomainName` in the non-SRV part of `fromString`, no?

renatocaval:
It makes sense to not check it at all. If we don't enforce it when calling `Lookup("somedomain")` we should not enforce it when calling `fromString`.

For instance, this `_portName._protocol.service_name.local` will generate a `Lookup` with an invalid service name according to DNS, but maybe not for other impl. 

If we want it to be really SRV compliant, then `service_name.local` should not pass.

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
```diff
@@ -3,6 +3,7 @@
  */
 package akka.discovery
 
+import java.lang.invoke.MethodHandles
```
dwijnand:
```suggestion
```

discovery/src/test/scala/akka/discovery/LookupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+          lookup.serviceName shouldBe str
+          lookup.portName shouldBe empty
+          lookup.protocol shouldBe empty
+        }
+      }
+    }
+
+  }
+
+  "Lookup.isSrvString" should {
```
dwijnand:
```suggestion
  "Lookup.isValidSrv" should {
```

discovery/src/main/scala/akka/discovery/SimpleServiceDiscovery.scala
renatocaval: yep, just notice it. It gets imported when I try to code completion Lookup

discovery/src/test/scala/akka/discovery/LookupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+    "portName.protocol.serviceName.local",
+    "serviceName.local",
+    "_portName.serviceName",
+    "_serviceName.local",
+    "_serviceName,local",
+    "-serviceName.local",
+    "serviceName.local-",
+  )
+
+  "Lookup.parseSrvString" should {
```
dwijnand:
```suggestion
  "Lookup.fromString" should {
```

discovery/src/test/scala/akka/discovery/LookupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+
+class LookupSpec extends WordSpec with Matchers with OptionValues {
+
+  // SRV strings with invalid domain names
+  // should fail to build lookups
+  val srvWithInvalidDomainNames = List(
+    "_portName._protocol.service_name.local",
+    "_portName._protocol.servicename,local",
+    "_portName._protocol.servicename.local-",
+    "_portName._protocol.-servicename.local",
```
dwijnand:
```suggestion
    "_portName._protocol.-servicename.local"
```

discovery/src/test/scala/akka/discovery/LookupSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  // No SRV that should result in simple A/AAAA lookups
+  val noSrvLookups = List(
+    "portName.protocol.serviceName.local",
+    "serviceName.local",
+    "_portName.serviceName",
+    "_serviceName.local",
+    "_serviceName,local",
+    "-serviceName.local",
+    "serviceName.local-",
```
dwijnand:
```suggestion
    "serviceName.local-"
```

dwijnand:
Failure in `ClusterHttpManagementRoutesSpec`:

```
[info]   - when calling GET /cluster/shard_regions/{name} *** FAILED *** (4 seconds, 551 milliseconds)
[info]     java.util.concurrent.TimeoutException: Futures timed out after [1 second]
[info]     at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:223)
[info]     at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:227)
[info]     at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:190)
[info]     at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[info]     at scala.concurrent.Await$.result(package.scala:190)
[info]     at akka.cluster.http.management.ClusterHttpManagementRoutesSpec$$anonfun$1$$anonfun$apply$mcV$sp$16$$anonfun$apply$mcV$sp$23.apply(ClusterHttpManagementRoutesSpec.scala:355)
[info]     at akka.cluster.http.management.ClusterHttpManagementRoutesSpec$$anonfun$1$$anonfun$apply$mcV$sp$16$$anonfun$apply$mcV$sp$23.apply(ClusterHttpManagementRoutesSpec.scala:303)
```

rerunning

ignasi35:
All green. `1 second` timeout for `ClusterHttpManagementRoutesSpec` may be too aggressive in Travis.

PR #383, opened 2018-11-08
 Improve logging for kubernetes API communication issues

### LINKED ISSUE #371: Kubernetes service discovery: Log k8s API entity if a non-200 response code
Atm it logs the stack trace which is quite useless as it just says it can't unmarshall the body.

If there is a non-200 we shouldn't try and unmarshall the body and just log it at ERROR. 

Currently to get a useful log DEBUG is required.

chbatey:
* Don't unmarshal the response for non-200 status codes
* Specific log message for 403
* Log body at error for non-200 so users don't need to turn on DEBUG

Fixes #371

### DIFF
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
@@ -109,19 +109,28 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
             log.debug("Kubernetes API entity: [{}]", entity.data.utf8String)
             val unmarshalled = Unmarshal(entity).to[PodList]
             unmarshalled.failed.foreach { t =>
-              log.error(t, "Failed to unmarshal Kubernetes API response.  Status code: [{}]; Response body: [{}]",
-                response.status.value, entity)
+              log.warning(
+                  "Failed to unmarshal Kubernetes API response.  Status code: [{}]; Response body: [{}]. Ex: [{}]",
+                  response.status.value, entity, t.getMessage)
             }
             unmarshalled
           case StatusCodes.Forbidden =>
-            log.error("Forbidden to communicate with Kubernetes API server; check RBAC settings. Response: [{}]",
-              entity)
+            Unmarshal(entity).to[String].foreach { body =>
+              log.warning("Forbidden to communicate with Kubernetes API server; check RBAC settings. Response: [{}]",
+                body)
+            }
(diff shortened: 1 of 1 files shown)


discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
```diff
@@ -99,16 +103,28 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
...
 
-        unmarshalled.failed.foreach { t =>
-          log.error(t, "Failed to unmarshal Kubernetes API response status [{}]; check RBAC settings",
-            response.status.value)
+        response.status match {
+          case StatusCodes.OK =>
+            log.debug("Kubernetes API entity: [{}]", entity.data.utf8String)
+            val unmarshalled = Unmarshal(entity).to[PodList]
+            unmarshalled.failed.foreach { t =>
+              log.error(t, "Failed to unmarshal Kubernetes API response.  Status code: [{}]; Response body: [{}]",
```
raboof: :+1:

raboof: Maybe 'error' is a bit strong? At least 'warning' though indeed

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
```diff
@@ -99,16 +103,28 @@ class KubernetesApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleSer
...
+                response.status.value, entity)
```
raboof: does `entity.toString` contain the whole response or is it truncated?

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
chbatey: I'll switch to warning. The only case now is a version mis-match so nothing is likely to work.

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiSimpleServiceDiscovery.scala
chbatey: It'll truncate. Will switch to Unmarshal as a String.

Issue #370, opened 2018-11-02
Marathon Api port name


ignacio-dc:
The MarathonApiSimpleServiceDiscovery doesn't take the port name from the Lookup, it only takes it from the configuration. I fixed it with the same method that the kubernetes service uses
```scala
val portName = lookup.portName match {
      case Some(name) => name
      case None => settings.appPortName
    }
```
I could make the PR if you'd like

PR #386, opened 2018-11-09
Marathon Api port name (#370)

ignacioPL:
-Use the port name from the lookup if present, otherwise use it from the config

### DIFF
--- a/discovery-marathon-api/src/main/scala/akka/discovery/marathon/MarathonApiSimpleServiceDiscovery.scala
+++ b/discovery-marathon-api/src/main/scala/akka/discovery/marathon/MarathonApiSimpleServiceDiscovery.scala
@@ -88,7 +88,7 @@ class MarathonApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
       case Some(name) => name
       case None => settings.appPortName
     }
-    
+
     for {
       response <- http.singleRequest(request)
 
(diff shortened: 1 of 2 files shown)


discovery-marathon-api/src/main/scala/akka/discovery/marathon/MarathonApiSimpleServiceDiscovery.scala
```diff
@@ -84,6 +84,11 @@ class MarathonApiSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
 
     log.info("Requesting seed nodes by: {}", request.uri)
 
+    val portName = lookup.portName match {
```
raboof: Is this variable used anywhere?

ignacioPL: My bad, I commit the change with the use in targets

raboof: Aah, that makes a little more sense. Shouldn't it also be used while constructing the query somehow, though?

ignacio-dc: you can't filter by portname in the marathon Rest API

ignacioPL: you can't filter by portname in the marathon Rest API

chbatey:
Thanks @ignacioPL

PR #387, opened 2018-11-13
Custom dns resolver

chbatey:


discovery-dns/src/test/scala/akka/cluster/bootstrap/discovery/DnsDiscoverySpec.scala
```diff
@@ -3,140 +3,120 @@
...
 object DnsDiscoverySpec {
 
-  val config = ConfigFactory.parseString("""
+  val config = ConfigFactory.parseString(s"""
      //#configure-dns
      akka {
-       discovery.method = akka-dns
-       io.dns.resolver = async-dns
+       discovery {
+        method = akka-dns
```
raboof: Needs another space to indent nicely.

Nice that no resolver needs to be configured explicitly anymore.

PR #389, opened 2018-11-16
Expose routes so read only can be used in isolation

chbatey:
* Add javadsl for accessing the routes
* Correct how to use Cluster HTTP
* Document how to use the routes outside of Akka management

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,11 +29,10 @@ lazy val `akka-management-root` = project
     docs
   )
   .settings(
-    parallelExecution in GlobalScope := false,
-    inThisBuild(List(
-      resolvers += Resolver.sonatypeRepo("comtypesafe-2189")
-
-    ))
+    parallelExecution in GlobalScope := false
+//    inThisBuild(List(
+//      resolvers += Resolver.sonatypeRepo("comtypesafe-2189")
+//    ))
   )
 
 // interfaces and extension for Discovery
--- a/build.sbt
+++ b/build.sbt
@@ -30,9 +30,6 @@ lazy val `akka-management-root` = project
   )
   .settings(
     parallelExecution in GlobalScope := false
-//    inThisBuild(List(
-//      resolvers += Resolver.sonatypeRepo("comtypesafe-2189")
-//    ))
   )
 
 // interfaces and extension for Discovery
--- a/cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
+++ b/cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
@@ -18,7 +18,7 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
 
   import akka.http.scaladsl.server.Directives._
 
-  def routeGetMembers(cluster: Cluster): Route =
+  private def routeGetMembers(cluster: Cluster): Route =
     get {
       complete {
         val readView = ClusterReadViewAccess.internalReacView(cluster)
@@ -41,7 +41,7 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
       }
     }
 
-  def routePostMembers(cluster: Cluster): Route =
+  private def routePostMembers(cluster: Cluster): Route =
     post {
       formField('address) { addressString ⇒
         complete {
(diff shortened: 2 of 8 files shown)


build.sbt
```diff
@@ -29,11 +29,10 @@ lazy val `akka-management-root` = project
...
   .settings(
-    parallelExecution in GlobalScope := false,
-    inThisBuild(List(
-      resolvers += Resolver.sonatypeRepo("comtypesafe-2189")
-
-    ))
+    parallelExecution in GlobalScope := false
+//    inThisBuild(List(
+//      resolvers += Resolver.sonatypeRepo("comtypesafe-2189")
+//    ))
```
patriknw: remove?

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
```diff
@@ -143,4 +143,22 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
       }
     )
 
+  /**
+   * Creates an instance of [[ClusterHttpManagementRoutes]] with only the read only routes.
+   * the specified path `pathPrefixName`.
```
patriknw: where is that `pathPrefixName` specified?

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
```diff
@@ -143,4 +143,22 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
...
+   */
+  def readOnly(cluster: Cluster): Route =
+    concat(
+      pathPrefix("cluster" / "members") {
+        concat(
+          pathEndOrSingleSlash {
+            routeGetMembers(cluster)
+          },
```
patriknw: `routeGetMember` also?

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
helena: Why remove private scope on all? I mean versus at least something a little less vs completely open. Are we now Exposing to users?

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
chbatey: that is part of `routesMember`

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
chbatey: That is done in akka-mangement when it hosts them so not valid. Will remove comemnt.

chbatey: I removed the privates in case users wanted a subset of them. I can add them back.

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
```diff
@@ -143,4 +142,21 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
...
+   * Creates an instance of [[ClusterHttpManagementRoutes]] with only the read only routes.
+   */
+  def readOnly(cluster: Cluster): Route =
+    concat(
+      pathPrefix("cluster" / "members") {
+        concat(
+          pathEndOrSingleSlash {
+            routeGetMembers(cluster)
+          },
+          routesMember(cluster)
```
raboof: perhaps rename this method to `routesGetMember` to make it more obvious that it's (currently) meant to be readonly?

docs/src/main/paradox/cluster-http-management.md
```diff
@@ -146,3 +146,23 @@ Example response:
          }
        ]
      }
+
+## Hosting the routes in an existing Akka HTTP server
+
+Starting `AkkaMangement` starts an Akka HTTP server and hosts the Cluster HTTP Routes. The routes instead can be added
```
raboof: perhaps 'The routes can also be added..' rather than 'instead'?

docs/src/main/paradox/cluster-http-management.md
```diff
@@ -146,3 +146,23 @@ Example response:
...
+to an existing Akka HTTP server. To access all the routes:
+
+Scala
+:  @@snip [CompileOnlySpec.scala](/cluster-http/src/test/scala/doc/akka/cluster/http/management/CompileOnlySpec.scala) { #all }
+
+Java
+:  @@snip [CompileOnlyTest.java](/cluster-http/src/test/java/jdoc/akka/cluster/http/management/CompileOnlyTest.java) { #all }
+
+Or for just the read only when destructive operations such as downing members isn't required:
```
raboof: `just the read-only routes, when ...`?

cluster-http/src/main/scala/akka/management/cluster/ClusterHttpManagementRoutes.scala
chbatey: :+1:

docs/src/main/paradox/cluster-http-management.md
chbatey: :+1:

PR #390, opened 2018-11-16
Correct year in headers

chbatey:


raboof:
Travis fails checking the test scope file headers I think

Issue #351, opened 2018-10-17
Clarify usage of akka.management.cluster.bootstrap.form-new-cluster in documentation


TimMoore:
It might be misunderstood that `akka.management.cluster.bootstrap.form-new-cluster` will _always_ form a new cluster when `true`, when in fact it will always attempt to join an existing cluster. It's generally safe to set this to `true`, even when scaling out an initial deployment, but safer to set it to `false` once the initial cluster has formed.

Suggestions:

- Clarify in the documentation for this setting that cluster bootstrap always attempts to join an existing cluster, even when set to `true`
- Discuss scenarios where setting it to `false` could prevent accidental formation of multiple clusters when a cluster exists but cannot be discovered (do we know of any?)
- Recommend following the initial deployment with `form-new-cluster=true` with an immediate re-deployment setting `form-new-cluster=false`

helena:
@TimMoore one thought:

```"Recommend following the initial deployment with form-new-cluster=true with an immediate re-deployment setting form-new-cluster=false"```

I question a recommendation requiring a user to deploy then immediately re-deploy, this seems like a possible flaw in our strategy, but regardless, something we should handle for them, and not require a re-deploy.

TimMoore:
It's a fair point… and this isn't strictly required, but I can't think of a way that this could be handled automatically. How would a new node that is starting know whether it's part of an initial deployment or not?

chbatey:
At the moment it is the safest way to deploy a cluster. It is a pain but until we come up with a better solution we should document it.

helena:
@chbatey does this have a milestone to set? I see your recent closed one in this project is 0.20.0

chbatey:
Yep we set the issue, and if there isn't one we set it on the PR so we can find them for release notes. Also can you squash commits in future to keep the history clean?

Issue #356, opened 2018-10-24
Document exit on jvm and unsuccessful join seed nodes


chbatey:
If contact points are discovered and then shutdown then the joinSeedNodes will never succeed. The Bootstrap coordinator could continue probing in the background and re-call joinSeedNodes if the seeds have changed.

The Scala docs for joinSeedNodes says subsequent calls are ignored but this is only if a cluster has formed.

patriknw:
I suggest using
```
akka.cluster.shutdown-after-unsuccessful-join-seed-nodes = 30s
akka.coordinated-shutdown.exit-jvm = on
```

and then restart the process from scratch.

patriknw:
>  less convenient when in more traditional environments

Why is that? Wouldn't you have a restart loop (checking exit code) also if you launch with a sh script? The process may anyway die for other reasons, such as OOME, and then you want to start it again.

> and bootstrapping via DNS

why is that different?

chbatey:
Good point. We should just document this + that nodes should be wrapped in an auto restart script.

chbatey:
Looks like this made it into #391

PR #391, opened 2018-11-21
Clarify usage of akka.management.cluster.bootstrap.form-new-cluster #351

helena:
* Clarify usage of akka.management.cluster.bootstrap.form-new-cluster in code and documentation [351](https://github.com/akka/akka-management/issues/351)
* Added further testing 
* Minor cleanup of a test in cluster bootstrap

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -8,9 +8,15 @@ akka.management {
   http.route-providers += "akka.management.cluster.bootstrap.ClusterBootstrap$"
 
   cluster.bootstrap {
-    # When enabled (default), a new cluster will be formed via the bootstrapping mechanism. This can be disabled to only
-    # allow a node to join already existing clusters.
-    form-new-cluster = on
+
+    # Cluster Bootstrap will always attempt to join an existing cluster if possible. However
+    # if no contact points are found during service discovery, a new cluster will be formed.
+    # Setting `new-cluster-enabled=off` after an initial cluster has formed is recommended,
+    # accompanied by a re-deploy. Defaults to 'on'.
+    new-cluster-enabled = on
+
+    # The timeout applied to the bootstrap process.
+    timeout = 1 d
 
     # Configuration for the first phase of bootstraping, during which contact points are discovered
     # using the configured service discovery mechanism (e.g. DNS records).
@@ -48,6 +54,9 @@ akka.management {
       # and reuse what is configured for it. You can set it explicitly to something else here, if you want to
       # use a different discovery mechanism for the bootstrap than for the rest of the application.
       discovery-method = akka.discovery
+
+      # The timeout applied to binding to `self` bootstrap routes. Defaults to 10 seconds.
+      self-discovery-timeout = 10 s
 
       # Amount of time for which a discovery observation must remain "stable"
       # (i.e. not change list of discovered contact-points) before a join decision can be made.
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -52,9 +52,6 @@ akka.management {
       # use a different discovery mechanism for the bootstrap than for the rest of the application.
       discovery-method = akka.discovery
 
-      # The timeout applied to binding to `self` bootstrap routes. Defaults to 10 seconds.
-      self-discovery-timeout = 10 s
-
       # Amount of time for which a discovery observation must remain "stable"
       # (i.e. not change list of discovered contact-points) before a join decision can be made.
       # This is done to decrease the likelyhood of performing decisions on fluctuating observations.
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -15,9 +15,6 @@ akka.management {
     # accompanied by a re-deploy. Defaults to 'on'.
     new-cluster-enabled = on
 
-    # The timeout applied to the bootstrap process.
-    timeout = 1 d
-
     # Configuration for the first phase of bootstraping, during which contact points are discovered
     # using the configured service discovery mechanism (e.g. DNS records).
     contact-point-discovery {
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
@@ -13,43 +13,33 @@ import com.typesafe.config.Config
 import scala.concurrent.duration.{ FiniteDuration, _ }
 
 final class ClusterBootstrapSettings(config: Config) {
-  private implicit class HasDefined(val config: Config) {
-    def hasDefined(key: String): Boolean =
-      config.hasPath(key) &&
-      config.getString(key).trim.nonEmpty &&
-      config.getString(key) != s"<$key>"
-  }
+  import akka.management.AkkaManagementSettings._
 
   val managementBasePath: Option[String] =
     Option(config.getString("akka.management.http.base-path")).filter(_.trim.nonEmpty)
 
   private val bootConfig = config.getConfig("akka.management.cluster.bootstrap")
 
-  val formNewCluster: Boolean = bootConfig.getBoolean("form-new-cluster")
+  val newClusterEnabled: Boolean = bootConfig.getBoolean("new-cluster-enabled")
+
+  val bootTimeout: FiniteDuration = bootConfig.getDuration("timeout", TimeUnit.MILLISECONDS).millis
 
   object contactPointDiscovery {
     private val discoveryConfig: Config = bootConfig.getConfig("contact-point-discovery")
 
-    val serviceName: Option[String] =
-      if (discoveryConfig.hasDefined("service-name")) Some(discoveryConfig.getString("service-name"))
-      else None
+    val serviceName: Option[String] = discoveryConfig.optDefinedValue("service-name")
 
-    val serviceNamespace: Option[String] =
-      if (discoveryConfig.hasDefined("service-namespace")) Some(discoveryConfig.getString("service-namespace"))
-      else None
+    val serviceNamespace: Option[String] = discoveryConfig.optDefinedValue("service-namespace")
 
-    val portName = getOptionalString("port-name")
+    val portName: Option[String] = discoveryConfig.optValue("port-name")
 
-    val protocol = getOptionalString("protocol")
+    val protocol: Option[String] = discoveryConfig.optValue("protocol")
 
     def effectiveName(system: ActorSystem): String =
-      if (discoveryConfig.hasDefined("effective-name")) {
-        discoveryConfig.getString("effective-name")
-      } else {
-        val service = serviceName match {
-          case Some(name) ⇒ name
-          case _ ⇒ system.name.toLowerCase(Locale.ROOT).replaceAll(" ", "-").replace("_", "-")
-        }
+      discoveryConfig.optDefinedValue("effective-name").getOrElse {
+        val service =
+          serviceName.getOrElse(system.name.toLowerCase(Locale.ROOT).replaceAll(" ", "-").replace("_", "-"))
+
         val namespace = serviceNamespace match {
           case Some(ns) ⇒ s".$ns"
           case _ ⇒ ""
@@ -58,6 +48,9 @@ final class ClusterBootstrapSettings(config: Config) {
       }
 
     val discoveryMethod: String = discoveryConfig.getString("discovery-method")
+
+    val selfDiscoveryTimeout: FiniteDuration =
+      discoveryConfig.getDuration("self-discovery-timeout", TimeUnit.MILLISECONDS).millis
 
     val stableMargin: FiniteDuration =
       discoveryConfig.getDuration("stable-margin", TimeUnit.MILLISECONDS).millis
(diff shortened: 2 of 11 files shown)


cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -8,9 +8,15 @@ akka.management {
...
-    form-new-cluster = on
+
+    # Cluster Bootstrap will always attempt to join an existing cluster if possible. However
+    # if no contact points are found during service discovery, a new cluster will be formed.
+    # Setting `new-cluster-enabled=off` after an initial cluster has formed is recommended,
+    # accompanied by a re-deploy. Defaults to 'on'.
+    new-cluster-enabled = on
+
+    # The timeout applied to the bootstrap process.
+    timeout = 1 d
```
TimMoore: Does anyone know why this was set to a day in the first place? It seems very long.

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -8,9 +8,15 @@ akka.management {
...
 
   cluster.bootstrap {
-    # When enabled (default), a new cluster will be formed via the bootstrapping mechanism. This can be disabled to only
-    # allow a node to join already existing clusters.
-    form-new-cluster = on
+
+    # Cluster Bootstrap will always attempt to join an existing cluster if possible. However
+    # if no contact points are found during service discovery, a new cluster will be formed.
+    # Setting `new-cluster-enabled=off` after an initial cluster has formed is recommended,
+    # accompanied by a re-deploy. Defaults to 'on'.
```
TimMoore: Nice improvement.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -13,43 +13,33 @@ import com.typesafe.config.Config
...
-  }
+  import akka.management.AkkaManagementSettings._
 
   val managementBasePath: Option[String] =
     Option(config.getString("akka.management.http.base-path")).filter(_.trim.nonEmpty)
 
   private val bootConfig = config.getConfig("akka.management.cluster.bootstrap")
 
-  val formNewCluster: Boolean = bootConfig.getBoolean("form-new-cluster")
+  val newClusterEnabled: Boolean = bootConfig.getBoolean("new-cluster-enabled")
```
TimMoore: The new property name is much more clear, but I guess this will do the wrong thing on upgrade for anyone who is setting the old property and doesn't notice that it changed. Maybe we could detect the old property name, and if it's present, use that and warn?

TimMoore: As a special case of that, the old property name is used in the Lightbend Orchestration tooling: https://github.com/lightbend/reactive-cli/search?q=%22form-new-cluster%22

docs/src/main/paradox/bootstrap/details.md
```diff
@@ -65,10 +65,16 @@ The illustration below may be of help in visualising this process:
 
 ![project structure](../images/bootstrap-forming-cluster.png)
 
-Additionally, a setting is provided, `akka.management.cluster.bootstrap.form-new-cluster` that can be disabled to only allow the
-node to join existing clusters (Case 2 below). This can be used to provide additional safety during the typical deployment
-as it is relatively rare that no cluster exists. It can be specified in your `application.conf` or provided as an
-argument to the process itself via `-Dakka.management.cluster.bootstrap.form-new-cluster=<true|false>`. By default, it is enabled.
+Cluster Bootstrap will always attempt to join an existing cluster if possible. However if no contact points are found during service discovery, a
+new cluster will be formed. A setting is provided, `akka.management.cluster.bootstrap.new-cluster-enabled` that can be disabled to only allow the
```
TimMoore: Maybe "a new cluster will be formed by default." It's implied by the next sentence, but it might be a little bit more clear for some readers.

docs/src/main/paradox/bootstrap/details.md
```diff
@@ -65,10 +65,16 @@ The illustration below may be of help in visualising this process:
...
+node to join existing clusters (Case 2 below). This can be used to provide additional safety during the typical deployment, even when scaling out an initial deployment, as it is relatively rare that no cluster exists. 
+
+* On initial deployment use the default `akka.management.cluster.bootstrap.new-cluster-enabled=on`
+* Following the initial deployment it is recommended to set `akka.management.cluster.bootstrap.new-cluster-enabled=off` 
+  with an immediate re-deployment once the initial cluster has formed
+ 
+This can be specified in your `application.conf` or provided as an argument to the process itself via `-Dakka.management.cluster.bootstrap.new-cluster-enabled=<on|off>`. 
+By default, it is enabled.
```
TimMoore: This is all much more clear, thanks.

cluster-bootstrap/src/main/resources/reference.conf
helena: +1 I can not think of a reason we would want 1 day

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
helena: Hrm, I'll wait for @patriknw or @chbatey's input on preference for proceeding on this one. Good point.

docs/src/main/paradox/bootstrap/details.md
helena: 👍

cluster-bootstrap/src/main/resources/reference.conf
patriknw: the code has a comment
```
// the boot timeout not really meant to be exceeded
```

it's the timeout of a local `ask`

I guess it makes sense to have some (shorter) timeout and if bootstrap isn't successful it should run CoordinatedShutdown (stop the process, and let K8 start over again). Right now it's only logging a warning if the timeout is exceeded and nothing more, which isn't a good solution. Perhaps create a separate issue for this.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
patriknw: I agree that we should detect if the old property is set and then fail immediatly at startup, directing users to the new property

```
bootConfig.hasPath("form-new-cluster")
```

docs/src/main/paradox/bootstrap/details.md
```diff
@@ -65,10 +65,16 @@ The illustration below may be of help in visualising this process:
...
 
-Additionally, a setting is provided, `akka.management.cluster.bootstrap.form-new-cluster` that can be disabled to only allow the
-node to join existing clusters (Case 2 below). This can be used to provide additional safety during the typical deployment
-as it is relatively rare that no cluster exists. It can be specified in your `application.conf` or provided as an
-argument to the process itself via `-Dakka.management.cluster.bootstrap.form-new-cluster=<true|false>`. By default, it is enabled.
+Cluster Bootstrap will always attempt to join an existing cluster if possible. However if no contact points are found during service discovery, a
+new cluster will be formed by default. A setting is provided, `akka.management.cluster.bootstrap.new-cluster-enabled` that can be disabled to only allow the
+node to join existing clusters (Case 2 below). This can be used to provide additional safety during the typical deployment, even when scaling out an initial deployment, as it is relatively rare that no cluster exists. 
+
+* On initial deployment use the default `akka.management.cluster.bootstrap.new-cluster-enabled=on`
```
patriknw: One question is if we should even use `off` as default? The drawback is that it will be a rather awkward first experience for those that are not reading the documentation carefully.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/SelfAwareJoinDecider.scala
```diff
@@ -0,0 +1,69 @@
...
+  protected def contactPointsString(contactPoints: Set[(String, Int)]): String =
+    contactPoints.map(_.productIterator.mkString(":")).mkString("[", ",", "]")
+
+  /**
+   * The value `ClusterBootstrap(system).selfContactPoints` is set prior to HTTP binding,
+   * during [[akka.management.AkkaManagement.start()]], hence we accept blocking on
+   * this initialization.
+   */
+  private[bootstrap] def selfContactPoints: Set[(String, Int)] =
+    Try(Await.result(ClusterBootstrap(system).selfContactPoints, discoveryTimeout))
```
chbatey: I think it is only ever a bug in our code that means this won't be ready so i don't think we should expose a timeout to the user.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/SelfAwareJoinDecider.scala
```diff
@@ -0,0 +1,69 @@
...
+      .getOrElse(throw new IllegalStateException(
+            "'Bootstrap.selfContactPoint' was NOT set, but is required for the bootstrap to work " +
```
chbatey: I can't think of a usecase where these routes aren't hosted in AkkaManagement. 

I had to look at the code to see what this meant, and what the user would need to do if using the bootstrap routes outside of Akka Management is call `ClusterBootstrap(system).setSelfContactPoint()` but it is a `private[akka]` method so really that'll only be us.

cluster-bootstrap/src/main/resources/reference.conf
helena: @chbatey I like that idea because it solves the reason I added it - not having something like that hard coded and also keeps it simple. May create another ticket per @patriknw's thought.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/SelfAwareJoinDecider.scala
helena: This is relocated code which was duplicated (oddly) in the lowest address decider - so I removed the duplication and relocated the function to here.

docs/src/main/paradox/bootstrap/details.md
helena: @patriknw I went back and forth with that and tried both. I don't recall why but defaulting to `off` was tricky.

cluster-bootstrap/src/main/resources/reference.conf
chbatey: The shorter timeout achieves what can already be done with `akka.cluster.shutdown-after-unsuccessful-join-seed-nodes` 

The restart the process was discussed in https://github.com/akka/akka-management/issues/356 and was convinced by @patriknw that it wasn't worth it as in nearly all environments if we shutdown we'll be restarted.

I think for now we should remove the timeout, make it a `tell` and don't log a warning. Bootstrap logs a lot if it can't form a cluster so the user won't be short of logs describing why a cluster hasn't formed

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/SelfAwareJoinDecider.scala
chbatey: Ah no sorry let me explain better. The only reason this is an Await is because we don't have the selfContactPoints until Akka Management calls https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala#L106  via https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala#L67 which it does before bootstrap is allowed to start so this should never actually have to wait. If this times out then it means there's a bug in Akka Management rather than it taking a long time. 

So in that case I don't think exposing a timeout to a user is helpful because it should either be set on startup or never (i.e. if we have a bug).

helena: @chbatey ACK, updating :)

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -20,9 +21,9 @@ final class ClusterBootstrapSettings(config: Config) {
 
   private val bootConfig = config.getConfig("akka.management.cluster.bootstrap")
 
-  val newClusterEnabled: Boolean = bootConfig.getBoolean("new-cluster-enabled")
-
-  val bootTimeout: FiniteDuration = bootConfig.getDuration("timeout", TimeUnit.MILLISECONDS).millis
+  val newClusterEnabled: Boolean =
```
chbatey: :+1:

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -49,6 +52,9 @@ akka.management {
       # use a different discovery mechanism for the bootstrap than for the rest of the application.
       discovery-method = akka.discovery
 
+      # The timeout applied to binding to `self` bootstrap routes. Defaults to 10 seconds.
```
chbatey: Can be removed

docs/src/main/paradox/bootstrap/details.md
patriknw: I agree, lets keep the default to `on` and consider that as a separate change that needs more thinking

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
patriknw: after some additional clarifications from @TimMoore I understand that we can't fail fast if the old property was set since that would break reactive-cli.

We can change reactive-cli to use the new property but there is anyway a risk that the old reactive-cli version is used with this newer version of bootstrap, so at least for a migration period we need to be backwards compatible.

We need to read both properties. If `form-new-cluster=off` then we should use `off`, otherwise we can use the value of the new property `new-cluster-enabled`. Does that cover it?

chbatey:
@helena FYI we normally squash on merge unless there is a reason to keep distinct commits

helena:
@chbatey I knew that and somehow forgot. Won't happen again.

helena:
This change also resolves [#356](https://github.com/akka/akka-management/issues/356).

PR #393, opened 2018-11-21
Health check endpoints for Cluster HTTP

### LINKED ISSUE #364: Create a health check endpoint
Based on the example ones in the demo bootstrap projects

chbatey:
* Readiness: indicates user traffic is ready to be reseived
* Liveness: process is running

Updated kubernetes demo to use it.

Closes #364


cluster-http/src/main/scala/akka/management/cluster/ClusterHealthCheckRoutes.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.management.cluster.ClusterHealthCheckRoutes.ClusterHealthCheckSettings
+import akka.util.Helpers
+import com.typesafe.config.Config
+
+import scala.collection.JavaConverters._
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
```
chbatey: We can open these up to be used outside of akka management later maybe?

cluster-http/src/main/resources/reference.conf
```diff
@@ -7,12 +7,21 @@
 
 akka.management {
 
-  # registers bootstrap routes to be included in akka-management's http endpoint
+  # registers http management to be included in akka-management's http endpoint
   http.route-providers += "akka.management.cluster.ClusterHttpManagement"
 
+  # registers cluster healthchec to be included in akka-management's http endpoint
```
raboof: 'check'

cluster-http/src/main/resources/reference.conf
```diff
@@ -7,12 +7,21 @@
...
   cluster.http {
     # settings for the cluster routes
     # endpoint details are to be configured in akka-management
-  }
 
+    healthcheck {
+      # Ready health check returns 200 when cluster membership is in the following states.
+      # Intended to be used to indicate this node is ready for user traffic so Up/WeaklyUp
+      # Valid values: "Joining", "WeaklyUp", "Up", "Leaving", "Exiting", "Down", "Removed"
+      ready-states = ["Up", "WeaklyUp"]
```
raboof: Nice to make this configurable

cluster-http/src/test/scala/akka/cluster/http/management/ClusterHealthCheckRoutesSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+      withRoute(MemberStatus.Up) { route =>
+        Get("/ready") ~> route ~> check {
+          status shouldEqual StatusCodes.OK
+        }
+      }
+    }
+
+    "return 500 for unready state" in {
+
+      withRoute(MemberStatus.Down) { route =>
```
raboof: Checking for `WeaklyUp` here would also exercise the configuration

docs/src/main/paradox/bootstrap/recipes.md
```diff
@@ -54,11 +54,21 @@ Note that the `app` is the same for both services as they both refer to the same
 
 ### Health checks
 
-TODO make health check part of [Akka management](https://github.com/akka/akka-management/issues/364)
+`akka-cluster-http-management` includes an Akka Management route provider for a readiness and a liveness check. To use it
+add the following dependency:
```
raboof: I wonder if this shouldn't be automatic, since bootstrap will not work without the management routes anyway.

raboof: But that's for a different PR

chbatey: It'll work if you don't add the health checks to your deployment but ofc will result in traffic before the cluster is formed

chbatey: But perhaps a good idea anyway

cluster-http/src/main/resources/reference.conf
chbatey: I didn't fancy the Weakly Up argument :)

PR #394, opened 2018-11-21
Pull jackson-databind to the latest and more secure version 2.9.7 when it's pulled in transitively

jrudolph:
It seems there have already been more vulnerabilities found that will be fixed
in 2.9.8. See https://github.com/FasterXML/jackson-databind/blob/master/release-notes/VERSION-2.x#L25https://github.com/FasterXML/jackson-databind/blob/081ef5b62fe1c79aebadde2f75c99bc26fa3e140/release-notes/VERSION-2.x#L25


bootstrap-demo/aws-api-ec2/build.sbt
```diff
@@ -6,6 +6,8 @@ libraryDependencies += "com.amazonaws" % "aws-java-sdk-cloudformation" % "1.11.2
 
 libraryDependencies += "com.amazonaws" % "aws-java-sdk-autoscaling" % "1.11.271" % IntegrationTest
 
+libraryDependencies += "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.7" // aws SDK depends on insecure jackson
```
dwijnand: Reuse `JacksonDatabind`?

jrudolph: That build is off the main build.

jrudolph:
> I'd suggest defining `dependencyOverrides in ThisBuild` instead of `libraryDependencies` to refine the version of a transitive dependency:

Good point. Forgot about that.

jrudolph:
> I'd suggest defining `dependencyOverrides in ThisBuild` instead of `libraryDependencies` to refine the version of a transitive dependency:

Does that work for libraries, so that it ends up in the pom file?

jrudolph:
> Does that work for libraries, so that it ends up in the pom file?

It doesn't seem to end up in the pom-file with `dependencyOverrides`. The idea was to add an explicit entry to a safe version of jackson-databind to the dependencies so that by eviction we don't transitively depend on a vulnerable version any more. I wonder, if that would even help for the whitesource report at all.

But maybe this way of solving it is not a good idea in the first place? We could say that we cannot fix transitive dependencies but only direct dependencies.

jrudolph:
Latest AWS SDK and consul-client still depend on vulnerable jackson-databind version.

dwijnand:
No, good point it doesn't translate to POM files (Maven does "closest" rather than "latest").  So I guess this is the best you can do :-/

project/Dependencies.scala
```diff
@@ -110,16 +126,16 @@ object Dependencies {
...
   )
 
   val DiscoveryAwsApiAsync = Seq(
     libraryDependencies ++=
-      DependencyGroups.AkkaActor ++ Seq(
-        "software.amazon.awssdk" % "ecs" % "2.0.0-preview-9",
-        "org.scala-lang.modules" %% "scala-java8-compat" % "0.8.0")
+      DependencyGroups.AkkaActor ++
+      DependencyGroups.Aws2Ecs ++ Seq(
+        "org.scala-lang.modules" %% "scala-java8-compat" % "0.8.0"
```
patriknw: The compat dependency should be removed
It was fixed in master in https://github.com/akka/akka-management/pull/399

project/Dependencies.scala
```diff
@@ -53,6 +53,27 @@ object Dependencies {
       "com.typesafe.akka" %% "akka-http-testkit"                  % AkkaHttpVersion % "test"
     )
 
+    // often called-in transitively with insecure versions of databind / core
+    val JacksonDatabind = Seq(
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.7"
```
patriknw: I have confirmed that the pom.xml of a `publishLocal` contains:
```
        <dependency>
            <groupId>com.fasterxml.jackson.core</groupId>
            <artifactId>jackson-databind</artifactId>
            <version>2.9.7</version>
        </dependency>
```

project/Dependencies.scala
jrudolph: Done

project/Dependencies.scala
jrudolph: Thanks for checking.

Issue #377, opened 2018-11-06
Ec2TagBasedSimpleServiceDiscovery blocks on the default dispatcher


johanandren:
The lookup is blocking and is simply done in a `Future { ... }` block (https://github.com/akka/akka-management/blob/master/discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala#L118). This could cause starvation of the default threadpool.

sebastianharko:
@johanandren Assign to me plz, I'll fix it :-)

PR #403, opened 2018-12-01
EC2 tag based discovery: no longer block on default dispatcher

sebastianharko:
Hopefully, the proper fix for  #377

**Note**: Tested against real AWS environment via the integration testing script .

### DIFF
--- a/discovery-aws-api/src/main/resources/reference.conf
+++ b/discovery-aws-api/src/main/resources/reference.conf
@@ -2,15 +2,6 @@
 # Akka Service Discovery AWS Config                  #
 ######################################################
 
-ec2-tag-based-blocking-dispatcher {
-  type = Dispatcher
-  executor = "thread-pool-executor"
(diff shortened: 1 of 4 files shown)


discovery-aws-api/src/main/resources/reference.conf
```diff
@@ -2,6 +2,15 @@
...
 ######################################################
 
+ec2-tag-based-blocking-dispatcher {
+  type = Dispatcher
+  executor = "thread-pool-executor"
+  thread-pool-executor {
+    fixed-pool-size = 4
+  }
+  throughput = 1
+}
```
patriknw: Perhaps better to use the default shared akka.actor.default-blocking-io-dispatcher

sebastianharko: got it, will fix, actually totally forgot that there's a default shared blocking io dispatcher ;-)

PR #405, opened 2018-12-03
EC2 tag based discovery: custom EC2 client configuration

### LINKED ISSUE #401: EC2 tag based - EC2 client not configurable
Running Akka cluster from within a VPC I need to pass in the ProxyHost to the client so it can go out to the EC2 service and get the tags. Unless I've missed it I currently can't see a way of doing this. Probably the best would be to be able to pass in a client instance directly to Akka.

sebastianharko:
Fixes #401


docs/src/main/paradox/discovery/index.md
```diff
@@ -482,6 +482,46 @@ setting `akka.discovery.aws-api-ec2-tag-based.tag-key` to something else.
   but will _discover_ the "host" IP (the EC2 private IP) on the AWS API. See @ref:[Basic
   Configuration](../akka-management.md) on how to separate the bind IP from the discovery IP.
 
+* Configuration of the EC2 client is supported and can be done via `application.conf`, by setting
+`akka.discovery.aws-api-ec2-tag-based.client-config` to the fully qualified class name of a class that extends
+`com.amazonaws.ClientConfiguration`.
+
+    Scala
```
dhpiggott: Formatting seems to be broken (at least when GitHub renders it anyway):
screenshot 2018-12-08 at 14 02 19

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -68,6 +70,33 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
 
   private val runningInstancesFilter = new Filter("instance-state-name", List("running").asJava)
 
+  lazy val defaultClientConfiguration = {
```
dhpiggott: Why not this:
```suggestion
  private val defaultClientConfiguration = {
```
(or inline it)?

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -68,6 +70,33 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
...
+        system.dynamicAccess.createInstanceFor[ClientConfiguration](fqcn, Nil) match {
+          case Success(clientConfig: ClientConfiguration) ⇒
+            if (clientConfig.getRetryPolicy != PredefinedRetryPolicies.NO_RETRY_POLICY) {
+              log.warning(
+                  "Akka Cluster Bootstrap has its own retry/back-off mechanism, to avoid RequestLimitExceeded errors from AWS, " +
+                  "disable retries in the EC2 client configuration")
+            }
+            clientConfig
+          case Failure(ex) ⇒
+            throw new Exception(s"could not create instance of '$fqcn'", ex)
```
dhpiggott:
```suggestion
            throw new Exception(s"Could not create instance of '$fqcn'", ex)
```

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -68,6 +70,33 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
...
+    // we have our own retry/back-off mechanism (in Cluster Bootstrap), so we don't need EC2Client's in addition
+    clientConfiguration.setRetryPolicy(PredefinedRetryPolicies.NO_RETRY_POLICY)
+    clientConfiguration
+  }
+
+  private val ec2Client: AmazonEC2 = {
+    val clientConfiguration = clientConfigFqcn match {
+      case Some(fqcn) ⇒
+        system.dynamicAccess.createInstanceFor[ClientConfiguration](fqcn, Nil) match {
+          case Success(clientConfig: ClientConfiguration) ⇒
```
dhpiggott:
```suggestion
          case Success(clientConfig) ⇒
```

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -68,6 +70,33 @@ class Ec2TagBasedSimpleServiceDiscovery(system: ActorSystem) extends SimpleServi
...
+            if (clientConfig.getRetryPolicy != PredefinedRetryPolicies.NO_RETRY_POLICY) {
+              log.warning(
```
dhpiggott: Wonder if it would be better to fail-fast and throw an exception here instead.

sebastianharko: @dhpiggott I think that there are users that use this just for service discovery (and not for cluster bootstrap) hence they may want to have a configuration where a predefined retry policy is actually used. Hence a warning is a good solution for now, IMO. WDYT ?

docs/src/main/paradox/discovery/index.md
sebastianharko: @dhpiggott Fixed; Also, I changed it to use the snip directive :-)  Here's how it looks when paradox renders it:  
image

discovery-aws-api/src/test/java/akka/discovery/awsapi/ec2/MyConfiguration.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (C) 2017-2018 Lightbend Inc. <http://www.lightbend.com>
```
dhpiggott: Shouldn't this just be 2018?

discovery-aws-api/src/test/java/akka/discovery/awsapi/ec2/MyConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+    setProxyHost(".."); // and/or other things you would like to set
+
+    setRetryPolicy(PredefinedRetryPolicies.NO_RETRY_POLICY);
+    // If you're using this module for bootstrapping your Akka cluster,
+    // Cluster Bootstrap already has its own retry/back-off mechanism. To avoid RequestLimitExceeded errors from AWS,
+    // disable retries in the EC2 client configuration.
+  }
+
+}
+//#custom-client-config
```
dhpiggott: Missing EOF newline.

discovery-aws-api/src/test/scala/akka/discovery/awsapi/ec2/Docs.scala
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (C) 2017-2018 Lightbend Inc. <http://www.lightbend.com>
```
dhpiggott:
```suggestion
 * Copyright (C) 2018 Lightbend Inc. <http://www.lightbend.com>
```

sebastianharko: @dhpiggott the header is auto-generated by sbt test:compile,  not sure why it's 2017-2018 and not just 2018, let's ask @raboof :-)

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
sebastianharko: 100% agree ; I'm changing the warning msg. a bit to make it more clear .

sebastianharko:
@dhpiggott Thanks for the review!  :-)  Lots of valuable suggestions! 
Cheers!
Seb

discovery-aws-api/src/test/scala/akka/discovery/awsapi/ec2/Docs.scala
raboof: I think currently we have the `headerPlugin` configured so that all files have the same copyright header.

We want all files to have at least the current year (see also #390).

Indeed it would also be fine to add only the current year to new files, but keep 2017 intact for existing files. I think we have that configured in Akka.

todor-kolev:
If you wanted, for testing purposes, to stub out the EC2 client then how would you go about that? Would it not be better to allow users to provide an implementation of AmazonEC2 instead of ClientConfiguration?

raboof:
The CI failure can probably be fixed by rebasing on master (it was fixed by #412)

sebastianharko:
@todor-kolev quick question: are you using this in conjunction with akka cluster bootstrap or are you using this for service discovery only ? And I'm curious about your use case,  what makes it necessary to stub out the EC2 client :-)

sebastianharko:
> 
> 
> @sebastianharko In conjunction with akka cluster bootstrap.

Thanks for the response. So, for testing purposes or for demoing my app locally on my machine, I normally set (via  application.conf, in src/test) a different type of discovery (most often "Configuration", so scroll down to **Discovery Method:Configuration** in the [Service Discovery](https://developer.lightbend.com/docs/akka-management/current/discovery/index.html#discovery-method-configuration) page) . I hope this helps. Cheers!

discovery-aws-api/src/main/scala/akka/discovery/awsapi/ec2/Ec2TagBasedSimpleServiceDiscovery.scala
```diff
@@ -40,21 +41,22 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-    clientConfiguration.setRetryPolicy(PredefinedRetryPolicies.NO_RETRY_POLICY)
-    AmazonEC2ClientBuilder.standard().withClientConfiguration(clientConfiguration).build()
-  }
-
   private implicit val ec: ExecutionContext = system.dispatchers.lookup("akka.actor.default-blocking-io-dispatcher")
 
   private val config = system.settings.config.getConfig("akka.discovery.aws-api-ec2-tag-based")
 
+  private val clientConfigFqcn: Option[String] = { // FQCN of a class that extends com.amazonaws.ClientConfiguration
+    if (config.hasPath("client-config")) {
```
patriknw: I would prefer to have this in `reference.conf`:
```
akka.discovery.aws-api-ec2-tag-based.client-config = ""
```

with comment explaining what it is and that it can be defined to a fqcn of a class implementing `ClientConfiguration` with a default constructor (no parameters).

In that way `reference.conf` becomes the complete source/documentation of what configuration is possible.

By the way, should it support an `ActorSystem` constructor parameter? Perhaps it might need to read additional config via `system.settings.config` or use logging or something?

todor-kolev:
Do you have an example of how to use the `client-config` attribute? How do I for example set the `proxyHost`?

PR #413, opened 2018-12-11
Implemented Kubernetes namespace discovery

### LINKED ISSUE #345: kubernetes-api discovery method: determine pod namespace automatically
Version 0.18 can only work if the pods run in the default namespace. Revision 57312c4 improves this by looking at the NAMESPACE environment variable to determine the namespace.  This method still requires that the user sets the environment variable in Kubernetes. 
The fully automated way to discover the namespace is to look at the `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file, which is always mounted into a pod and contains the current namespace.

jroper:
Fixes #345

This also changes the api token lookup so it's read once using blocking IO at startup, rather than on every request.

One thing I wasn't sure about, I've removed support for automatically picking up the `NAMESPACE` environment variable. I'm not sure that this should ever be needed anymore, since the file it's reading it from should always be present. We could add support for the `NAMESPACE` environment variable back, it would just mean when it is defined, it will override what's found in the service account secret namespace. I guess if it ever disagrees, then not supporting it could break things, but to my understanding, if they did disagree, that would be a configuration bug and cluster bootstrap would likely fail (or worse, nodes would join the wrong cluster, eg test nodes might join the production cluster because the test nodes were configured with the production namespace).

### DIFF
--- a/bootstrap-demo/kubernetes-api-java/kubernetes/akka-cluster.yml
+++ b/bootstrap-demo/kubernetes-api-java/kubernetes/akka-cluster.yml
@@ -39,13 +39,6 @@ spec:
         - name: management
           containerPort: 8558
           protocol: TCP
-#namespace
-        env:
-        - name: NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
-#namespace
 
 ---
 kind: Role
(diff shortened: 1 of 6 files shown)


chbatey:
As this is done in the same as well then it is being tested in travis in minikube

chbatey:
Rebased and fixed build in #415

PR #414, opened 2018-12-11
API to allow embedding routes in an existing server

jroper:
This PR exposes the Akka Management router so that it can be served from the same HTTP server as an existing application, eg, so that it can be embedded in a Play/Lagom server. This means one less port for users to understand and configure.

### DIFF
--- a/management/src/main/scala/akka/management/AkkaManagement.scala
+++ b/management/src/main/scala/akka/management/AkkaManagement.scala
@@ -45,7 +45,7 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
   /**
    * Set async authenticator to be used for management routes.
    *
-   * Must be called BEFORE `start()` to take effect.
+   * Must be called BEFORE [[start()]] and [[routes()]] to take effect.
    * Do not call concurrently.
    */
   def setAsyncAuthenticator(auth: AsyncAuthenticator[String]): Unit =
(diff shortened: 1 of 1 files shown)


management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -77,28 +77,33 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
   private val bindingFuture = new AtomicReference[Future[Http.ServerBinding]]()
   private val selfUriPromise = Promise[Uri]() // TODO has to keep config as well as the Uri, so we can reject 2nd calls with diff uri
 
+  private val providerSettings = {
+    // port is on purpose never inferred from protocol, because this HTTP endpoint is not the "main" one for the app
+    val protocol = if (_connectionContext.isSecure) "https" else "http"
```
raboof: referring to `_connectionContext` from a `val` makes it impossible to customize this setting `setHttpsContext` - I guess we want to make this a `private def`?

chbatey: :+1:

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -77,28 +77,33 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
...
+    ManagementRouteProviderSettingsImpl(selfBaseUri)
+  }
+
   /**
-   * Start the HTTP management endpoint
+   * Get the routes for the HTTP management endpoint.
+   *
+   * This method can be used to embed the Akka management routes in an existing Akka HTTP server.
+   */
+  def routes: Try[Route] = prepareCombinedRoutes(providerSettings)
```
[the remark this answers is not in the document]
chbatey: Maybe not due to _connectionContext being overridable

chbatey:
Rebased + fixed build in #416

PR #415, opened 2018-12-11
Implemented Kubernetes namespace discovery

### LINKED ISSUE #345: kubernetes-api discovery method: determine pod namespace automatically

chbatey:
Fixes #345

This also changes the api token lookup so it's read once using blocking
IO at startup, rather than on every request.

Rebased version of https://github.com/akka/akka-management/pull/413

### DIFF
--- a/discovery-kubernetes-api/src/main/resources/reference.conf
+++ b/discovery-kubernetes-api/src/main/resources/reference.conf
@@ -15,18 +15,15 @@ akka.discovery {
     api-service-host-env-name = "KUBERNETES_SERVICE_HOST"
     api-service-port-env-name = "KUBERNETES_SERVICE_PORT"
 
-    # Namespace to query for pods
-    pod-namespace = "default"
+    # Namespace discovery path
+    #
+    # If this path doesn't exist, the namespace will default to "default".
+    pod-namespace-path = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
-    # Picks up namespace from $NAMESPACE env var if set. Add the following to the container spec to avoid having to
-    # hard code this
-    # env:
-    # - name: NAMESPACE
-    #   valueFrom:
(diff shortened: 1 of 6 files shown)


discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -15,18 +15,15 @@ akka.discovery {
...
-    pod-namespace = ${?NAMESPACE}
+    # Namespace discovery path
+    #
+    # If this path doesn't exist, the namespace will default to "default".
+    pod-namespace-path = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+    # Namespace to query for pods.
+    #
+    # Set this value to a specific string to override discovering the namespace using pod-namespace-path.
+    pod-namespace = "<pod-namespace>"
```
raboof: Can we assume this is always mounted? If not allowing this to be configured via the environment variable might be good to keep... (though we don't need to include it in our examples anymore, keeping them cleaner)

chbatey: Tiy can set this value using an env var to override getting it from a file

chbatey: I checked this was mounted on minikube and open shift

raboof: Ah, right, so you'd have to override it in your own `application.conf`. Yeah that sounds sufficient.

PR #421, opened 2018-12-18
Update whitesource 'master' project from master

raboof:


raboof:
Should be automatic -  but let's test :smile:

raboof:
Apparently this didn't have the intended effect - looking into it

Issue #425, opened 2018-12-20
Runtime check that all akka-http modules have the same version


patriknw:
In same way as in Akka with `akka.util.ManifestInfo`

PR #426, opened 2018-12-20
check that all Akka Management modules have the same version, #425

patriknw:
Refs #425


management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -32,6 +32,20 @@ object AkkaManagement extends ExtensionId[AkkaManagement] with ExtensionIdProvid
 }
 
 final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extension {
+
+  ManifestInfo(system).checkSameVersion(productName = "Akka Management",
+    dependencies = List(
+      "akka-discovery-consul",
```
raboof: Hmm now that the 'meat' of akka-discovery moved to akka itself, it seems a bit strange to keep the 'contrib' akka-discovery modules with akka-management. Perhaps we should more them to their own repo(s) at some point?

patriknw: Perhaps, but as long as they are released together I don't think we should complicate it and allow different versions.

PR #427, opened 2019-01-02
Adopt JDK 8 & 11, and Akka HTTP 10.1.6

patriknw:
Used https://github.com/akka/alpakka-kafka/blob/master/.travis.yml as template

### DIFF
--- a/.jvmopts-travis
+++ b/.jvmopts-travis
@@ -1,0 +1,7 @@
+# This is used to configure the sbt instance that Travis launches
+
+-Xms1G
+-Xmx1G
+-Xss2M
+-XX:ReservedCodeCacheSize=128m
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,19 @@ dist: xenial
 language: scala
 
 before_install:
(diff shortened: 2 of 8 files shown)


patriknw:
I think the remaining Travis build error is real, but probably fixed when we release new Akka HTTP compiled with Scala 2.12.8


With JDK 11:
```
[error] /home/travis/build/akka/akka-management/bootstrap-demo/kubernetes-api-java/src/main/java/akka/cluster/bootstrap/demo/DemoApp.java:37: reference to get is ambiguous
[error]   both method get(akka.actor.ActorSystem) in akka.http.javadsl.Http and method get(akka.actor.ActorSystem) in akka.http.javadsl.Http match
[error]     Http.get(system).bindAndHandle(complete("Hello world").flow(system, mat), ConnectHttp.toHost("0.0.0.0", 8080), mat);
```

patriknw:
hmm, Akka HTTP 10.1.6 was built with that and I tried to defined the dependencies of the demo app explicitly but it seems like maven is not using that?

patriknw:
ah, that is not built by maven

patriknw:
Updated to Akka HTTP 10.1.6, now it should work (fingers crossed).
I think that update is ok but we should confirm with Play/Lagom team.

PR #428, opened 2019-01-02
Towards stable APIs

patriknw:


project/Dependencies.scala
```diff
@@ -4,7 +4,7 @@ import Keys._
 object Dependencies {
 
   val AkkaVersion = "2.5.19"
-  val AkkaHttpVersion = "10.0.13" // TODO #281 update to 10.1.x when play and lagom are released
+  val AkkaHttpVersion = "10.0.13" // FIXME API #281 update to 10.1.x when play and lagom are released
```
patriknw: @renatocaval Do you know if it's ok if we bump this to 10.1.x now, when releasing Akka management 0.50 or later? New Play and Lagom releases should be ok and we will release new reactive-lib for 0.50.

renatocaval: Yes, that's ok for 0.50.

patriknw: This was done in 
https://github.com/akka/akka-management/pull/427
also needed for JDK 11

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -37,7 +37,7 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
 
   AkkaVersion.require("cluster-bootstrap", "2.5.19")
 
-  val settings = ClusterBootstrapSettings(system.settings.config, log)
+  val settings: ClusterBootstrapSettings = ClusterBootstrapSettings(system.settings.config, log)
```
chbatey: private?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -120,10 +120,15 @@ object ClusterBootstrap extends ExtensionId[ClusterBootstrap] with ExtensionIdPr
...
-  // TODO get the Initialized state once done
-  private[bootstrap] case object Initialized extends BootstrapStep
+  /**
+   * INTERNAL API
+   */
+  private[bootstrap] object Internal {
+    sealed trait BootstrapStep
+    case object NotRunning extends BootstrapStep
+    case object Initializing extends BootstrapStep
+    // TODO get the Initialized state once done
```
chbatey: Shall we just remove the Initialized? Not used and ClusterBootstrap sends a fire and forget start rather than an ask

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -5,20 +5,24 @@
...
 
 import akka.actor.ActorSystem
 import akka.event.LoggingAdapter
 import com.typesafe.config.Config
-
 import scala.concurrent.duration.{ FiniteDuration, _ }
+import scala.compat.java8.OptionConverters._
+import akka.util.JavaDurationConverters._
 
 final class ClusterBootstrapSettings(config: Config, log: LoggingAdapter) {
```
chbatey: Can we make this settings class internal?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/JoinDecider.scala
```diff
@@ -24,6 +24,8 @@ trait JoinDecider {
    */
   def decide(info: SeedNodesInformation): Future[JoinDecision]
 
+  // FIXME API support implementations in Java, returning CompletionStage?
```
chbatey: Low priority I think as not sure anyone will implement a JoinDecider

cluster-http/src/main/java/akka/management/cluster/ClusterReadViewAccess.java
```diff
@@ -18,7 +18,7 @@
    * Exposes the internal {@code readView} of the Akka Cluster, not reachable from Scala code because it is {@code private[cluster]}.
    */
   @InternalApi
-  public static ClusterReadView internalReacView(Cluster cluster) {
+  public static ClusterReadView internalReadView(Cluster cluster) {
```
chbatey: doh

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
```diff
@@ -4,17 +4,41 @@
 
 package akka.discovery.kubernetes
 
-import scala.collection.immutable.Seq
+import java.util.Optional
```
chbatey: All of this can be internal I think?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
patriknw: I think we typically have settings classes as public

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
patriknw: ok, looks like it's only used in JsonFormat. Is there a reason why we "always" make JsonFormat classes public?
if we make this internal we must make the JsonFormat internal

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
patriknw: ok

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -86,25 +102,38 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
...
+   * @throws IllegalArgumentException if routes configured for akka management
+   */
+  def getRoutes: akka.http.javadsl.server.Route = RouteAdapter(routes.get)
+
   /**
    * Start an Akka HTTP server to serve the HTTP management endpoint.
    */
-  // FIXME make it accept config object that would have all the `withHttps`
   def start(): Future[Uri] = {
+    // FIXME API make it accept config object that would have all the `withHttps`
```
patriknw: @chbatey What do you think about this one? I think it has major influence of the API? See `setHttpsContext` and `setHttpsContext` above?

Should it be moved to the `ManagementRouteProviderSettings` and should this method then take that as parameter. Same parameter is then need for `routes`, which is the alternative way of using this class.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/JoinDecider.scala
patriknw: ok, I agree

patriknw:
Added javadsl for ManagementRouteProvider, and therefore also moved the two interfaces into scaladsl and javadsl packages. 6d8516f

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -100,25 +119,38 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
...
+    // FIXME API return CompletionStage for Java API, must use different name
```
chbatey: While we're making a breaking change for 0.50 should we add java/scala dsl packages or is this over kill?

management/src/main/scala/akka/management/AkkaManagement.scala
chbatey: I think it is simpler to offer an overloaded API rather than two methods that need to be called in a certain order. But also don't think it is too important as it is something a user learns once and then never looks at it again

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
chbatey: It is useful for anyone wanting to consume the APIs programatically in the case of the Akka management formats but this one is for interacting with the k8s API so if it can be private it I think we should make it so

management/src/main/scala/akka/management/AkkaManagement.scala
```diff
@@ -100,25 +119,38 @@ final class AkkaManagement(implicit system: ExtendedActorSystem) extends Extensi
...
 
   /**
-   * Get the routes for the HTTP management endpoint.
+   * Scala API: Get the routes for the HTTP management endpoint.
    *
    * This method can be used to embed the Akka management routes in an existing Akka HTTP server.
    */
   def routes: Try[Route] = prepareCombinedRoutes(providerSettings)
 
+  // FIXME should `routes` return `Try` of throw IllegalArgumentException?
```
chbatey: This is likely used on start up and I'd say it is irrecoverable, I wouldn't want my app to start up if the management routes hadn't been setup correctly. But if they try to use the route they'll find out the Try is failed.

I don't have a strong opinion either way

patriknw:
I will follow up with the additional adjustments in a separate PR

management/src/main/scala/akka/management/AkkaManagement.scala
patriknw: done in https://github.com/akka/akka-management/pull/436

management/src/main/scala/akka/management/AkkaManagement.scala
patriknw: done in https://github.com/akka/akka-management/pull/436

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
patriknw: done in https://github.com/akka/akka-management/pull/436

PR #433, opened 2019-01-07
Removed Kubernetes self contact points hack

### LINKED ISSUE #424: K8s-api: <pod-namespace> is not always replaced correctly
In https://travis-ci.org/akka/akka-management/jobs/470381438 we saw:

```
12/20 07:29:01 INFO [Appka-akka.actor.default-dispatcher-4] a.m.c.b.LowestAddressJoinDecider - Exceeded stable margins without locating seed-nodes, however this node [172.17.0.8:8558,172-17-0-8.<pod-namespace>.pod.cluster.local:8558] is NOT the lowest address out of the discovered endpoints in this deployment, thus NOT joining self. Expecting node [ResolvedTarget(172-17-0-6.akka-bootstrap.pod.cluster.local,Some(8558),Some(/172.17.0.6))] (out of [ResolvedTarget(172-17-0-6.akka-bootstrap.pod.cluster.local,Some(8558),Some(/172.17.0.6)), ResolvedTarget(172-17-0-8.akka-bootstrap.pod.cluster.local,Some(8558),Some(/172.17.0.8)), ResolvedTarget(172-17-0-7.akka-bootstrap.pod.cluster.local,Some(8558),Some(/172.17.0.7))]) to perform the self-join and initiate the cluster.
```

This suggests the placeholder value `<pod-namespace>` is not always correctly replaced. Related to #415

jroper:
Fixes #424.

Issue #238 introduced a hack that meant that cluster bootstrap used the pod host name when the kubernetes configuration was present, rather than the IP address, since Kubernetes discovery returned host names rather than IP addresses to work with Istio. That hack was made redundant however by #242, which introduced matching by either hostname, or IP address, so it didn't matter if the self host name was an IP address or hostname. However, the hack wasn't removed and stayed there, benign, until #415 changed the way namespace configuration is consumed. At that point, the hack was still benign, but resulted in some confusing log messages. This removes the hack.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -99,16 +99,9 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     _selfContactPointUri.success(baseUri)
 
   /** INTERNAL API */
-  @InternalApi private[akka] def selfContactPoints: Future[Set[(String, Int)]] =
+  @InternalApi private[akka] def selfContactPoint: Future[(String, Int)] =
     _selfContactPointUri.future.map { uri =>
-      settings.joinDecider.selfDerivedHost match {
-        case Some(selfDerivedHost) if uri.authority.host.isIPv4 =>
-          val derivedHost = s"${uri.authority.host.toString.replace('.', '-')}.$selfDerivedHost"
-          log.info(s"Derived self contact point $derivedHost:${uri.authority.port}")
-          Set((uri.authority.host.toString, uri.authority.port), (derivedHost, uri.authority.port))
-        case _ =>
-          Set((uri.authority.host.toString, uri.authority.port))
-      }
+      (uri.authority.host.toString, uri.authority.port)
     }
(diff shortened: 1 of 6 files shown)


raboof:
I'm not completely convinced #242 makes this hack obsolete for the Istio (though I'd definitely be happy to see it go). I'll look into adding an Istio test to travis.

jroper:
That'd be great.

From what I understand, Istio communication should work because the Kubernetes API discovery returns the pod host name in the `host` field, which is what cluster bootstrap uses for HTTP and remoting connections, meanwhile Istio self host matching should work because the IP address returned by `getLocalHost.getAddress` is the public IP of the pod (even though it might not be accessible), and that will be matched against the IP address returned by the Kubernetes API. Previously, the matching worked because both the Kubernetes API discovery converted this IP address to the pod hostname, and the feature being deleted converted the discovered IP address via `getLocalHost.getAddress` to a pod hostname, and so they matched. So, matching the discovered address against the unconverted address will also work.

raboof:
I'm afraid I won't get to adding an Istio test in the near future, so this is OK to merge if you're confident the hack is no longer necessary.

patriknw:
Let's merge and we can look into it again if it's reported as a problem with Istio (so far it hasn't been top priority for us to look into Istio)

PR #436, opened 2019-01-09
Addition round of API changes before making it stable

patriknw:
Following up on https://github.com/akka/akka-management/pull/428

Pushed first thing, more will come...

* introduce scaladsl/javadsl package for AkkaManagement
  * because different return types of start, stop and getRoutes
  * removed the http package
  * also separate ManagementRouteProviderSettings because scalads/javadsl Uri


patriknw:
This is ready for review, and I did the changes for HttpsConnectionContext and AsyncAuthenticator in https://github.com/akka/akka-management/pull/437

PR #438, opened 2019-01-10
[WIP] Pluggable health check

chbatey:
Still needs some clean up and docs


patriknw:
> when thinking about where the docs go it made me consider whether this should be a different module? WDYT? 

Not sure if it adds much value to define it separately. Rather essential feature of Akka management, and the actual checks can be implemented elsewhere.

> The health check API could even be in akka as it isn't tied to HTTP

Let's leave it here to not "pollute" Akka core with too many things.

patriknw:
I'll review this in the afternoon

cluster-http/src/main/resources/reference.conf
```diff
@@ -11,7 +11,11 @@ akka.management {
   http.route-providers += "akka.management.cluster.ClusterHttpManagement"
 
   # registers cluster healthcheck to be included in akka-management's http endpoint
-  http.route-providers += "akka.management.cluster.ClusterHealthCheck"
+  http {
```
chbatey: this should probably move out of the http config object as health checks aren't specific to http

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+  def ready(): Future[Boolean] = {
+    check(readiness)
+  }
+
+  def alive(): Future[Boolean] = {
+    check(liveness)
+  }
+
+  private def check(checks: immutable.Seq[HealthCheck]): Future[Boolean] = {
+    Future.traverse(checks)(check => check()).map(_.forall(identity))
```
chbatey: Should we have a timeout? Configurable?

cluster-http/src/main/resources/reference.conf
patriknw: I agree

management/src/main/resources/reference.conf
```diff
@@ -44,9 +44,16 @@ akka.management {
...
-    route-providers = ${?akka.management.http.route-providers} []
+    route-providers =   ${?akka.management.http.route-providers} ["akka.management.http.HealthCheckRoutes"]
+
+    health-checks {
+
+      readiness-path = "ready"
+      liveness-path = "alive"
 
+      readiness-checks = ${?akka.management.http.health-checks.readiness} []
+      liveness-checks = ${?akka.management.http.health-checks.liveness}[]
```
patriknw: comparing this with the above `route-providers`, is it missing `-checks`? like this
```
      readiness-checks = ${?akka.management.http.health-checks.readiness-checks} []
      liveness-checks = ${?akka.management.http.health-checks.liveness-checks}[]
```

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+@InternalApi
+final private[akka] class HealthChecksImpl(system: ExtendedActorSystem, settings: HealthCheckSettings)
+    extends HealthChecks {
+  import HealthChecks._
+  import system.dispatcher
+
+  private val log = Logging(system, classOf[HealthChecksImpl])
+
+  log.info("Loading readiness checks {}", settings.readinessChecks)
+  log.info("Loading liveness checks {}", settings.livenessChecks)
```
patriknw: debug?

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+              fqcn,
+              immutable.Seq((classOf[ActorSystem], system))
+            )
+            .recoverWith {
+              case _: ClassCastException =>
+                system.dynamicAccess
+                  .createInstanceFor[java.util.function.Supplier[CompletionStage[Boolean]]](fqcn,
+                    immutable.Seq((classOf[ActorSystem], system)))
+                  .map(sup => () => sup.get().toScala)
+          }
```
patriknw: also try constructors without ActorSystem

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
patriknw: I think that would be good

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
```diff
@@ -0,0 +1,90 @@
...
+        case Failure(t) =>
+          throw new InvalidHealthCheckException(
+            "Uncaught exception from Health check construction",
+            t
+          )
+      }
+  }
+
+  def ready(): Future[Boolean] = {
+    check(readiness)
```
patriknw: maybe also protect against faulty implementations that throw exception instead of returning failed future

chbatey: :+1:

management/src/main/scala/akka/management/internal/HealthChecksImpl.scala
chbatey: :+1: will add that now

management/src/main/resources/reference.conf
chbatey: Looks like I have that but maybe I didn't in the version you saw

chbatey:
@patriknw I added docs, handling uncaught exceptions and added a timeout. Ready for final review.

cluster-http/src/main/scala/akka/management/cluster/ClusterHealthCheck.scala
```diff
@@ -1,17 +1,3 @@
 /*
  * Copyright (C) 2017-2018 Lightbend Inc. <http://www.lightbend.com>
  */
```
TimMoore: Looks like this somehow ended up as an otherwise empty file with a copyright header.

docs/src/main/paradox/healthchecks.md
```diff
@@ -0,0 +1,68 @@
...
+* Liveness checks: should the application be left running 
+
+Readiness checks can be used to decide if a load balancer should route traffic where
+as liveness checks can be used in environments which can restart a hung process.
+
+This matches [Kubernetes Health checks](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).
+
+## Defining a Health Check
+
+A health check is conceptually of the shape @scala[`ActorSystem => Future[Boolean]`]@java[`ActorSystem => CompletionStage<Boolean>`].
```
TimMoore: Will Java users understand the arrow function type syntax? Maybe better to write it out?

```suggestion
A health check is conceptually a function from an `ActorSystem` to a @scala[`Future[Boolean]`]@java[`CompletionStage<Boolean>`].
```

TimMoore: Actually, now that I read below that these actually take the `ActorSystem` in a constructor, I wonder if this sentence just confuses things. Maybe it's better to drop this sentence, and move "A health check must extend..." up here.

TimMoore:
I'll tag in @longmuir for a review too (particularly the docs)

cluster-http/src/main/scala/akka/management/cluster/ClusterHealthCheck.scala
chbatey: Doh!

docs/src/main/paradox/healthchecks.md
chbatey: Agreed!

patriknw: also, I think we moved the ActorSystem to a constructor parameter
(and Java syntax would be `->` if any arrow)

cluster-http/src/main/resources/reference.conf
```diff
@@ -24,4 +20,14 @@ akka.management {
       ready-states = ["Up", "WeaklyUp"]
     }
   }
```
patriknw: should we remove the `http` part for above `ready-states` config?

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterReadinessCheck.scala
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2017-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.management.cluster.scaladsl
```
patriknw: One important question you raised yesterday was about if this ClusterReadinessCheck should be in `akka-management-cluster-http` or in `akka-cluster` itself. As mentioned elsewhere we might add more specific checks for Cluster Sharding and similar.

I'm on the fence here.

On one hand it's nice to define the checks in the library modules themselves, so that `akka-management-cluster-http` doesn't have to dependencies to a lot of things. It already has dependencies to `akka-cluster` and `cluster-sharding` so that is not a problem for the checks we know about currently.

On the other hand I think these health checks are rather Akka Management (for K8 and similar env) and therefore don't belong in core Akka modules.

We had a similar discussion about if the contact point probing should be moved to akka-cluster and using Akka remoting instead of http. We decided for now that we continue with the http probing. Perhaps this question is similar? For now we define the cluster health checks in `akka-management-cluster-http`?

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterReadinessCheck.scala
```diff
@@ -0,0 +1,53 @@
...
+          s"'$invalid' is not a valid MemberStatus. See reference.conf for valid values"
+        )
+    }
+  def apply(config: Config): ClusterReadinessCheckSettings =
+    new ClusterReadinessCheckSettings(config.getStringList("ready-states").asScala.map(memberStatus).toSet)
+}
+
+final class ClusterReadinessCheckSettings(val readyStates: Set[MemberStatus])
+
+final class ClusterReadinessCheck @InternalApi private[akka] (system: ActorSystem,
```
patriknw: Perhaps rename this to the more specific `ClusterMembershipReadinessCheck`?

docs/src/main/paradox/healthchecks.md
```diff
@@ -0,0 +1,66 @@
...
+
+* Readiness checks: should the application receive external traffic, for example waiting for the following to complete
+    * Joining a Cluster
+    * Establishing a connection to a database or queuing system
+* Liveness checks: should the application be left running 
+
+Readiness checks can be used to decide if a load balancer should route traffic where
+as liveness checks can be used in environments which can restart a hung process.
+
+This matches [Kubernetes Health checks](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).
```
patriknw: Should we also link to Colin Breck's advice? https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/

docs/src/main/paradox/healthchecks.md
```diff
@@ -0,0 +1,66 @@
...
+Health checks can be hosted as via the Akka management HTTP server. To enable this add the `HealthCheckRoutes` as an Akka management route provider:
+
+@@snip [application.conf](/bootstrap-demo/local/src/main/resources/application.conf)  { #health-route }
+
+By default all readiness checks are hosted on `/ready` and liveness checks are hosted on `/alive`. If all of the checks
+for an endpoint succeed a `200` is returned, if any fail or return `false` a `500` is returned. The paths are configurable via `akka.management.health-checks.readiness-path` and `akka.management.health-checks.liveness-path` e.g.
+
+@@snip [application.conf](/bootstrap-demo/local/src/main/resources/application.conf)  { #health }
+
+
```
patriknw: docs looking good

management/src/main/resources/reference.conf
```diff
@@ -44,9 +44,15 @@ akka.management {
     # `route-providers += "com.example.MyExtraHealthCheckRoutes"` in the library `reference.conf`.
     #
     # Should not be set by end user applications in 'application.conf', use the extensions property for that
-    #
     route-providers = ${?akka.management.http.route-providers} []
+  }
 
+  health-checks {
```
patriknw: some docs here

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterReadinessCheck.scala
chbatey: I think the same checks would be used in a traditional environment for the LoadBalancer health checks so as long as we make it easy to use them without bootstrap I think they can start their life in akka management.

The checks are agnostic of http but perhaps they can stay in `cluster-http` until someone requests to use them without http? 

I foresee us adding ones for `akka-persistence-cassandra` to check the connection to Cassandra (perhaps requiring an Extension to get hold of the cluster like Couchbase has)

If we add checks for Singleton, Sharding then this probably means creating a `akka-management-cluster-singleton` etc rather than polluting the classpath when using just cluster (not sure how often cluster is used without singleton/sharding).

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterReadinessCheck.scala
chbatey: Yep, much better

chbatey: Went with `ClusterMembershipCheck` as it can be a readiness or liveness check in configuration

patriknw: ok, but it should never be used as a liveness check

cluster-http/src/main/resources/reference.conf
chbatey: Yep

docs/src/main/paradox/healthchecks.md
chbatey: Good idea

cluster-http/src/main/resources/reference.conf
```diff
@@ -24,4 +20,14 @@ akka.management {
...
+
+}
+
+# registers cluster healthcheck to be included in akka-management's http endpoint
+#health
+akka.management {
+  health-checks {
+    readiness-checks += "akka.management.cluster.scaladsl.ClusterReadinessCheck"
```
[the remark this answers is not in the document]
chbatey: As the user needs to call `AkkaManagement.start()` and add the health check route let's go with the automatic and the alternative configuration for health checks.

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterReadinessCheck.scala
chbatey: Leaving these in here for now

management/src/main/resources/reference.conf
```diff
@@ -55,10 +55,22 @@ akka.management {
...
-    liveness-checks = ${?akka.management.health-checks.liveness-checks} []
+    # Add readiness and liveness checks to the below config objects with the synax:
+    #
+    # name = "FQCN"
+    #
+    # For example:
+    #
+    # cluster-membership = "akka.management.cluster.scaladsl.ClusterMembershipCheck"
+    #
+    # Modules that contribute checks are expected to add their own checks to their reference.conf
```
patriknw: Libraries and frameworks that contribute checks are expected to add their own checks to their reference.conf. Applications can add their own checks to application.conf.

management/src/main/scala/akka/management/HealthCheckSettings.scala
```diff
@@ -11,11 +11,32 @@ import scala.collection.JavaConverters._
 import scala.collection.immutable
 import scala.concurrent.duration.FiniteDuration
 
+final case class NamedHealthCheck(name: String, fullyQualifiedClassName: String)
+
+// TODO filter out blank strings
```
patriknw: done

bootstrap-demo/local/src/test/scala/akka/management/LocalBootstrapTest.scala
```diff
@@ -0,0 +1,127 @@
...
+          aliveStatusCode(port) shouldEqual StatusCodes.OK
+        }
+      }
+    }
+    "form a cluster" in {
+      systems.foreach(ClusterBootstrap(_).start())
+      eventually {
+        clusters.foreach(
+          c =>
+            c.state.members.toList.map(_.status) shouldEqual List(
```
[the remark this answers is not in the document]
chbatey: Wanted to test all three were up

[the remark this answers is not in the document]
chbatey: We may get it where the first need just sees its self as up

Issue #268, opened 2018-08-22
Cross-build to Scala 2.13


dwijnand:
Depends on:
- [x] https://github.com/akka/akka/issues/25105

raboof:
Akka is released for 2.13.0-M3, though, so as a first step we could cross-build to that instead?

dwijnand:
Yep, sounds like a good first step.

dwijnand:
I was going to use this one for 2.13.. but oh well.

raboof:
(sorry, misunderstood then ;) )

PR #456, opened 2019-01-28
Scala 2.13.0-M5 support

raboof:
Refs #268

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpBootstrapJsonProtocol.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpBootstrapJsonProtocol.scala
@@ -30,6 +30,7 @@ object HttpBootstrapJsonProtocol extends DefaultJsonProtocol {
 
   // we use Address since we want to know which protocol is being used (tcp, artery, artery-tcp etc)
   final case class ClusterMember(node: Address, nodeUid: Long, status: String, roles: Set[String])
+  implicit val clusterMemberOrdering: Ordering[ClusterMember] = Ordering.by(_.node)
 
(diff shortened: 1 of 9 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/contactpoint/HttpBootstrapJsonProtocol.scala
```diff
@@ -30,6 +30,7 @@ object HttpBootstrapJsonProtocol extends DefaultJsonProtocol {
 
   // we use Address since we want to know which protocol is being used (tcp, artery, artery-tcp etc)
   final case class ClusterMember(node: Address, nodeUid: Long, status: String, roles: Set[String])
+  implicit val clusterMemberOrdering: Ordering[ClusterMember] = Ordering.by(_.node)
```
patriknw: how did it get the ordering before?

raboof: Before 2.12, `SortedSet.map[F, T](F => T)` would produce an unsorted `Set`.
Since 2.13, `SortedSet.map[F, T](F => T)` will produce a set that is kept sorted according the ordering on `T`.

So before 2.12 an `Ordering[ClusterMember`]` was not needed when we mapped a `SortedSet` from `Member` to `ClusterMember`. There are of course other ways to solve this at the `map` side, but those got kind of ugly or needlessly broke binary compat, so this seemed like a reasonable way to deal with it (even though we don't really care about the ordering here)

patriknw: good to know, thanks

Issue #457, opened 2019-01-29
HealthChecks via ActorSystem setup


patriknw:
To support Play/Lagom dependency injection into health checks.

PR #458, opened 2019-01-29
HealthChecks via ActorSystem setup, #457

patriknw:
* To support Play/Lagom dependency injection into health checks.

Refs #457

### DIFF
--- a/management/src/main/scala/akka/management/scaladsl/HealthChecks.scala
+++ b/management/src/main/scala/akka/management/scaladsl/HealthChecks.scala
@@ -3,15 +3,18 @@
  */
 
 package akka.management.scaladsl
+import scala.collection.immutable
+import scala.concurrent.Future
+
+import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
+import akka.actor.setup.Setup
 import akka.annotation.DoNotInherit
 import akka.management.HealthCheckSettings
 import akka.management.internal.HealthChecksImpl
 
-import scala.concurrent.Future
-
 /**
- * Loads health checks from configuration
+ * Loads health checks from configuration and ActorSystem Setup
  */
 object HealthChecks {
   def apply(system: ExtendedActorSystem, settings: HealthCheckSettings): HealthChecks =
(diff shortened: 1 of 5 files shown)


management/src/test/scala/akka/management/HealthChecksSpec.scala
```diff
@@ -206,5 +211,23 @@ class HealthChecksSpec
         HealthChecks(eas, settings(checks, checks))
       }.getCause shouldEqual ctxException
     }
+    "be possible to define via ActorSystem Setup" in {
+      val readinessSetup = ReadinessCheckSetup(system => List(new Ok(system), new False(system)))
+      val livenessSetup = LivenessCheckSetup(system => List(new False(system)))
+      // bootstrapSetup is needed for config (otherwise default config)
+      val bootstrapSetup = BootstrapSetup(ConfigFactory.parseString("some=thing"))
+      val actorSystemSetup = ActorSystemSetup(bootstrapSetup, readinessSetup, livenessSetup)
```
patriknw: @marcospereira @renatocaval This shows how health checks defined via ActorSystem Setup can be passed in when creating the ActorSystem. That might be a way to support dependency injection in the health checks?

Note that such Setup is used in Lagom for the PlayJson serializers:
https://github.com/lagom/lagom/blob/master/service/scaladsl/server/src/main/scala/com/lightbend/lagom/scaladsl/server/LagomApplicationLoader.scala#L261-L268

management/src/main/scala/akka/management/scaladsl/HealthChecks.scala
```diff
@@ -39,3 +42,39 @@ abstract class HealthChecks {
    */
   def alive(): Future[Boolean]
 }
+
+object ReadinessCheckSetup {
```
patriknw: TODO: add similar in javadsl

management/src/test/scala/akka/management/HealthChecksSpec.scala
renatocaval: This looks great!

I think this will also require some basic infrastructure in Play/Lagom. Which is ok, but probably won't make it in the 2.7.0/1.5.0 releases. 

We can add it later and we don't need to be in a hurry to have it ready right now in akka-management.

patriknw:
added javadsl

Issue #298, opened 2018-09-13
Typed extensions (ClassicActorSystemProvider)


chbatey:
Currently users have to:

```
AkkaManagement(system.toUntyped).start()
ClusterBootstrap(system.toUntyped).start()
```

404-:
I'd like to tackle this. I'm guessing, it would be best to reuse the existing, untyped implementations and do something similar to [AdaptedClusterImpl]. Right?
Also, should there be two separate tickets - one for `AkkaManagement` and another for `ClusterBootstrap`?

[AdaptedClusterImpl]: https://github.com/akka/akka/blob/5514afbe5400811cb60437c9e136878525893697/akka-cluster-typed/src/main/scala/akka/cluster/typed/internal/AdaptedClusterImpl.scala#L138-L139

johanandren:
When 2.5.25 has been released I think this could be achieved using the `akka.actor.ClassicActorSystemProvider` as an apply method parameter.

patriknw:
yes, that would be nice if we can make classic extensions work with typed.ActorSystem parameter. For java we probably have to add `get(ClassicActorSystemProvider)` overload.

jrudolph:
> yes, that would be nice if we can make classic extensions work with typed.ActorSystem parameter. For java we probably have to add `get(ClassicActorSystemProvider)` overload.

Is that something we still want to do (in 2.6)? I didn't find a ticket in akka/akka.

patriknw:
My experiment in https://github.com/akka/akka/pull/27760 didn't work out, and since no other brilliant idea have been proposed we have stayed with the explicit `toClassic` approach.

PR #460, opened 2019-02-05
[WIP] Add typed flavors of extensions #298

404-:
Very much WIP. For now, looking for feedback about the general approach, build settings, package structure, etc.

- [ ] Agreement on general approach
- [ ] Typed extension for `AkkaManagement` Scala DSL
- [ ] Typed extension for `AkkaManagement` Java DSL
- [ ] Typed extension for `ClusterBootstrap` Scala DSL
- [ ] Typed extension for `ClusterBootstrap` Java DSL

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,7 @@ lazy val `akka-management-root` = project
     `akka-discovery-kubernetes-api`,
     `akka-discovery-marathon-api`,
     `akka-management`,
+//TODO: `akka-management-typed`,
     `integration-test-aws-api-ec2-tag-based`,
     `integration-test-local`,
     `integration-test-aws-api-ecs`,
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -90,6 +91,16 @@ lazy val `akka-management` = project
     Dependencies.ManagementHttp
   )
 
+lazy val `akka-management-typed` = project
+  .in(file("management-typed"))
+  .dependsOn(`akka-management`)
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(unidocSettings)
+  .settings(
+    name := "akka-management-typed",
```
patriknw: this will result many artifact, would probably be needed for cluster-http, cluster-bootstrap as well.

Maybe we should take a slightly different approach here and include the typed stuff in the original modules but make the typed dependencies `provided`. wdyt?

404-: @patriknw just to clarify, you are talking about getting rid of the dedicated submodule `akka-management-typed` and adding the typed extensions to the existing `akka-management`, right? while making any additional dependencies `Provided`?

patriknw: Exactly, that was my thinking

raboof:
(closing due to inactivity for now - feel free to reopen when work resumes!)

PR #461, opened 2019-02-06
Upgrade akka and log in int-tests on pass

chbatey:
Mainly doing this to see the logs from DNS without setting it up locally


renatocaval:
Will this be included in v1.0.0?

Lagom is already using 2.5.20, but Akka Management is bringing in some deps on 2.5.19. I can overwrite it explicitly for the moment.

guizmaii:
Hey,

Anyone to merge and release this ?

Thanks :)

chbatey:
We're not using anything in 2.5.20 so shouldn't affect users that can override to 2.5.20 on their own

Issue #464, opened 2019-02-11
Inconsistencies in the doc for akka-management port name


guizmaii:
Hi Lightbenders,

How are you since time? Still bending light ? 🙂

While reading entirely the Akka-management doc, I saw some inconsistencies between some parts of the doc.

## 1. Discovery configuration

In `akka-management` doc, to setup the discovery mechanism, it's said to set:

> Set `akka.management.cluster.bootstrap.contact-point-discovery.discovery-method` to `akka.discovery.kubernetes-api`

Here: https://developer.lightbend.com/docs/akka-management/current/bootstrap/kubernetes-api.html

While, in the `akka-discovery-kubernetes-api`, it's said:

> And configure it to be used as default discovery implementation in your application.conf:
>
>```
>akka.discovery {
>  method = kubernetes-api
>}
>```

Here: https://developer.lightbend.com/docs/akka-management/current/discovery/kubernetes.html

Which one says the truth ?

## 2. Akka-management port name

In Akka-management and conf (`reference.conf`), the name by default and advised for the "management" port (`8558`) is: `"management"`.

While in the `akka-discovery-kubernetes-api` doc, the used name is `"akka-mgmt-http"`.
I didn't checked if this key is set or overridden in the  `akka-discovery-kubernetes-api` `reference.conf`.

Again, which one says the truth ?

Thanks 🙂

guizmaii:
Hum maybe there's no inconsistency, the two docs are maybe mutually exclusive ? 🤔

TimMoore:
I think the Kubernetes API method isn't really a good choice as the default service discovery method. It really only makes sense for cluster bootstrap. The docs should be updated.

TimMoore:
@guizmaii to answer your question 2, the port name doesn't really matter, except that what's set in the application configuration needs to match how the port is declared in the Kubernetes `Deployment` resource YAML. Other than that, it's an arbitrary lookup key, so it could be "management", "akka-mgmt-http", "i-like-pie", or anything else you want.

That said, we should stick to a single convention in the documentation, and "management" is preferred over "akka-mgmt-http".

TimMoore:
Also, we should clarify how these values connect using comments in the example configuration.

patriknw:
Renamed "akka-mgmt-http" to "management" in https://github.com/akka/akka-management/pull/477 but note the additional discussion in that PR. Not closing this ticket until we have an agreement.

TimMoore:
[...]

PR #468, opened 2019-02-15
Update kubernetes.md

glammers1:
I think that this documentation is not updated with the latest changes. Likewise, the configuration is hard coded in the md file instead of referencing code snippets from another file (integration tests or something like that). Is ok? or should it be improved in some way?

### DIFF
--- a/docs/src/main/paradox/discovery/kubernetes.md
+++ b/docs/src/main/paradox/discovery/kubernetes.md
@@ -31,10 +31,18 @@ values in your `application.conf`.
 ```
 akka.discovery {
   kubernetes-api {
-    pod-namespace = "default"
+    # Namespace discovery path
+    #
+    # If this path doesn't exist, the namespace will default to "default".
+    pod-namespace-path = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+  
+    # Namespace to query for pods.
+    #
+    # Set this value to a specific string to override discovering the namespace using pod-namespace-path.
+    pod-namespace = "<pod-namespace>"
 
-    # %s will be replaced with the configured effective name, which defaults to
-    # the actor system name
+    # Selector value to query pod API with.
+    # `%s` will be replaced with the configured effective name, which defaults to the actor system name
     pod-label-selector = "app=%s"
   }
 }
@@ -117,4 +125,3 @@ roleRef:
   name: pod-reader
   apiGroup: rbac.authorization.k8s.io
 ```
-


raboof:
I see this PR now conflicts with master - could you resolve those?

Indeed we often include snippets from (integration) tests in the documentation, perhaps that was hard to do in this case for some reason?

glammers1:
Conflicts came from #477 that resolve the question 2 in #464. But I have realized that the question 1 in #464 is related to this PR... Maybe this PR could be closed with a new one in the context of #464 answering the following questions:

* Question 1 of #464 
* Why the configuration is hard coded in the md file instead of referencing code snippets from another file?

OTOH, I think that the changes introduced in this PR related to the namespace behavior are correct.

TimMoore:
Note that I split question 1 from #464 into a new issue: #488

docs/src/main/paradox/discovery/kubernetes.md
```diff
@@ -19,25 +19,43 @@ Using `akka-discovery-kubernetes-api` is very simple, as you simply need to depe
 
 And configure it to be used as default discovery implementation in your `application.conf`:
```
TimMoore: This line needs to be updated as well, since it is no longer used as the default, only for cluster bootstrap.

docs/src/main/paradox/discovery/kubernetes.md
```diff
@@ -19,25 +19,43 @@ Using `akka-discovery-kubernetes-api` is very simple, as you simply need to depe
...
 
+* Set `akka.management.cluster.bootstrap.contact-point-discovery.discovery-method` to `kubernetes-api`
+* For bootstrap it is recommended to set the service discovery method via `akka.management.cluster.bootstrap.contact-point-discovery.discovery-method` rather then overriding the global service discovery mechanism with `akka.discovery.method`
```
TimMoore: Maybe this could be explained a little bit more:

```suggestion
* For bootstrap it is recommended to set the service discovery method via `akka.management.cluster.bootstrap.contact-point-discovery.discovery-method` rather then overriding the global service discovery mechanism with `akka.discovery.method`
* Most Kubernetes deployments should use [`akka-dns`](https://doc.akka.io/docs/akka/current/discovery/index.html#discovery-method-dns) as the default `akka.discovery.method`, which will use DNS SRV lookups for service-to-service discovery
```

docs/src/main/paradox/discovery/kubernetes.md
TimMoore:
```suggestion
And configure it to be used for cluster bootstrap in your `application.conf`:
```

raboof:
(the CI failure is https://github.com/akka/akka-management/pull/486)

TimMoore:
@glammers1 could you also merge from `master` again now that #486 is fixed? Thanks!

glammers1:
According to #488 I have done a few changes. I changed the first paragraph trying to avoid the use of Kubernetes API as service discovery method (maybe should be updated with more explanations about the real reasons, but I don't know the details in depth...).

docs/src/main/paradox/discovery/kubernetes.md
```diff
@@ -1,43 +1,48 @@
 ## Kubernetes API
 
-Another discovery implementation provided is one that uses the Kubernetes API. Instead of doing a DNS lookup,
-it sends a request to the Kubernetes service API to find the other pods. This method allows you to define health
-and readiness checks that don't affect the discovery method. Configuration options are provided to adjust
-the namespace, label selector, and port name that are used in the pod selection process.
+Kubernetes API method isn't really a good choice as the default service discovery method. It really only makes sense for cluster bootstrap. Most Kubernetes deployments should use [`akka-dns`](https://doc.akka.io/docs/akka/current/discovery/index.html#discovery-method-dns) as the default `akka.discovery.method`, which will use DNS SRV lookups for service-to-service discovery.
```
TimMoore: I feel like with this change, the page starts off sounding like it's discouraging people from using it at all. I think the original paragraph still applies. This caveat might be better placed as a [callout](https://developer.lightbend.com/docs/paradox/current/directives/callouts.html#note-callout) in the usage section, maybe by rewording the current bullet list down below on line 19.

glammers1: But in that case we have to explain how to use it as discovery method, so the following text still applies, isn't it?

```
akka.discovery {
  method = kubernetes-api
}
```

The original text could be updated just adding the "note callout" with the advice to use for cluster bootstrap. Am I misunderstanding something?

Related to the inline files:

* I don't know if the inline yaml examples are outdated. They are similar (but more generic) to those in [cluster bootstrap](https://github.com/akka/akka-management/blob/master/integration-test/kubernetes-api/kubernetes/akka-cluster.yml), maybe the same file could be used with the appropriate annotations above the code.

* The other inline text is just some parts of [reference.conf](https://github.com/akka/akka-management/blob/master/discovery-kubernetes-api/src/main/resources/reference.conf), there is not any example with that level of detail right now, because the examples don't override the default values.

TimMoore: The terminology can be a little confusing. "Discovery method" means an implementation of the Service Discovery API that can be used to lookup addresses given a particular service and port name. There are multiple use cases for using it, which have different needs:

* Cluster Bootstrap is for one node of a service finding its peers, and needs to resolve directly to the pod IPs
* Other cases typically are for one service to find another service, and often the name resolves to the address of a load balancer

So I'm suggesting that it's better to introduce the Kubernetes discovery method by saying what it _is_ good for (cluster bootstrap) rather than what it _isn't_ good for.

TimMoore: #501 rewrote this paragraph entirely, so I think if you take this change out and rebase again, the rest of the changes should be good to merge

TimMoore: Hmm actually some of the other details have also changed in the latest version so it will need a closer look.

glammers1: I think with #501 this PR is outdated, so I've merely changed the file with minor changes after rebase it with master.

docs/src/main/paradox/discovery/kubernetes.md
```diff
@@ -31,10 +31,18 @@ values in your `application.conf`.
...
-    # the actor system name
+    # Namespace discovery path
+    #
+    # If this path doesn't exist, the namespace will default to "default".
+    pod-namespace-path = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+  
+    # Namespace to query for pods.
+    #
+    # Set this value to a specific string to override discovering the namespace using pod-namespace-path.
+    pod-namespace = "<pod-namespace>"
```
raboof: :+1:

PR #471, opened 2019-02-19
[WIP] Include Cluster Health Checks by default

### LINKED ISSUE #470: Should HealthCheckRoutes be enabled by default?
Currently they aren't, though the documentation implies they are, and makes no mention of having to add them to the route providers:

https://developer.lightbend.com/docs/akka-management/current/bootstrap/recipes.html

Also, the documentation for cluster health checks references `akka-cluster-http-management` in the wording, but then in the code sample, instructs to add the `akka-management` dependency. Actually, both of these are wrong, it should be `akka-management-cluster-http`.

renatocaval:
Fixes #470

### DIFF
--- a/cluster-http/src/main/resources/reference.conf
+++ b/cluster-http/src/main/resources/reference.conf
@@ -11,16 +11,17 @@ akka.management {
   http.route-providers += "akka.management.cluster.ClusterHttpManagementRouteProvider"
 
   # includes HealthCheckRoutes by default, can be disabled by setting
-  # akka.management.cluster.health-check.default = null
-  http.route-providers += ${akka.management.cluster.health-check.default}
+  # akka.management.cluster.health-check.default.route-provider = null
+  http.route-providers += ${akka.management.cluster.health-check.default.route-provider}
 
   cluster {
+    health-check.default.route-provider = "akka.management.HealthCheckRoutes"
+
     health-check {
       # Ready health check returns 200 when cluster membership is in the following states.
       # Intended to be used to indicate this node is ready for user traffic so Up/WeaklyUp
       # Valid values: "Joining", "WeaklyUp", "Up", "Leaving", "Exiting", "Down", "Removed"
       ready-states = ["Up", "WeaklyUp"]
-      default = "akka.management.HealthCheckRoutes"
     }
 
   }
--- a/cluster-http/src/main/resources/reference.conf
+++ b/cluster-http/src/main/resources/reference.conf
@@ -11,8 +11,8 @@ akka.management {
   http.route-providers += "akka.management.cluster.ClusterHttpManagementRouteProvider"
 
   # includes HealthCheckRoutes by default, can be disabled by setting
-  # akka.management.cluster.default = null
-  http.route-providers += ${akka.management.cluster.default}
+  # akka.management.cluster.health-check.default = null
+  http.route-providers += ${akka.management.cluster.health-check.default}
 
   cluster {
     health-check {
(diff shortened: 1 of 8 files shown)


cluster-http/src/main/resources/reference.conf
```diff
@@ -10,13 +10,20 @@ akka.management {
   # registers http management to be included in akka-management's http endpoint
   http.route-providers += "akka.management.cluster.ClusterHttpManagementRouteProvider"
 
+  # includes HealthCheckRoutes by default, can be disabled by setting
+  # akka.management.cluster.health-check.default.route-provider = null
```
patriknw: I'm not sure that works, because variable substitution `$` isn't across reference and application conf

I'm also skeptical to that enabling `akka.management.HealthCheckRoutes` belongs to cluster-http.

If we really want to enable `HealthCheckRoutes` by default I think it's better to do that in akka-management, and have another very explicit flag there to disable `akka.management.HealthCheckRoutes`.

renatocaval: > I'm not sure that works, because variable substitution `$` isn't across reference and application conf 

I didn't know about that. There are a few things in Typesafe config that surprises me from time to time. That's probably one of them. 

> I'm also skeptical to that enabling `akka.management.HealthCheckRoutes` belongs to cluster-http.

I added there because it's also including the configuration for it. 
https://github.com/akka/akka-management/blob/master/cluster-http/src/main/resources/reference.conf#L13-L19

But I agree, that this should be where `HealthCheckRoutes`. Including its configuration.

management/src/main/resources/reference.conf
```diff
@@ -52,7 +52,13 @@ akka.management {
   }
 
   # Health checks for readiness and liveness
+  http.route-providers += "akka.management.HealthCheckRoutes"
```
renatocaval: This is including the HealthCheckRoutes by default

management/src/main/resources/reference.conf
```diff
@@ -52,7 +52,13 @@ akka.management {
...
+
   health-checks {
+
+    # whether health-check route-providers should be enabled
+    enabled = true
```
renatocaval: enabled by default, although I'm not so sure about this. We want it for sure when in combination with Cluster Bootstrap, but maybe not in other cases.

management/src/main/scala/akka/management/HealthCheckRoutes.scala
```diff
@@ -40,6 +40,22 @@ private[akka] class HealthCheckRoutes(system: ExtendedActorSystem) extends Manag
...
+            get {
+              onComplete(healthChecks.ready())(healthCheckResponse)
+            }
+          },
+          path(PathMatchers.separateOnSlashes(settings.livenessPath)) {
+            get {
+              onComplete(healthChecks.alive())(healthCheckResponse)
+            }
+          })
+    } else complete(StatusCodes.OK)
```
renatocaval: Not sure if the right thing to do. I was looking for a Route that could simply make this compile but with zero effect. 

I guess that if we add this here, all subsequential routes will be ignored. Can someone confirm it? What are the alternatives here?

One possible solution is to add the `akka.management.HealthCheckRoutes` programmatically if the flag is enabled (instead of using `route-providers`).

management/src/main/resources/reference.conf
patriknw: We shouldn't make it complicated. Either it is enabled or disabled by default, then the actual liveness and readiness checks are contributed by other modules such as `ClusterMembershipCheck` from akka-management-cluster-http.

I think it's fine to have it enabled by default like this, with the option to disable it.

management/src/main/scala/akka/management/HealthCheckRoutes.scala
patriknw: yes, this is probably not correct

@jrudolph is there a noop (non-matching) `Route`?

jrudolph: It's `reject` so the request backtracks out of that branch and tries the next ones.

patriknw: thanks

patriknw:
Superseded by https://github.com/akka/akka-management/pull/476

PR #472, opened 2019-02-19
Pick all ports before releasing for MultiDcSpec

chbatey:


patriknw:
I fixed the same in https://github.com/akka/akka-management/pull/476 but at many more places

PR #475, opened 2019-02-20
Documents how to disable or replace the default cluster health check

ignasi35:
As mentioned in https://github.com/akka/akka-management/issues/470#issuecomment-465004498

### DIFF
--- a/cluster-http/src/main/resources/reference.conf
+++ b/cluster-http/src/main/resources/reference.conf
@@ -26,6 +26,8 @@ akka.management {
 akka.management {
   health-checks {
     readiness-checks {
+      # Default HHealth chheck for cluster. Overwri the setting to replace it with
+      # your implementation or set it to "" (empty string) to disable this check.
       cluster-membership = "akka.management.cluster.scaladsl.ClusterMembershipCheck"
     }
   }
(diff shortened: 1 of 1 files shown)


ignasi35:
the build fails for different reasons:

 * fail to bind a port (flaky test?)
 * can't access a key file (some change on travis infra?)

cluster-http/src/main/resources/reference.conf
```diff
@@ -26,6 +26,8 @@ akka.management {
 akka.management {
   health-checks {
     readiness-checks {
+      # Default HHealth chheck for cluster. Overwri the setting to replace it with
```
patriknw: typo HH and Overwri

patriknw:
I fixed something that I think could have caused the port bind problems.

There is another build failure in the integration tests that I don't know what it is about:
```
Error in configuration: unable to read client-key /home/travis/.minikube/client.key for minikube due to open /home/travis/.minikube/client.key: permission denied
```

cluster-http/src/main/resources/reference.conf
ignasi35:
```suggestion
      # Default health check for cluster. Overwrite the setting to replace it with
```

My `h` key is acting up (see `chheck` too).

Issue #473, opened 2019-02-20
route-providers and application.conf ?


patriknw:
reference.conf has this comment
```
    # Should not be set by end user applications in 'application.conf', use the extensions property for that
    route-providers = ${?akka.management.http.route-providers} []
```

but I'm not sure we document it like that? For example plenty of the examples have 
```
route-providers += "akka.management.HealthCheckRoutes"
```

try out that it really works with custom route provider defined in application.conf combined with the built in ones like `HealthCheckRoutes` and `akka.management.cluster.bootstrap.ClusterBootstrap$`

PR #476, opened 2019-02-21
Change config structure for route providers, #470

patriknw:
* to make it possible to disable route providers
* enable HealthCheckRoutes by default

Refs https://github.com/akka/akka-management/issues/470

### DIFF
--- a/docs/src/main/paradox/akka-management.md
+++ b/docs/src/main/paradox/akka-management.md
@@ -150,15 +150,42 @@ are few pointers about how it all works together.
 
 The `akka-management` module provides the central HTTP endpoint to which extensions can register themselves.
 
-An extension can contribute to the exposed HTTP routes by appending to the `akka.management.http.route-providers` list in
-its own `reference.conf` (make sure to use `+=` instead of `=`). The core `AkkaManagement` extension
-collects all the routes and serves them together under the Management HTTP server. This enables
+An extension can contribute to the exposed HTTP routes by defining named route providers in the
+`akka.management.http.routes` configuration section in its own `reference.conf`. The core `AkkaManagement`
+extension collects all the routes and serves them together under the Management HTTP server. This enables
 easy extension of management capabilities (such as health-checks or cluster information etc)
 without the boilerplate and overhead to start separate HTTP servers for each extension.
 
 For example, the "Cluster HTTP Management" module exposes HTTP routes that can be used to monitor,
 and even trigger joining/leaving/downing decisions via HTTP calls to these routes. The routes and
 logic for these are implemented inside the `akka-management-cluster-http`.
+
+Management route providers should be regular extensions that aditionally extend the
+`akka.management.scaladsl.ManagementRoutesProvider` or `akka.management.javadsl.ManagementRoutesProvider`
+interface.
+
+Libraries may register routes into the management routes by defining entries to this setting
+the library `reference.conf`:
+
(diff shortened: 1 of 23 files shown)


patriknw:
Also solves https://github.com/akka/akka-management/issues/473

docs/src/main/paradox/akka-management.md
```diff
@@ -150,16 +150,43 @@ are few pointers about how it all works together.
...
+}
+```
+
+Where the `name` of the entry should be unique to allow different route providers to be registered
+by different libraries and applications.
+
+The FQCN is the fully qualified class name of the `ManagementRoutesProvider`.
+
+Route providers included by a library (from reference.conf) can be excluded by an application
+by using `""` as the FQCN of the named entry, for example:
```
renatocaval: What about `null`? Or maybe, a little bit more explicit, we could have a reserved word for it, for instance:

```
akka.management.http.routes {
  cluster-management = disabled
}
```

management/src/main/scala/akka/management/AkkaManagementSettings.scala
```diff
@@ -49,7 +49,17 @@ final class AkkaManagementSettings(val config: Config) {
...
+    val RouteProviders: immutable.Seq[NamedRouteProvider] = {
+      cc.getConfig("routes")
+        .root
+        .unwrapped
+        .asScala
+        .map {
+          case (name, value) => NamedRouteProvider(name, value.toString)
+        }
+        .toList
+        .filter(_.fullyQualifiedClassName != "")
```
renatocaval: Instead of `map` and `filter`, we could use a `collect`.

```
def valideFQCN(value: String) = {
  value != null || value.trim.nonEmpty || value != "disabled"
}

...

.collect {
  case (name, value) if valideFQCN(value.toString) => NamedRouteProvider(name, value.toString)
} 
```

docs/src/main/paradox/akka-management.md
patriknw: added support for `null` and `"null"` in 1f84992
I don't think we should introduce a new reserved word because it's difficult to support that consistently in all Akka (Lightbend) products. We already use `""` with the meaning of undefined in other places in Akka. `null` has a special meaning in Typesafe config so that's also ok.

docs/src/main/paradox/akka-management.md
```diff
@@ -150,16 +150,43 @@ are few pointers about how it all works together.
...
+`akka.management.http.routes` configuration section in its own `reference.conf`. The core `AkkaManagement`
+extension collects all the routes and serves them together under the Management HTTP server. This enables
 easy extension of management capabilities (such as health-checks or cluster information etc)
 without the boilerplate and overhead to start separate HTTP servers for each extension.
 
 For example, the "Cluster HTTP Management" module exposes HTTP routes that can be used to monitor,
 and even trigger joining/leaving/downing decisions via HTTP calls to these routes. The routes and
 logic for these are implemented inside the `akka-management-cluster-http`.
 
+Management route providers should be regular extensions that aditionally extend the
```
dwijnand:
```suggestion
Management route providers should be regular extensions that additionally extend the
```

patriknw:
Travis is too slow for my trail and error with fixing the minikube issue. Will follow up in separate PR. It's "only" the integration tests that are failing because minikube installation.

PR #477, opened 2019-02-21
use  as port name, #464

patriknw:
Trying to clarify based on comments for question 2 in issue #464. Could you confirm if this is what you intended @TimMoore ?

Refs #464

### DIFF
--- a/discovery-kubernetes-api/src/main/resources/reference.conf
+++ b/discovery-kubernetes-api/src/main/resources/reference.conf
@@ -34,6 +34,7 @@ akka.discovery {
 
     # Only used in the case that Lookup.portName is not set. Bootstrap sets this from
     # akka.management.cluster.bootstrap.contact-point-discovery.port-name
+    # This name must match the ports name in the deployment resource YAML.
     pod-port-name = "management"
   }
 }
--- a/docs/src/main/paradox/discovery/kubernetes.md
+++ b/docs/src/main/paradox/discovery/kubernetes.md
@@ -38,7 +38,8 @@ akka.discovery {
     # the actor system name
     pod-label-selector = "app=%s"
 
-    pod-port-name = "akka-mgmt-http"
+    # This name must match the ports name in the deployment resource YAML.
+    pod-port-name = "management"
   }
 }
 ```
@@ -72,7 +73,9 @@ spec:
           containerPort: 2552
           protocol: TCP
         # akka-management bootstrap
-        - name: akka-mgmt-http
+        # This name must match the name defined in
+        # akka.discovery.kubernetes-api.pod-port-name configuration
+        - name: management
           containerPort: 8558
           protocol: TCP
 ```


jroper:
Actually, I've been thinking of taking a different approach.

The application knows which port to connect to, because it's connecting to itself. There's no need to get Kubernetes involved in discovering the port here, it's overkill, it adds additional configuration overhead and more opportunity for users to get it wrong - following the guide I wrote the other day I missed the step of adding the port to my YAML spec and my cluster failed to start. It wasn't obvious from the logs what mistake I had made, though because I'm intimately familiar with the process I worked it out pretty quickly. If I could miss it, so will our users, and unlike me, they won't be able to work out what went wrong very quickly at all, they'll likely need to raise a support request.

My thought is that `pod-port-name` should default to `${?akka.management.http.port}`. And then, when read, if the value is a number, don't ask Kubernetes for the port number, just use that number directly. Otherwise, look for a port with that name in the pod. The default will work for at least 90% of users, probably 99%. The users it won't work for are users that don't get Akka management to serve itself, but instead mix its routes into their own HTTP servers routes. Those users can reconfigure it to by the port number of their server. The only time where it would be useful to use the port name instead of the port number is if from one version of the software to the next, you want to change the port number, but I expect that would be very rare.

patriknw:
> `pod-port-name` should default to `${?akka.management.http.port}`

Isn't this the classical problem with that Typesafe config can't do variable replacements across reference.conf and application.conf

if both pod-port-name is only in reference.conf as  `${?akka.management.http.port}` it will not be resolved to the value of `akka.management.http.port` defined in application.conf

jroper:
All depends on how you load your config. Typesafe config loading is in two phases, a parse phase, and a resolution phase. The resolution phase is what does the variable substitution. As long as you load all your config files and do all the fallbacks before resolution, it all just works. But after resolution, it doesn't. To demonstrate:

```
scala> import com.typesafe.config._
import com.typesafe.config._

scala> val c1 = ConfigFactory.parseString("a=1\nb=${a}")
c1: com.typesafe.config.Config = Config(SimpleConfigObject({"a":1,"b":${a}}))

scala> val c2 = ConfigFactory.parseString("a=2")
c2: com.typesafe.config.Config = Config(SimpleConfigObject({"a":2}))

scala> c2.withFallback(c1).resolve().getInt("b")
res1: Int = 2

scala> c2.withFallback(c1.resolve()).resolve().getInt("b")
res2: Int = 1
```

You can see that in the second time, I resolve `c1` before adding it as a fallback to `c2`, this means it does the property substitution of `a` when it still equals `1`, and so doesn't use the new `a` value when it's made a fallback of `c2`.

This would likely impact Akka a lot because I think Akka does a lot of manual post resolution fallbacks, I'm sketchy on the details because I haven't looked at it for a while. But, for a config object loaded by doing a `ConfigFactory.load()`, everything should just work.

jroper:
And this output demonstrates even better:

```
scala> c1.resolve()
res3: com.typesafe.config.Config = Config(SimpleConfigObject({"a":1,"b":1}))

scala> c2.withFallback(c1)
res4: com.typesafe.config.Config = Config(SimpleConfigObject({"a":2,"b":${a}}))

scala> c2.withFallback(c1).resolve()
res5: com.typesafe.config.Config = Config(SimpleConfigObject({"a":2,"b":2}))

scala> c2.withFallback(c1.resolve())
res6: com.typesafe.config.Config = Config(SimpleConfigObject({"a":2,"b":1}))
```

jroper:
Ah ok, I'm wrong, Typesafe configs config loading resolves `reference.conf` by default before using it. But we override this in Play, loading `reference.conf` without resolving it, as well as `play/reference-overrides.conf`, and the `application.conf`, we set them all up, and then resolve them all. So, if you get an `ActorSystem` from Play, it'll work because Play initialises the config so that `reference.conf` can reference things defined in `application.conf`, otherwise, it won't work.

patriknw:
ok, I'm merging this PR since it seems better than what is in master but I'll leave the issue open for further discussion and improvements.

raboof:
Defaulting to `${?akka.management.http.port}` will also not work when the container binds to `akka.management.http.port` on the 'inside' of the container but that port is mapped to another port on the 'outside' of the container (might be common when using host networking rather than bridge networking). Also it won't work when you have akka-management running on one port on some nodes and on another on others (e.g. during an upgrade in which the port changes): just that _you_ are listening on port X doesn't necessarily mean your peers are.

I guess both of those situations are rare, but they do make me doubt whether sidestepping discovery and getting the port name through a 'shortcut' like this really makes things so much simpler...

TimMoore:
@raboof it would still be overridable for rare cases like those. I agree with @jroper that we should optimize defaults for the common cases, while handling rare cases with optional configuration.

(BTW continued discussion has moved back to the original issue #464.)

Issue #463, opened 2019-02-08
akka-discovery-dns_2.12 version 0.20.0 is the last, but 1.0.0-RC2 is in the documentation page


fancywriter:
This page is wrong https://developer.lightbend.com/docs/akka-management/current/bootstrap/index.html

It says that **akka-management-cluster-bootstrap** and **akka-discovery-dns** have version "1.0.0-RC2", but for the second one only 0.20.0 is available.

https://search.maven.org/search?q=g:com.lightbend.akka.discovery%20AND%20a:akka-discovery-dns_2.12&core=gav
![screen shot 2019-02-08 at 12 31 30](https://user-images.githubusercontent.com/1200256/52478520-80aa1200-2b9d-11e9-84d6-edfcea019788.png)

guizmaii:
`akka-discovery-dns` should not be used anymore with 1.0.0-RC2. More info, see https://developer.lightbend.com/docs/akka-management/current/migration.html#akka-discovery

fancywriter:
So this doc page still need to be updated.

TimMoore:
Would it be possible for this doc page to include a snippet from one of the integration test projects, rather than hardcoding it in the page?

PR #478, opened 2019-02-21
update doc dependency of akka-discovery-dns, #463

patriknw:
Refs #463

### DIFF
--- a/docs/src/main/paradox/bootstrap/index.md
+++ b/docs/src/main/paradox/bootstrap/index.md
@@ -40,6 +40,19 @@ and bootstrap extensions:
   version2=$akka.version$
 }
 
+@@@ note
+
+`akka-discovery` is already a transitive dependency of `akka-management-cluster-bootstrap` but it can
+be good to define it explicitly in the build of the application to align the Akka versions with other
+dependencies from the application. The version must be the same across all Akka modules, e.g.
+`akka-actor`, `akka-discovery` and `akka-cluster` must be of the same version.
+
+The minimum supported Akka version is $akka.version$. Use the same Akka version for `akka-discovery`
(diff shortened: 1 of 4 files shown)


docs/src/main/paradox/bootstrap/index.md
```diff
@@ -28,15 +28,16 @@ A discovery mechanism needs to be chosen. A good default choice is DNS.
...
 @@dependency[sbt,Gradle,Maven] {
   group=com.lightbend.akka.management
   artifact=akka-management-cluster-bootstrap_$scala.binary_version$
   version=$version$
-  group2=com.lightbend.akka.discovery
-  artifact2=akka-discovery-dns_$scala.binary_version$
-  version2=$version$
+  group2=com.typesafe.akka
+  artifact2=akka-discovery_$scala.binary_version$
+  version2=$akka.version$
```
TimMoore: It isn't strictly necessary to include a direct dependency on `akka-discovery`, since `akka-management-cluster-bootstrap` already depends on it. If you think it's better to keep it for clarity sake or to ensure version alignment between core Akka modules, that's OK but I wanted to check.

patriknw: right, I didn't think of that it's already a transitive dependency. We might want to keep it explicit here for clarity and I think this is a high risk of being the wrong version if included via transitive, e.g. Akka Management will typically be built with an older version (2.5.20) and the application is using a newer (2.5.21). This will show the version that Akka management was built with but then it's at least easy to see what to override.

I'll add a note that the Akka version here should be the same as for other Akka dependencies included by the application, which also must be same or higher than what is shown here.

patriknw: added the note: ce24b5f

PR #480, opened 2019-02-22
debug minikube problem

patriknw:


raboof:
Refs #481

raboof:
Should be fixed with #486

Issue #481, opened 2019-02-22
minikube tests not working in Travis


patriknw:
Something with permissions. Maybe new installation of minikube.
```
Error in configuration: unable to read client-key /home/travis/.minikube/client.key for minikube due to open /home/travis/.minikube/client.key: permission denied
travis_time:end:0f523344:start=1550758468951922684,finish=1550758871172077623,duration=402220154939
```

raboof:
Should now work again

PR #482, opened 2019-02-22
Upgrade to Akka 2.5.21

renatocaval:


renatocaval:
Can this be added to upcoming v1.0.0?

patriknw:
Yes, we can start out with latest, but otherwise a library like this shoild be conservative to updating to latest, because that forces all dependencies to be at that at a minimum.

dwijnand:
Everyone should be using latest.

patriknw:
Yes, but by own will and not enforced by someone else :)

dwijnand:
"Enforced" implies there's a resistance.  What resistance is there to be upgraded to latest Akka?  (And by latest, I mean latest binary-compatible, patch version).

dwijnand:
Oh, wait!!  I keep making the same mistake and thinking that a bump for 2.5.20 to 2.5.21 is just bug fixes... that's not true for Akka, it could also be new buggy enhancements or features. Wow, I can't see anything but SemVer anymore.

renatocaval:
> "Enforced" implies there's a resistance. What resistance is there to be upgraded to latest Akka?

It doesn't mean that there is an inherent resistance to upgrade to latest Akka. I understand that rather as: a user may have a reason for not updating. We never know what are their reasons.

raboof:
Build failed due to #481

raboof:
Build failed due to #481

chbatey:
Needs a rebase

Issue #483, opened 2019-02-23
using consul discovery


winguse:
I got quite confused about how to use consul discovery.

## Using Consul Agent to register the management service

I was first to register the management service by consul agent client, something like this:
```scala
        clientAgent.register(
          /* port = */ 8558,
          /* tcp = */ HostAndPort.fromParts("127.0.0.1", 8558),
          /* interval = */ 10 seconds,
          /* name = */ "my-service-name",
          /* id = */ "my-service-id",
          /* tags = */ List("system:my-service", "akka-management-port:8558").asJava,
          /* meta = */ Map.empty.asJava
        )
```

But as debugging, I find the [`address`](https://github.com/akka/akka-management/blob/master/discovery-consul/src/main/scala/akka/discovery/consul/ConsulServiceDiscovery.scala#L72) `discovery-consul` get is Empty. So, it failed to get correct HTTP REST connection.

[And according to consul document](https://www.consul.io/api/catalog.html#serviceaddress):

> ServiceAddress is the IP address of the service host — if empty, node address should be used

So I think this might be a bug.

## Using Consul Catalog to register the management service

As I find it's using catalog agent to get catalog info, and I also check [the test code](https://github.com/akka/akka-management/blob/master/discovery-consul/src/test/scala/akka/cluster/bootstrap/discovery/ConsulDiscoverySpec.scala#L38). I try to switch to using catalog client:

```scala
            catalogClient
              .register(
                ImmutableCatalogRegistration
                  .builder()
                  .service(
                    ImmutableService
                      .builder()
                      .addTags("system:my-service", "akka-management-port:8558")
                      .address(address)
                      .id("my-service-id")
                      .service("my-service")
                      .port(8558)
                      .build()
                  )
                  .node(nodeName)
                  .address(address)
                  .build()
              )
```

but i find by using this way, the registered service will deregister after a while (I don't know why).

I read some document about consul, if I understand correctly, the recommended way is using agent instead of catalog. So, why we are using catalog here?

TimMoore:
@mlosiewicz-pl, since you contributed the Consul discovery implementation, are you able to comment on this? Thanks!

winguse:
update:

Regarding [consul doc](https://www.consul.io/docs/internals/anti-entropy.html#anti-entropy-1):

[...]

PR #485, opened 2019-02-26
improve the consul discovery

winguse:
Regarding #483 , here is our implementation of using consul discovery.

- using consul service discovery for discover management service instead of the tags on another service:
  - strong typed for port number
  - better consul query performance (consider there can be a lot services to be filter by tags)
- correct the behavior of getting `ServiceAddress` / `Address` from consul

This solution works for us, but may be not fit in every scenario.


lightbend-cla-validator:
Hi @winguse,

raboof:
The K8s errors on travis might be due to #486

DhashS:
I've addressed the collections incompatibility in https://github.com/Tubitv/akka-management/pull/1

DhashS:
signed the cla

raboof:
Manually verified you indeed signed the CLA. It was not obvious that the Travis failure was related to this change, so triggered it again.

radix-aw:
Is there anything left to be done on this PR? This could be very useful.

discovery-consul/src/main/resources/reference.conf
```diff
@@ -13,13 +13,9 @@ akka.discovery {
...
-    # services with this tag present will be found by the discovery mechanism
-    # i.e. `system:test` will be found in cluster if the cluster system is named `test`
-    application-name-tag-prefix = "system:"
-
-    # Prefix for tag containing port number where akka management is set up so that
-    # the seed nodes can be found, an example value for the tag would be `akka-management-port:19999`
-    application-akka-management-port-tag-prefix = "akka-management-port:"
+    # the suffix to discover management service
+    # your code should have logic to register SERVICE_NAME-akka-mangement to consul
+    management-service-suffix = "-akka-management"
```
chbatey: The old implementation had the same issue that it assumes that it is discovering the akka managementport, where as it could be service discovery for anything e.g. a gRPC service

winguse: i am not following. did you suggest we to use another suffix?

DhashS:
We have an internal tool that bootstraps an akka cluster using consul without going through the normal akka management flow, but instead using consul’s KV store (it supports atomic compare-and-set) to bootstrap our cluster. There is little configuration required, only the location of the consul server, and an optional key in consul that contains the boostrapseed nodes list.

discovery-consul/src/main/resources/reference.conf
winguse: @chbatey we are using this piece of code in our production by a separated file. And as for now, it has been over a year.

For this line, are you suggestion it's better to change it like this:

```suggestion
    service-suffix = "-akka-management"
```

or

```suggestion
    management-service-suffix = "-some-other-suffix"
```

?

winguse:
@chbatey  any update for this?

johanandren:
Hi @winguse, thanks for this contribution, since we got no volunteers stepping up to maintain the consul integration and review PRs for it for quite a long while, I'll close this PR for now.

Issue #411, opened 2018-12-10
Provide default service-name configuration through environment variable


jroper:
Just as the Kubernetes API namespace is looked up from a `$NAMESPACE` environment variable if present, we can do the same for the cluster bootstrap service name, by providing a `$SERVICE_NAME` environment variable. When the kubernetes-api method for discovery is being used, this can be supplied to the pod using:

```yaml
            - name: SERVICE_NAME
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: "metadata.labels['app']"
```

Note though that in documenting this, we need to be clear about when this can and can't be used. It works for the `kubernetes-api` discovery method because that looks up pods by their labels, rather than services. It doesn't work for the the `async-dns` discovery method, since that uses services, whose names don't necessarily have any relation to the pods labels, and indeed usually the service wanted is suffixed with `-internal`.

jroper:
Also, I've researched as to whether this could be discovered automatically, in a method similar to documented in #345, and the answer is not really. There's no automatic place that the labels are exposed to the container. You can supply the labels to the container on the filesystem as documented [here](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/), but that's 7 lines of YAML to define the volume, and 4 lines of YAML to define the volume mount, the environment variable is a lower config overhead, much simpler to implement, and conceptually simpler too.

chbatey:
It is based on more convention that we have a label called app so would be good to document this but not rely on it. For DNS discovery mechanism in k8s it'll depend on the internal service name which the pods know nothing about.

jroper:
Yes, but there's nothing in the code that we'll write that uses the app label, just the documentation that will pass the app label, so we can mention that in the docs. That said, there is a place where we're assuming the app label already, that is in the selector configuration. I think that's ok, all the Kubernetes docs use the app label, I think it's more than a convention, it's a defacto standard. We just need to explicitly state that we rely on it in our default configuration, and how to change if necessary.

TimMoore:
Calling `service-name` seems good to me. Calling it `actor-system` could be misleading, since it's not actually coupled to the name of the actor system (though it is usually a good idea to align them for comprehensibility).

patriknw:
Look into if this should be part of 1.0

raboof:
[...]

PR #500, opened 2019-03-06
Pass in bootstrap service name through environment variable

raboof:
By passing it in as an environment variable, the values that should line up for
bootstrap to be successful are all in the deployment .yml, instead of being
spread across the deployment .yml and the packaged application.conf.

Refs #411

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -25,8 +25,10 @@ akka.management {
     contact-point-discovery {
 
       # Define this name to be looked up in service discovery for "neighboring" nodes
-      # If undefined, the name will be extracted from the ActorSystem name
+      # If undefined, the name will be taken from the AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME
+      # environment variable or extracted from the ActorSystem name
       service-name = "<service-name>"
+      service-name = ${?AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME}
 
       # The portName passed to discovery. This should be set to the name of the port for Akka Management
       # If set to "" None is passed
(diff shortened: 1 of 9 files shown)


integration-test/dns-api-mesos/src/main/resources/application.conf
```diff
@@ -37,7 +37,8 @@ akka {
       contact-point-discovery {
 
         # Define this name to be looked up in service discovery for "neighboring" nodes
-        # If undefined, the name will be extracted from the ActorSystem name
+        # If undefined, the name will be taken from the AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME
+        # environment variable or extracted from the ActorSystem name
```
TimMoore: The comment doesn't seem to match the definition.

raboof: How do you mean? If `service-name` is undefined in this `application.conf` it will fall back to the `reference.conf` above, where `service-name` is taken from the `AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME` environment variable if filled, or otherwise falls back to `<service-name>` which gets replaced by the actor system name?

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -25,8 +25,10 @@ akka.management {
     contact-point-discovery {
 
       # Define this name to be looked up in service discovery for "neighboring" nodes
-      # If undefined, the name will be extracted from the ActorSystem name
+      # If undefined, the name will be taken from the AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME
+      # environment variable or extracted from the ActorSystem name
       service-name = "<service-name>"
+      service-name = ${?AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME}
```
chbatey: :+1:

integration-test/kubernetes-api-java/kubernetes/akka-cluster.yml
```diff
@@ -38,6 +38,15 @@ spec:
         - name: management
           containerPort: 8558
           protocol: TCP
+        env:
+        # The Kubernetes API discovery will use this service name to look for
```
chbatey: :+1:

PR #502, opened 2019-03-07
Remove pod-port-name option from kubernetes-api discovery

raboof:
It does not make sense to configure a default at this level. It might make
sense to configure a default at the cluster-bootstrap level.

If no port name is specified in the query, return all ports.

Simplification triggered by the discussion in #464

### DIFF
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
@@ -34,14 +34,19 @@ object KubernetesApiServiceDiscovery {
    * to do that.
    */
   private[kubernetes] def targets(podList: PodList,
-                                  portName: String,
+                                  portName: Option[String],
                                   podNamespace: String,
                                   podDomain: String): Seq[ResolvedTarget] =
     for {
       item <- podList.items
       if item.metadata.flatMap(_.deletionTimestamp).isEmpty
       container <- item.spec.toVector.flatMap(_.containers)
-      port <- container.ports.getOrElse(Seq.empty).find(_.name.contains(portName))
+      port <- portName match {
+        case None =>
+          container.ports.getOrElse(Seq.empty)
+        case Some(name) =>
+          container.ports.getOrElse(Seq.empty).filter(_.name.contains(name))
+      }
       itemStatus <- item.status
       ip <- itemStatus.podIP
       host = s"${ip.replace('.', '-')}.${podNamespace}.pod.${podDomain}"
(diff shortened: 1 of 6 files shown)


discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
```diff
@@ -34,14 +34,19 @@ object KubernetesApiServiceDiscovery {
    * to do that.
    */
   private[kubernetes] def targets(podList: PodList,
```
patriknw: InternalApi annotation and doc comment

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
```diff
@@ -30,18 +31,26 @@ import akka.event.Logging
...
-                                  portName: String,
+                                  portName: Option[String],
                                   podNamespace: String,
                                   podDomain: String): Seq[ResolvedTarget] =
     for {
       item <- podList.items
       if item.metadata.flatMap(_.deletionTimestamp).isEmpty
       container <- item.spec.toVector.flatMap(_.containers)
-      port <- container.ports.getOrElse(Seq.empty).find(_.name.contains(portName))
+      port <- portName match {
```
chbatey: So if using service discovery not with bootstrap and not passing in a port this will result in all the ports being returned?

docs/src/main/paradox/discovery/kubernetes.md
```diff
@@ -72,9 +69,9 @@ spec:
         - name: remoting
           containerPort: 2552
           protocol: TCP
-        # akka-management bootstrap
-        # This name must match the name defined in
-        # akka.discovery.kubernetes-api.pod-port-name configuration
+        # When
```
chbatey: Why is this when and not a mandatory property? Won't we get back all the ports if None is passed in, does that ever make sense?

raboof: In the non-discovery use case, indeed we get all ports back if None is passed in, which might make sense in situations where a service opens a single port, so the service name uniquely identifies the service and no port name is needed to discover it.

In the discovery case, we either 1) look for the `management` port name or 2) look for the port akka-management is configured to listen on. I think I prefer '1', but also prototyped '2' in #503 to see how that turns out.

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
raboof: Yes. Indeed this only really makes sense when the service only exposes one port, or when the results are post-processed like in #503 .

docs/src/main/paradox/discovery/kubernetes.md
chbatey: :+1:

PR #503, opened 2019-03-07
By default assume akka-management.http.port for bootstrap

raboof:
Refs #464: Instead of discovering by port name 'management', avoid a
named port lookup.

When multiple results are found, disambiguate based on 
the `fallback-port` which by default is populated with
`akka-management.http.port`.

This arguably makes discovery on Kubernetes a bit easier, since the
akka-management port does not need to be named any specific way in the
deployment .yml.

On the other hand, it makes the actual bootstrap/discovery process a bit
more complicated and thus harder to explain or troubleshoot.

I'm not convinced the convenience of not having to declare the port name in
the Kubernetes case weighs up to the extra complexity in the
bootstrap/discovery logic.

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -101,8 +101,8 @@ akka.management {
 
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
       # Also, when no port-name is used and multiple results are returned for a given service, this port is
-      # used to disambiguate. When 0, defaults to the value of akka.management.http.port
-      fallback-port = 0 # port pun, it "complements" 2552 which is often used for Akka remoting
+      # used to disambiguate. When set to <fallback-port>, defaults to the value of akka.management.http.port
+      fallback-port = "<fallback-port>" # port pun, it "complements" 2552 which is often used for Akka remoting
 
       # If some discovered seed node will keep failing to connect for specified period of time,
       # it will initiate rediscovery again instead of keep trying.
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -100,7 +100,9 @@ akka.management {
     contact-point {
 
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
-      fallback-port = 8558 # port pun, it "complements" 2552 which is often used for Akka remoting
+      # Also, when no port-name is used and multiple results are returned for a given service, this port is
+      # used to disambiguate. When 0, defaults to the value of akka.management.http.port
+      fallback-port = 0 # port pun, it "complements" 2552 which is often used for Akka remoting
 
       # If some discovered seed node will keep failing to connect for specified period of time,
       # it will initiate rediscovery again instead of keep trying.
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -31,8 +31,8 @@ akka.management {
       service-name = ${?AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME}
 
       # The portName passed to discovery. This should be set to the name of the port for Akka Management
-      # If set to "" None is passed
-      port-name = "management"
+      # If set to "" None is passed and ${akka.management.http.port} is assumed.
+      port-name = ""
 
       # The protocol passed to discovery.
       # If set to "" None is passed.
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
@@ -123,7 +123,11 @@ final class ClusterBootstrapSettings(config: Config, log: LoggingAdapter) {
     private val contactPointConfig = bootConfig.getConfig("contact-point")
 
     // FIXME this has to be the same as the management one, we currently override this value when starting management, any better way?
-    val fallbackPort: Int = contactPointConfig.getInt("fallback-port")
+    val fallbackPort: Int =
+      contactPointConfig
+        .optDefinedValue("fallback-port")
+        .map(_.toInt)
+        .getOrElse(config.getInt("akka.management.http.port"))
 
     val probingFailureTimeout: FiniteDuration =
       contactPointConfig.getDuration("probing-failure-timeout", TimeUnit.MILLISECONDS).millis
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
@@ -122,7 +122,6 @@ final class ClusterBootstrapSettings(config: Config, log: LoggingAdapter) {
   object contactPoint {
     private val contactPointConfig = bootConfig.getConfig("contact-point")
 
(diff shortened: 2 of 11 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -250,7 +252,23 @@ private[akka] final class BootstrapCoordinator(discovery: ServiceDiscovery,
...
+            resolved.serviceName,
+            resolved.addresses.map(target => ResolvedTarget(target.host, Some(defaultManagementPort), target.address))
+          )
+        else
+          Resolved(
+            resolved.serviceName,
+            resolved.addresses.filter(_.port.contains(defaultManagementPort))
+          )
+      }
+      .pipeTo(self)
```
TimMoore: I think it could be a lot simpler if the fallback logic happens at the point of probing. In fact, it looks like there is already some kind of partially-implemented fallback logic there:

https://github.com/akka/akka-management/blob/9c97f6aa830db1da5f5fdff7febe07d7883b00a9/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L292

https://github.com/akka/akka-management/blob/23bc509c6d45df9c26d0b1b8f67bf9bf92846d95/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala#L125-L126

https://github.com/akka/akka-management/blob/03a30b1b06c1dbc0e3e3522ee28872bec6e8b4ec/cluster-bootstrap/src/main/resources/reference.conf#L97-L101

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -174,8 +174,21 @@ private[akka] final class BootstrapCoordinator(discovery: ServiceDiscovery,
...
-      onContactPointsResolved(contactPoints)
+      val filteredContactPoints: Iterable[ResolvedTarget] =
+        if (lookup.portName.isDefined)
+          contactPoints
+        else
+          contactPoints.groupBy(_.host).flatMap {
+            case (host, immutable.Seq(singleResult)) =>
+              immutable.Seq(singleResult)
+            case (host, multipleResults) =>
+              println(s"Looking for ${settings.contactPoint.fallbackPort}")
```
patriknw: println

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -174,8 +174,21 @@ private[akka] final class BootstrapCoordinator(discovery: ServiceDiscovery,
...
+              multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
+          }
+
+      log.info("Located service members based on: [{}]: [{}], filtered to [{}]", lookup, contactPoints.mkString(", "),
```
patriknw: should this be info? done for each probing?

raboof: It was already logged each time the probing endpoints are discovered (it just moved around a bit in this PR) and did come in handy when diagnosing problems, so I think it is nice to keep this at INFO at least for now.

patriknw: ok

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -29,8 +29,8 @@ akka.management {
       service-name = "<service-name>"
 
       # The portName passed to discovery. This should be set to the name of the port for Akka Management
-      # If set to "" None is passed
-      port-name = "management"
+      # If set to "" None is passed and ${akka.management.http.port} is assumed.
+      port-name = ""
```
chbatey: Ah i see from the other PR now.

chbatey: But won't we then get back pods that match the label selector that don't have akka management enabled? This could result in some confusing connection errors. Perhaps we filter down below, will see

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -98,7 +98,9 @@ akka.management {
     contact-point {
 
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
-      fallback-port = 8558 # port pun, it "complements" 2552 which is often used for Akka remoting
+      # Also, when no port-name is used and multiple results are returned for a given service, this port is
+      # used to disambiguate. When set to <fallback-port>, defaults to the value of akka.management.http.port
+      fallback-port = "<fallback-port>" # port pun, it "complements" 2552 which is often used for Akka remoting
```
chbatey: pun?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrapSettings.scala
```diff
@@ -122,8 +122,11 @@ final class ClusterBootstrapSettings(config: Config, log: LoggingAdapter) {
   object contactPoint {
     private val contactPointConfig = bootConfig.getConfig("contact-point")
 
-    // FIXME this has to be the same as the management one, we currently override this value when starting management, any better way?
-    val fallbackPort: Int = contactPointConfig.getInt("fallback-port")
+    val fallbackPort: Int =
```
chbatey: fallbackPort still a good name if we set it to the akka management port?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -174,8 +174,20 @@ private[akka] final class BootstrapCoordinator(discovery: ServiceDiscovery,
...
-      onContactPointsResolved(contactPoints)
+      val filteredContactPoints: Iterable[ResolvedTarget] =
+        if (lookup.portName.isDefined)
+          contactPoints
+        else
+          contactPoints.groupBy(_.host).flatMap {
+            case (host, immutable.Seq(singleResult)) =>
+              immutable.Seq(singleResult)
+            case (host, multipleResults) =>
+              multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
```
chbatey: Aha so we filter here :+1:

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -174,8 +174,20 @@ private[akka] final class BootstrapCoordinator(discovery: ServiceDiscovery,
...
     case ServiceDiscovery.Resolved(_, contactPoints) ⇒
-      log.info("Located service members based on: [{}]: [{}]", lookup, contactPoints.mkString(", "))
-      onContactPointsResolved(contactPoints)
+      val filteredContactPoints: Iterable[ResolvedTarget] =
+        if (lookup.portName.isDefined)
+          contactPoints
+        else
+          contactPoints.groupBy(_.host).flatMap {
+            case (host, immutable.Seq(singleResult)) =>
+              immutable.Seq(singleResult)
```
chbatey: this may not be the akka management port, could be a user http port if the pod label selector picks up pods from different deployments. Not the normal but user could easily do it

Issue #520, opened 2019-03-28
Kubernetes discovery discovers non-running pods which can break cluster bootstrap


fredfp:
The discovery api discovers pods which are not running and not deleted.

This is problematic (at least) for cluster bootstrap because the IP address of the not running and not deleted pod can be allocated to another running pod.

I had a case where the IP was currently used by a running pod which was in another namespace, and which was member of another akka cluster. The bootstrap algorithm chose to connect to that pod which lead to a failure (because the actor system name and the remoting used were different).

chbatey:
Thanks for reporting. It also is problematic as bootstrap defaults to requiring responses from each of the nodes.

PR #521, opened 2019-03-28
Kubernetes discovery api: only consider running pods

fredfp:
fix for https://github.com/akka/akka-management/issues/520

### DIFF
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
@@ -19,7 +19,7 @@ import akka.annotation.InternalApi
 
   final case class PodSpec(containers: immutable.Seq[Container])
 
-  final case class PodStatus(podIP: Option[String])
+  final case class PodStatus(podIP: Option[String], phase: Option[String])
 
   final case class Pod(spec: Option[PodSpec], status: Option[PodStatus], metadata: Option[Metadata])
 }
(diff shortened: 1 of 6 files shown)


discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/PodList.scala
```diff
@@ -19,7 +19,7 @@ import akka.annotation.InternalApi
 
   final case class PodSpec(containers: immutable.Seq[Container])
 
-  final case class PodStatus(podIP: Option[String])
+  final case class PodStatus(podIP: Option[String], phase: Option[String])
```
raboof: Why an `Option`? Is it not guaranteed to be there?

(I guess the same could be said about the `podIP`, though)

fredfp: I wondered the same and preferred to be on the safe side

chbatey:
Thanks, this is definitely the right thing to do.

Does bootstrap work if it gets no contact points back? Thinking of initial deployment when no pods will be running.

fredfp:
Is it even possible that no contact point is returned? I don't think so. From [pod lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase), a pod will be `Running` when `The Pod has been bound to a node, and all of the Containers have been  created. At least one Container is still running, or is in the process  of starting or restarting`. As such the initial pod should at least discover itself and bootstrapping with `contact-point-discovery.required-contact-point-nr` set to `1` should work.

chbatey:
Sorry I got confused between Running and healthy when I wrote that. LGTM

Issue #525, opened 2019-04-07
start() does not always return the correct port


fredfp:
It is allowed to set `akka.management.http.port = 0`, I'm assuming this means the same as usual: choose a random port. Binding a random port is very useful for tests as it allows to avoid port conflicts for tests running in parallel.

However when setting `akka.management.http.port = 0`, the uri returned by `start()` does not contain the randomly selected port, but port 80 instead. As such the returned uri cannot be used to connect to the akka management server and is useless.

chbatey:
:+1:

PR #526, opened 2019-04-07
Return bound port when port was randomly selected

fredfp:
Fix for https://github.com/akka/akka-management/issues/525


raboof:
Travis failure seems unrelated, filed #527 and retriggered

PR #528, opened 2019-04-19
Uses networkBindings to retrieve IP and port

renatocaval:
This relates with https://github.com/lagom/lagom/issues/1873

Not sure if network bindings is providing what is really needed in that case, but seems to be what we are looking for.


renatocaval:
Test failed with unrelated error. Flakky?

renatocaval:
I'm checking all pending PRs I have and found this one. @chbatey, does the change suggested here makes any sense?

See discussion at https://github.com/lagom/lagom/issues/1873 as well for background info.

gdrago22123:
Tried the code in the PR, but noticed that AWS keeps returning `NetworkBindings=[]`in the task description. I searched the task description result, hoping to see the port listed in some other section of the result, but it was nowhere to be found :(

gdrago22123:
Be happy to test any other possible solutions

chbatey:
I don't know about ECS so not sure what would make sense for port. I'll close this for now and we can discuss in an issue what makes sense for the discovery impl to do

Issue #527, opened 2019-04-15
should return shard region details when calling GET /


raboof:
Probably just travis not meeting the 1-second deadline?

```
should return shard region details
[info]   - when calling GET /cluster/shard_regions/{name} *** FAILED *** (3 seconds, 800 milliseconds)
[info]     java.util.concurrent.TimeoutException: Futures timed out after [1 second]
```

PR #536, opened 2019-05-06
Enable logging when running consul tests

raboof:
Consul appears to use slf4j directly


raboof:
Failures are #537 and #527

PR #541, opened 2019-05-09
Don't return a port from kubernetes-api when none requested

### LINKED ISSUE #538: Allow kubernetes-api discovery without declaring ports in the pod spec
Currently, kubernetes-api discovery only returns results that have the management port configured in the pod spec. This presents a problem for Istio if you want to bypass envoy (see https://github.com/akka/akka-management/issues/209#issuecomment-490764774), and you do want to bypass envoy because you want to use readiness checks and you don't want envoy shutting down to kill your connections to the other nodes in the cluster. The reason it's a problem for Istio is that Istio routes all ports declared in your pod through envoy.

To solve this, it needs to be possible to run kubernetes api discovery in a mode such that it doesn't look at the declared pod ports, it simply returns a list of all pods that match the selector, and assumes that the correct port is the same port that the local machine uses, and so returns that.

jroper:
Fixes #538

If no port is requested from kubernetes-api (because none is configured), it should just return one target per pod, with no port set.  Cluster bootstrap already falls back to using the configured Akka HTTP port if no port is returned from the lookup, so this will work.

I have tested this, and confirm it works (in particular, it works in an Istio environment where the ports can't be specified in the container spec).

### DIFF
--- a/discovery-kubernetes-api/src/test/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscoverySpec.scala
+++ b/discovery-kubernetes-api/src/test/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscoverySpec.scala
@@ -6,8 +6,8 @@ package akka.discovery.kubernetes
 
 import java.net.InetAddress
 
-import org.scalatest.{ Matchers, WordSpec }
-import PodList._
+import org.scalatest.{Matchers, WordSpec}
+import PodList.{Pod, _}
 import akka.actor.ActorSystem
 import akka.discovery.Discovery
 import akka.discovery.ServiceDiscovery.ResolvedTarget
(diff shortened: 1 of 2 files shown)


discovery-kubernetes-api/src/test/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscoverySpec.scala
```diff
@@ -45,6 +45,44 @@ class KubernetesApiServiceDiscoverySpec extends WordSpec with Matchers {
         "cluster.local") shouldBe List.empty
     }
 
+    "return a single result per host with no port when no port name is requested" in {
```
raboof: Great to add this test. Might be good to document in a little more detail *why* this behavior is desired. Perhaps even explicitly link to https://github.com/jroper/akka-grpc-ddata-shopping-cart/tree/istio-future and explain it is needed for that use case?

jroper: Document where? In the documentation? Or in comments?

I actually think this is a bug fix, regardless of why it's being implemented. If you do a lookup on any arbitrary discovery implementation, and don't specify any port in the lookup, do you expect to get a port back in the resolved target? I would have thought no, you certainly won't get one back from the dns implementation, it only does a SRV lookup if you specify a port. So the same should go for the kubernetes-api implementation, I just expect it to return me a list of hosts without ports if I don't specify a port to lookup. So I'm not sure that any special explanation is needed, it's now behaving intuitively, if not according to spec, whereas before it was behaving unintuitively, and that was probably a bug.

I think once the Istio patch is merged, we should add some documentation (perhaps to the OpenShift guide) on how to form an Akka cluster in Istio, and the need to not specify a port can be mentioned there. Otherwise I don't think there's any need to document it.

raboof: > Document where? In the documentation? Or in comments?

I was thinking of in a comment along with this test

> If you do a lookup on any arbitrary discovery implementation, and don't specify any port in the lookup, do you expect to get a port back in the resolved target

In some other context I might desire to get back *all* the ports. After all, devils' advocate, what good is getting back a service IP if I still don't receive enough information to actually connect to it? Especially if the information is readily available from the discovery backend? If someone asks for that feature in the future, this test will fail. I would like to explain to future-me that this test doesn't just describe a malleable status-quo, but there are actually use cases depending on this behavior.

What's 'intuitive' can change a lot based on context ;).

jroper: Fair points. I've added the comment.

Issue #453, opened 2019-01-23
Akka remoting bootstrap probing


chbatey:
Probing is currently via HTTP. 

For applications that don't use HTTP e.g. communication via Kafka it may be  inconvenient to open up a HTTP port for cluster bootstrap.

Probing via remoting also means that bootstrap could move to core Akka.

Low priority until users request a non HTTP version of bootstrap.

patriknw:
I agree, it would be possible but since we have a working solution it's not top priority to change it.

I'm trying to recall why we selected http for the probing. I can't find any strong reasons. A few that comes to mind but not anything critical:
- we already had cluster http management, and that is what we prefer as the management/insight to the cluster, tagging along on that http server that we prefer to have active anyway
- http needed for health checks anyway
- remoting connections are long lived while http requests fits the probing use case better, but the end goal is to join the discovered nodes anyway and then the remoting connections will be established anyway
- maybe some fear that some of the discovered nodes would not be real candidates for the cluster and that could only be observed after probing, and therefore it would be unnecessary to open "heavy" remoting connections to those nodes (I don't think we have that feature currently, all discovered nodes should be real candidates for the cluster)
- for large clusters, many discovered nodes, there might be a performance problem to create remote connections to all of them at the same time at startup (once joining the cluster those connections are created more gradually as needed)

PR #546, opened 2019-05-20
WIP - Cluster bootstrap remoting probe method

jroper:
Some initial work on #453. This has been done in an effort to try and slim down cluster bootstrap, both in terms of complexity and the amount of resources it uses.

This works, but there are a few things to think about.

1. With this in place, we could decouple cluster bootstrap and Akka management HTTP. This would require pulling the HTTP probing method out into its own module, which would effectively make remoting the default.
2. Currently the serializer for the seed nodes reuses the JSON serializers - we might need to reconsider this and use protobuf instead, this would in particular be a good idea if we decouple from Akka management HTTP as the JSON serialization means keeping a dependency on spray-json which could otherwise be dropped.
3. While I've implemented unit tests that show that bootstrapping a cluster of 3 actor systems in the one process works, I haven't implemented any integration tests, eg on k8s, yet.

One nice thing is that remoting doesn't need the same hack that is used to discover the Akka management HTTP bind address.

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -99,9 +99,14 @@ akka.management {
     # Configured how we communicate with the contact point once it is discovered
     contact-point {
 
+      # The probe method. Valid values are akka-management and remoting. If akka-management, will use akka-management
+      # HTTP interface to discover seed nodes. If remoting, will use Akka remoting to discover seed nodes.
+      probe-method = "akka-management"
+
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
       # Also, when no port-name is used and multiple results are returned for a given service, this port is
-      # used to disambiguate. When set to <fallback-port>, defaults to the value of akka.management.http.port
+      # used to disambiguate. When set to <fallback-port>, defaults to the value of akka.management.http.port.
+      # Only used by http probe method, for remoting, the cluster remoting port is used.
       fallback-port = "<fallback-port>" # port pun, it "complements" 2552 which is often used for Akka remoting
 
       # If some discovered seed node will keep failing to connect for specified period of time,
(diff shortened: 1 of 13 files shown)


jroper:
Any thoughts on the points raised above? Should this go the full way and decouple cluster bootstrap from Akka management HTTP? Should the spray-json serializers used by remoting be replaced with protobuf serializers?

Also, currently the probing method has two hard coded options and is not pluggable, but it could be made such that we extract an interface that the two options can implement, and then allow other probing methods to be introduced. Not sure if that's a good idea or not, it would increase the effective public API and required semantics, whereas if we keep it as two hardcoded possibilities then that stuff can all stay internal which allows us to modify it as we see fit going forward.

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -99,9 +99,14 @@ akka.management {
     # Configured how we communicate with the contact point once it is discovered
     contact-point {
 
+      # The probe method. Valid values are akka-management and remoting. If akka-management, will use akka-management
+      # HTTP interface to discover seed nodes. If remoting, will use Akka remoting to discover seed nodes.
+      probe-method = "akka-management"
```
patriknw: why do we want to support both methods?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/RemotingContactPointBootstrap.scala
```diff
@@ -0,0 +1,49 @@
...
+  contactPoint: ResolvedTarget
+) extends AbstractContactPointBootstrap(settings, contactPoint) {
+
+  override protected val uri: String = {
+    val cluster = Cluster(context.system)
+    val targetPort = contactPoint.port
+      .orElse(cluster.selfAddress.port)
+      .getOrElse(throw new IllegalArgumentException("Cannot infer port for contact point"))
+    val address = cluster.selfAddress.copy(host = Some(contactPoint.host), port = Some(targetPort))
+    (self.path.parent.parent / RemotingContactPoint.RemotingContactPointActorName).toStringWithAddress(address)
```
patriknw: might be better with `RootActorPath`:
```
RootActorPath(address) / "system" / RemotingContactPoint.RemotingContactPointActorName
```

cluster-bootstrap/src/main/resources/reference.conf
jroper: You won't be able to do a rolling upgrade to the new method if you don't support both concurrently.

PR #547, opened 2019-05-20
Update RELEASING.md

chbatey:
No longer need to update whitesource as 1.0


raboof:
(Now conflicted because the release steps moved to the release train issue template)

PR #550, opened 2019-05-22
Use a global JDK env var, set to adopt@1.8.192-12

dwijnand:


dwijnand:
A fight for another day.

PR #551, opened 2019-05-22
Cache sbt's launcher jars

dwijnand:


raboof:
(failure was failing to download jabba from github, restarted)

PR #555, opened 2019-05-27
Move 'port pun' comment to the right place

raboof:


raboof:
yeah would be OK as well ;)

PR #557, opened 2019-05-29
Cross build to Scala 2.13.0-RC2

renatocaval:


ignasi35:
> Not sure we'll have trouble with akka-http being available only for jdk11?

`[error] java.lang.UnsupportedClassVersionError: akka/http/javadsl/model/HttpHeader has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0`

We do. :-(

raboof:
I'll publish for RC3 with jdk8

ignasi35:
> I'll publish for RC3 with jdk8

We're not even there for RC2. :-(

project/Dependencies.scala
```diff
@@ -3,15 +3,16 @@ import Keys._
 
 object Dependencies {
 
-  val AkkaVersion = "2.5.21"
-  val AkkaHttpVersion = "10.1.7"
+  val AkkaVersion = "2.5.23"
+  val AkkaHttpVersion = "10.1.8"
```
dwijnand: Use https://dl.bintray.com/akka/snapshots/com/typesafe/akka/akka-http-core_2.13.0-RC2/10.1.8+26-f33ec39a/ (or the latest https://dl.bintray.com/akka/snapshots/com/typesafe/akka/akka-http-core_2.13.0-RC2/10.1.8+26-f33ec39a/) so it works on JDK 8.

You need to add `resolvers in ThisBuild += Resolver.bintrayRepo("akka", "snapshots")`

Here's how to do that in a way that makes it use that version only for Scala 2.13:

```scala
  val akkaHttpVersion = Def.setting {
    val sv = scalaVersion.value
    CrossVersion.partialVersion(sv) match {
      case Some((2, 13)) => "10.1.8+26-f33ec39a"
      case _             => "10.1.8"
    }
  }
```

But you have to use `akkaHttpVersion.value` inside `Def.setting { .. }` or move the setting.  e.g. https://github.com/playframework/playframework/commit/98588e3725668aa6e36dd8b4f7ee26b77a6ab5cd#diff-0ecdbc5a001d52fb34f5eafb7cd1aaa6L334

raboof:
I'll push the update suggested by @dwijnand

raboof:
(1.0.1 for RC2 is now on Maven Central)

raboof:
reviewed by @ignasi35 via #564

raboof:
(failure on master was #566)

PR #558, opened 2019-05-30
Use ServiceAccount subject kind rather than user

jroper:
This is much simpler from a configuration perspective as it doesn't require understanding the service account user name string format, and it also allows the namespace to default to the namespace that the role binding is deployed to, which is usually what you want, and so means that role binding yaml specs can be used as is regardless of which namespace they are deployed to.

One caveat, I could not find any documentation on what should happen when no namespace is specified. Intuitively, it should use the namespace the role binding is deployed to, and I have confirmed through experimentation that this appears to be the case. I've raised this issue to get this documented:

https://github.com/kubernetes/website/issues/14624

### DIFF
--- a/docs/src/main/paradox/bootstrap/kubernetes-api.md
+++ b/docs/src/main/paradox/bootstrap/kubernetes-api.md
@@ -15,7 +15,7 @@ An example deployment (used for integration testing):
 
 An example role and rolebinding to allow the nodes to query the Kubernetes API server:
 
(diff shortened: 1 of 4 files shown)


jroper:
The initial comments on https://github.com/kubernetes/website/issues/14624 don't disagree with my understanding of how this works, and I've been using this in a few places now and it works for me.

TimMoore:
I just pushed a new commit that updates the tests, and also updates the documentation to include the snippet from the tests.

PR #563, opened 2019-06-07
Upgrade akka version to 2.5.23

skvithalani:


raboof:
Typically we don't update the Akka dependency of satellite projects like akka-management unless there is a reason to do so: after all, because of our binary compatibility approach, you can always use akka-management with newer versions of Akka, but not with earlier.

Unfortunately we don't have this documented clearly anywhere yet, that's tracked in https://github.com/akka/akka/issues/26985

Is there any particular reason you'd like to see the Akka dependency updated here?

skvithalani:
@raboof thanks for the quick reply. We are on akka 2.5.23 and akka-management at 1.0.1 with akka 2.5.21. So we are getting a warning from akka for incompatible types. It is failing a test for us. We can have a work-around to explicitly depend on akka module 2.5.23 which are transitively coming from akka-management. But this did not seem like a clean way of doing things. Hence, this PR to upgrade akka-management to 2.5.23.

raboof:
We actually recommend depending on the latest version of Akka explicitly in your project., rather than having (possibly inconsistent) versions come in as transitive dependencies. Our binary compatibility approach makes this safe, as long as you don't update the sattelite libraries without also updating Akka itself if needed.

skvithalani:
@raboof ok in that case we will make it(akka-cluster-shard and akka-persisence) explicit. But we did not needed this dependencies in our project in the first place. It is only to solve the compatibility issue, we will have it explicit in our build.

chbatey:
> We can at-least have the versioning scheme of akka-management written on it github readme page I guess.

Good idea, and in the docs. I'll close this PR for now but would you be up for rasing a PR for readme and docs?

PR #564, opened 2019-06-11
Update to Scala 2.13.0

raboof:
Supersedes #562 

To keep attribution, let's merge #557 first

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,5 @@
 ThisBuild / resolvers += Resolver.jcenterRepo
 
-resolvers in ThisBuild += Resolver.bintrayRepo("akka", "snapshots")
 
 // root
 lazy val `akka-management-root` = project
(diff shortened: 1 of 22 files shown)


raboof:
1.0.1 for Scala 2.13 is on its way to Maven Central now

build.sbt
```diff
@@ -1,5 +1,7 @@
 ThisBuild / resolvers += Resolver.jcenterRepo
 
+resolvers in ThisBuild += Resolver.bintrayRepo("akka", "snapshots")
```
ignasi35: This can be removed already, correct?

ignasi35:
```suggestion

```

ignasi35:
Both  #557 and this seem ready to merge :-)

raboof:
rebased as #567

Issue #584, opened 2019-08-10
Release 1.0.2 incampatible with jdk8


jasonqu:
### Versions used 

Akka version: 2.5.22
akka-management version: 1.0.2
java version: 8

```
"com.lightbend.akka.management" %% "akka-management" % "1.0.2",
"com.lightbend.akka.management" %% "akka-management-cluster-http" % "1.0.2",
```

```
java -version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
```

### Actual Behavior

crashed when access via http

```
curl -i http://127.0.0.1:8558/cluster/members/ 
curl: (56) Failure when receiving data from the peer
```

### Relevant logs

```
[ERROR] [SECURITY][08/10/2019 17:22:40.080] [ChannelJobSystem-akka.actor.default-dispatcher-23] [akka.actor.ActorSystemImpl(ChannelJobSystem)] Uncaught error from thread [ChannelJobSystem-akka.actor.default-dispatcher-23]: akka/management/cluster/ClusterReadViewAccess has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[ChannelJobSystem]
java.lang.UnsupportedClassVersionError: akka/management/cluster/ClusterReadViewAccess has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
        at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[...]

PR #588, opened 2019-08-14
Target JDK8 for Java classes

raboof:
Refs #584

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -43,8 +43,12 @@ object Common extends AutoPlugin {
           "-target:jvm-1.8"
         ),
       javacOptions ++= Seq(
-          "-Xlint:unchecked"
+          "-Xlint:unchecked",
         ),
+      javacOptions ++= (
+        if (isJdk8) Seq.empty
+        else Seq("--release", "8")
+      ),
       Compile / doc / scalacOptions := scalacOptions.value ++ Seq(
           "-doc-title",
           "Akka Management",
@@ -67,4 +71,7 @@ object Common extends AutoPlugin {
       testOptions += Tests.Argument(TestFrameworks.JUnit, "-v", "-a"),
       scalaVersion := "2.12.8"
     )
+
+  private def isJdk8 =
+    VersionNumber(sys.props("java.specification.version")).matchesSemVer(SemanticSelector(s"=1.8"))
 }


olijaun:
I'm sorry I don't want to push... but this seems like a serious problem. I had to downgrade from 1.0.4 to  1.0.1 of akka-management. I think that JDK 8 is still widely in use. It would be great if this could be merged. Thanks

raboof:
@olijaun thanks for the reminder! Just to check: are you having trouble with `akka.management.cluster.ClusterReadViewAccess` or some other class?

olijaun:
@raboof Yes. Here is an example when I run `curl http://127.0.0.1:8558/cluster/members`

`
[INFO] [11/12/2019 08:35:07.044] [my-cluster-akka.actor.default-dispatcher-21] [akka://my-cluster@127.0.0.1:2550/system/singletonProxymy-singleton-actor-no-dc] Singleton identified at [akka://my-cluster/system/singletonManagermy-singleton-actor/my-singleton-actor]
received message
Uncaught error from thread [my-cluster-akka.actor.default-dispatcher-13]: akka/management/cluster/ClusterReadViewAccess has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[my-cluster]
java.lang.UnsupportedClassVersionError: akka/management/cluster/ClusterReadViewAccess has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
`

raboof:
> So this relies on that we run the publish in travis with java 8?

The other way around: before this PR, `ClusterReadViewAccess.java` was built for whatever JDK you were publishing with. With this PR, it will target JDK8 even when publishing from JDK11.

Travis is already set up to publish with jdk8, but the travis release for 1.0.4 failed due to missing/outdated artifacts (https://travis-ci.org/akka/akka-management/jobs/607515807), so I suspect we released manually with JDK11 causing the problem @olijaun is seeing. This PR would prevent that from happening in the future.

chbatey:
Aha doh, I misread which way around the is was for isJDK8. Gotcha.

patriknw:
sounds important to publish 1.0.5 asap  if 1.04 doesn't run on jdk8

PR #596, opened 2019-10-02
Use wildcard minor version for jabba in travis

jrudolph:
The previous versions just don't work.


raboof:
failure should be #594

PR #600, opened 2019-10-09
Use specific logger for bootstrap

chbatey:


chbatey:
Closing this for now as we can use package name of the loggers

Issue #545, opened 2019-05-14
Run scalafmt on travis


raboof:
It seems scalafmt has not been configured to run automatically when compiling, and is not ran on travis explicitly either.

PR #604, opened 2019-10-21
Uniform URL scheme for bootstrap URLs

### LINKED ISSUE #603: Impossible to secure bootstrap traffic via HTTPS
The bootstrap HTTP enpoint is hard-coded to `http`, even though Akka exposes endpoints via `https` if a `ManagementRouteProviderSettings` is provided.

The protocol to access contact points (peers) is hard-coded here:

https://github.com/akka/akka-management/blob/1db736af3661fd98af299dd8bfc0dbbccff2900d/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L287-L288

And then passed to the `HttpContactPointBootstrap` here:

https://github.com/akka/akka-management/blob/1db736af3661fd98af299dd8bfc0dbbccff2900d/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L310

Which uses the URL as-is to obtain seed-nodes:
https://github.com/akka/akka-management/blob/1db736af3661fd98af299dd8bfc0dbbccff2900d/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala#L82

https://github.com/akka/akka-management/blob/1db736af3661fd98af299dd8bfc0dbbccff2900d/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala#L102

> On a side-note, I notice that this line does nothing, `req` is not used, and its a duplication of a previous line: https://github.com/akka/akka-management/blob/1db736af3661fd98af299dd8bfc0dbbccff2900d/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala#L99
>
> It's kind of misleading because the next line makes it look like it's using `req`, even though it's using `probeRequest` from 20 lines ago outside the block.

Setup:

```scala
  val sslConfig: AkkaSSLConfig =  AkkaSSLConfig().mapSettings(s => s.withLoose(s.loose.withDisableSNI(true)))
  val sslContext: SSLContext = Http().createServerHttpsContext(sslConfig).sslContext
  val https: HttpsConnectionContext = ConnectionContext.https(sslContext)
  AkkaManagement(system).start(_.withHttpsConnectionContext(https))
```

```
akka {
  ssl-config {
    keyManager = {
      stores = [
        {type = "PKCS12", path = "/tmp/keystore.p12", password = "foobar"}
      ]
    }
    trustManager = {
      stores = [
        {type = "PKCS12", path = "/tmp/truststore.p12", password = "foobar"}
      ]
    }
  }
}
```

The application correctly exposes HTTPS, and is contactable as such via the browser
```
[INFO] [10/21/2019 18:51:50.159] [main] [ClusterBootstrap(akka://ClusterSystem)] Using self contact point address: https://192.168.3.200:8558
```

But it tries to connect to other nodes via HTTP:
```
[WARN] [10/21/2019 18:51:59.660] [ClusterSystem-akka.actor.default-dispatcher-30] [akka.ssl.tcp://ClusterSystem@127.0.0.1:3553/system/bootstrapCoordinator/contactPointProbe-192.168.3.200-8558] Probing [http://192.168.3.200:8558/bootstrap/seed-nodes] failed due to: The connection closed with error: Connection reset by peer
```

This is where the self-address is changed to https, but I can't see any such provision for remote addresses

https://github.com/akka/akka-management/blob/fc8d1627fadb75eb0405d24232d5beb3054ebe89/management/src/main/scala/akka/management/javadsl/ManagementRouteProviderSettings.scala#L80-L83

https://github.com/akka/akka-management/search?q=scheme

That means, at the moment it's impossible to secure your bootstrap traffic via HTTPS, even if you manually put an SSL proxy in front of it, because Akka will only ever use the hard-coded `http` protocol. Kind of comes as a surprise that I'm apparently the first person to point this out.

For me, this as a very basic and essential security concern. I was hoping to set up mutual SSL authentication over HTTPS to secure access to the management endpoints.

otto-dev:
Fixes #603 

When bootstrapping, all nodes must necessarily use the same URL scheme for bootstrapping endpoints, either HTTP or HTTPS. Otherwise, the nodes that use HTTP do not see the nodes that use HTTPS, and vice versa.

Currently, the URL scheme is hard-coded to `"http"` which makes nodes that expose their endpoints via HTTPS uncontactable. This pull request makes it so that contact points will be contacted with the same scheme of the node's own base URL. That is, if the own node exposes it's management enpoints via HTTPS, then other nodes will also be contacted via HTTPS. This is based on the premise that the scheme must be uniform.

PS: I don't think this is a "backwards incompatible change" because change only affects cluster-bootstrap deployments that use HTTPS. But in the current version it's not possible to deploy bootstrap with HTTPS, because those nodes cannot be contacted due to the hard-coded `"http"` protocol. So in the broad scope this change only affects deployments that are already non-functional, so it's purely a fix, not a change.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -86,8 +86,9 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
       val bootstrapProps = BootstrapCoordinator.props(discovery, joinDecider, settings)
       val bootstrap = system.systemActorOf(bootstrapProps, "bootstrapCoordinator")
       // Bootstrap already logs in several other execution points when it can't form a cluster, and why.
-      val initiateBootstrapping = selfContactPoint.map(BootstrapCoordinator.Protocol.InitiateBootstrapping)
-      initiateBootstrapping pipeTo bootstrap
+      selfContactPoint.foreach {
+        uri => bootstrap ! BootstrapCoordinator.Protocol.InitiateBootstrapping(uri)
+      }
     } else log.warning("Bootstrap already initiated, yet start() method was called again. Ignoring.")
 
   /**
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -8,7 +8,7 @@ import java.util.concurrent.atomic.AtomicReference
 
 import akka.AkkaVersion
 
-import scala.concurrent.{Future, Promise, TimeoutException}
+import scala.concurrent.{ Future, Promise, TimeoutException }
 import scala.concurrent.duration._
 import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
@@ -17,7 +17,7 @@ import akka.actor.ExtensionId
 import akka.actor.ExtensionIdProvider
 import akka.annotation.InternalApi
 import akka.cluster.Cluster
-import akka.discovery.{Discovery, ServiceDiscovery}
+import akka.discovery.{ Discovery, ServiceDiscovery }
 import akka.event.Logging
 import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.server.Route
@@ -55,8 +55,10 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
 
   private val joinDecider: JoinDecider = {
     system.dynamicAccess
-      .createInstanceFor[JoinDecider](settings.joinDecider.implClass,
-        List((classOf[ActorSystem], system), (classOf[ClusterBootstrapSettings], settings)))
+      .createInstanceFor[JoinDecider](
+        settings.joinDecider.implClass,
+        List((classOf[ActorSystem], system), (classOf[ClusterBootstrapSettings], settings))
+      )
       .get
   }
 
@@ -72,10 +74,11 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
   def start(): Unit =
     if (Cluster(system).settings.SeedNodes.nonEmpty) {
       log.warning(
-          "Application is configured with specific `akka.cluster.seed-nodes`: {}, bailing out of the bootstrap process! " +
-          "If you want to use the automatic bootstrap mechanism, make sure to NOT set explicit seed nodes in the configuration. " +
-          "This node will attempt to join the configured seed nodes.",
-          Cluster(system).settings.SeedNodes.mkString("[", ", ", "]"))
+        "Application is configured with specific `akka.cluster.seed-nodes`: {}, bailing out of the bootstrap process! " +
+        "If you want to use the automatic bootstrap mechanism, make sure to NOT set explicit seed nodes in the configuration. " +
+        "This node will attempt to join the configured seed nodes.",
+        Cluster(system).settings.SeedNodes.mkString("[", ", ", "]")
+      )
     } else if (bootstrapStep.compareAndSet(NotRunning, Initializing)) {
       log.info("Initiating bootstrap procedure using {} method...", settings.contactPointDiscovery.discoveryMethod)
 
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -25,6 +25,7 @@ import akka.management.cluster.bootstrap.contactpoint.HttpClusterBootstrapRoutes
 import akka.management.cluster.bootstrap.internal.BootstrapCoordinator
 import akka.management.scaladsl.ManagementRouteProviderSettings
 import akka.management.scaladsl.ManagementRouteProvider
+import akka.pattern.pipe
 
 final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Extension with ManagementRouteProvider {
 
@@ -81,7 +82,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
       val bootstrapProps = BootstrapCoordinator.props(discovery, joinDecider, settings)
       val bootstrap = system.systemActorOf(bootstrapProps, "bootstrapCoordinator")
       // Bootstrap already logs in several other execution points when it can't form a cluster, and why.
-      bootstrap ! BootstrapCoordinator.Protocol.InitiateBootstrapping
+      val initiateBootstrapping = selfContactPoint.map(BootstrapCoordinator.Protocol.InitiateBootstrapping)
+      initiateBootstrapping pipeTo bootstrap
     } else log.warning("Bootstrap already initiated, yet start() method was called again. Ignoring.")
 
   /**
@@ -91,18 +93,13 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
    * This allows us to "reverse lookup" from a lowest-address sorted contact point list,
    * that we discover via discovery, if a given contact point corresponds to our remoting address,
    * and if so, we may opt to join ourselves using the address.
-   *
-   * @return true if successfully set, false otherwise (i.e. was set already)
    */
   @InternalApi
   private[akka] def setSelfContactPoint(baseUri: Uri): Unit =
     _selfContactPointUri.success(baseUri)
 
   /** INTERNAL API */
-  @InternalApi private[akka] def selfContactPoint: Future[(String, Int)] =
-    _selfContactPointUri.future.map { uri =>
-      (uri.authority.host.toString, uri.authority.port)
-    }
+  @InternalApi private[akka] def selfContactPoint: Future[Uri] = _selfContactPointUri.future
 }
 
 object ClusterBootstrap extends ExtensionId[ClusterBootstrap] with ExtensionIdProvider {
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -7,9 +7,9 @@ package akka.management.cluster.bootstrap
 import java.util.concurrent.atomic.AtomicReference
 
 import akka.AkkaVersion
-import scala.concurrent.Future
-import scala.concurrent.Promise
 
+import scala.concurrent.{Future, Promise, TimeoutException}
+import scala.concurrent.duration._
 import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
 import akka.actor.Extension
@@ -17,7 +17,7 @@ import akka.actor.ExtensionId
 import akka.actor.ExtensionIdProvider
 import akka.annotation.InternalApi
 import akka.cluster.Cluster
-import akka.discovery.{ Discovery, ServiceDiscovery }
+import akka.discovery.{Discovery, ServiceDiscovery}
 import akka.event.Logging
 import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.server.Route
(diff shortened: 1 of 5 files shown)


lightbend-cla-validator:
Hi @otto-dev,

otto-dev:
I signed the CLA. I thought I had already :man_shrugging:

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/SelfAwareJoinDecider.scala
```diff
@@ -34,10 +34,13 @@ import akka.event.Logging
    * during [[akka.management.scaladsl.AkkaManagement.start()]], hence we accept blocking on
    * this initialization.
    */
-  private[bootstrap] def selfContactPoint: (String, Int) =
-    Try(Await.result(ClusterBootstrap(system).selfContactPoint, 10.seconds)).getOrElse(throw new IllegalStateException(
+  private[bootstrap] def selfContactPoint: (String, Int) = {
```
otto-dev: I'd say a `lazy val` would be a good match here instead of re-evaluating the `Await` block, but I don't know if `lazy vals` are kinda looked down upon here. Poor lazy vals.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala
```diff
@@ -96,11 +96,8 @@ private[bootstrap] class HttpContactPointBootstrap(
 
   override def receive = {
     case ProbeTick =>
-      val req = ClusterBootstrapRequests.bootstrapSeedNodes(baseUri)
```
otto-dev: This variable never did anything, except being logged

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/HttpContactPointBootstrap.scala
```diff
@@ -96,11 +96,8 @@ private[bootstrap] class HttpContactPointBootstrap(
...
-      log.debug("Probing [{}] for seed nodes...", req.uri)
```
otto-dev: This variable never did anything, except being logged

otto-dev: Fake news

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -91,18 +93,13 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
   @InternalApi
   private[akka] def setSelfContactPoint(baseUri: Uri): Unit =
     _selfContactPointUri.success(baseUri)
 
   /** INTERNAL API */
-  @InternalApi private[akka] def selfContactPoint: Future[(String, Int)] =
-    _selfContactPointUri.future.map { uri =>
-      (uri.authority.host.toString, uri.authority.port)
-    }
+  @InternalApi private[akka] def selfContactPoint: Future[Uri] = _selfContactPointUri.future
```
otto-dev: We already have a future here that resolves to the self contact-point. The mapping function is better refactored out where the future is currently used. Then we can re-use the future for our own purposes.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -81,7 +82,8 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
       val bootstrapProps = BootstrapCoordinator.props(discovery, joinDecider, settings)
       val bootstrap = system.systemActorOf(bootstrapProps, "bootstrapCoordinator")
       // Bootstrap already logs in several other execution points when it can't form a cluster, and why.
-      bootstrap ! BootstrapCoordinator.Protocol.InitiateBootstrapping
+      val initiateBootstrapping = selfContactPoint.map(BootstrapCoordinator.Protocol.InitiateBootstrapping)
+      initiateBootstrapping pipeTo bootstrap
```
otto-dev: As soon as we know our own self contact-point, we start bootstrapping. The `SelfAwareJoinDecider` will also wait for the self-contact point future, and throw an exception + error if it is not received

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -123,6 +123,7 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
 
   private var lastContactsObservation: Option[ServiceContactsObservation] = None
   private var seedNodesObservations: Map[ResolvedTarget, SeedNodesObservation] = Map.empty
+  private var selfUri: Uri = _ // set via InitiateBootstrapping
```
otto-dev: This could be a `None`, but it is guaranteed to be defined before `become` is called

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -284,7 +285,7 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
 
   private[internal] def ensureProbing(contactPoint: ResolvedTarget): Option[ActorRef] = {
     val targetPort = contactPoint.port.getOrElse(settings.contactPoint.fallbackPort)
-    val rawBaseUri = Uri("http", Uri.Authority(Uri.Host(contactPoint.host), targetPort))
+    val rawBaseUri = Uri(selfUri.scheme, Uri.Authority(Uri.Host(contactPoint.host), targetPort))
```
otto-dev: This is what it's all about

otto-dev:
Ok I leave it alone now, let me know what you think

chbatey:
Needs a rebase now

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -42,7 +42,7 @@ private[akka] object BootstrapCoordinator {
     Props(new BootstrapCoordinator(discovery, joinDecider, settings))
 
   object Protocol {
-    final case object InitiateBootstrapping
+    final case class InitiateBootstrapping(selfUri: Uri)
```
raboof: Perhaps call it `selfContactPoint` to make it clear that this is the URI of the contact point (and not of something else)?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
raboof: Instead of adding another `var` here, could we pass this value in as an extra parameter to `bootstrapping`?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
raboof: Good to not hard-code `"http"` here.

Taking the URI of the self-contactpoint is an improvement, though I guess it might not be universal if there is additional infrastructure building blocks in between. Eventually perhaps it would be good to make this a configuration option.

Until that time this is a nice improvement, though - and as you already mentioned, it should not be a breaking change, since situations where the self-contact-point scheme is 'https' would not have worked before either.

otto-dev:
Thanks for reviewing, will get back to this once I have time.

otto-dev:
Finally time to look into this. I've rebased the code, and made the requested changes.

I think the suggestion to pass the (now renamed) `selfContactPoint` to `become(bootstrapping)` instead of having it as a class var is worth another look, in case something got missed. 

It means that the variable has to be forward passed through four methods from [here](https://github.com/akka/akka-management/blob/ba9ca61fd452ab76991473f1985568b635496c27/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L185) to [here](https://github.com/akka/akka-management/blob/ba9ca61fd452ab76991473f1985568b635496c27/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L296), which includes one partially applied function in a `foreach`.

Personally, I like neither having a `var` as it was before, nor passing the value forwards so many times. An inspiration how to solve this right now is lacking. The only improvement I could see is to have a `var` that contains the `scheme` only, and defaults to `http`. That way, at least you never have an evil `null` value.

Maybe you have a better idea, or we can leave it as it is if that was intended, either works for me.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -7,24 +7,25 @@ package akka.management.cluster.bootstrap
...
+import scala.concurrent.duration._
 import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
 import akka.actor.Extension
 import akka.actor.ExtensionId
 import akka.actor.ExtensionIdProvider
 import akka.annotation.InternalApi
 import akka.cluster.Cluster
-import akka.discovery.{ Discovery, ServiceDiscovery }
+import akka.discovery.{Discovery, ServiceDiscovery}
```
raboof: we should really #545

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -78,31 +79,43 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
     } else if (bootstrapStep.compareAndSet(NotRunning, Initializing)) {
       log.info("Initiating bootstrap procedure using {} method...", settings.contactPointDiscovery.discoveryMethod)
 
+      ensureSelfContactPoint()
       val bootstrapProps = BootstrapCoordinator.props(discovery, joinDecider, settings)
       val bootstrap = system.systemActorOf(bootstrapProps, "bootstrapCoordinator")
       // Bootstrap already logs in several other execution points when it can't form a cluster, and why.
-      bootstrap ! BootstrapCoordinator.Protocol.InitiateBootstrapping
+      val initiateBootstrapping = selfContactPoint.map(BootstrapCoordinator.Protocol.InitiateBootstrapping)
+      initiateBootstrapping pipeTo bootstrap
```
raboof: While I like how short this is, it might be somewhat more readable to write this something like
```
selfContactPoint.onSuccess {
  uri => bootstrap ! BootstrapCoordinator.Protocol.InitiateBootstrapping(uri)
}
```
Not a blocker though.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -78,31 +79,43 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
+
   /**
    * INTERNAL API
    *
    * Must be invoked by whoever starts the HTTP server with the `HttpClusterBootstrapRoutes`.
    * This allows us to "reverse lookup" from a lowest-address sorted contact point list,
    * that we discover via discovery, if a given contact point corresponds to our remoting address,
    * and if so, we may opt to join ourselves using the address.
-   *
-   * @return true if successfully set, false otherwise (i.e. was set already)
```
raboof: :+1:

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -289,12 +288,12 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
 
     // TODO stop the obsolete children (they are stopped when probing fails for too long)
 
-    newObservation.observedContactPoints.foreach(ensureProbing)
+    newObservation.observedContactPoints.foreach(ensureProbing(selfContactPoint, _))
   }
 
-  private[internal] def ensureProbing(contactPoint: ResolvedTarget): Option[ActorRef] = {
+  private[internal] def ensureProbing(selfContactPoint: Uri, contactPoint: ResolvedTarget): Option[ActorRef] = {
```
raboof: We only need the `scheme` here, not the whole `uri`

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -275,7 +274,7 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
     discovery.lookup(lookup, settings.contactPointDiscovery.resolveTimeout).pipeTo(self)
   }
 
-  private def onContactPointsResolved(contactPoints: Iterable[ResolvedTarget]): Unit = {
+  private def onContactPointsResolved(contactPoints: Iterable[ResolvedTarget], selfContactPoint: Uri): Unit = {
```
raboof: we only need the `scheme` here, not the whole `Uri`

otto-dev:
I have implemented all your recommendations @raboof

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -52,19 +55,25 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
-        }
-      }))
-      coordinator ! InitiateBootstrapping
+      val coordinator = system.actorOf(
+        Props(new BootstrapCoordinator(discovery, joinDecider, settings) {
+          override def ensureProbing(
+              selfContactPointScheme: String,
+              contactPoint: ResolvedTarget
+          ): Option[ActorRef] = {
+            println(s"Resolving $contactPoint")
```
patriknw:
```suggestion
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -88,19 +97,20 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
-            ))
+            )
+          )
       )
 
       val targets = new AtomicReference[List[ResolvedTarget]](Nil)
       val coordinator = system.actorOf(Props(new BootstrapCoordinator(discovery, joinDecider, settings) {
-        override def ensureProbing(contactPoint: ResolvedTarget): Option[ActorRef] = {
+        override def ensureProbing(selfContactPointScheme: String, contactPoint: ResolvedTarget): Option[ActorRef] = {
           println(s"Resolving $contactPoint")
```
patriknw:
```suggestion
```

PR #606, opened 2019-11-04
Upgrade akka http 10.1.10 to work with Akka 2.6

chbatey:


taeguk:
Excuse me, it means that I must use `akka management 1.0.4 above` and `akka http 10.1.10 above` for using `akka 2.6.0`. Right?

raboof:
I think it would be best to upgrade if you can. (that said, technically it should be possible to use a previous version of akka-management as long as you make sure to upgrade akka-http to 10.1.10 or higher in your project. not sure why you would want that though :) )

PR #552, opened 2019-05-22
Multiple ecs containers

sfali:
## Purpose

To allow ECS service discovery to work with multiple containers in ECS task definition.

## References

References #542

## Changes

Introduced optional `akka.discovery.aws-api-ecs.container-name` config to provide container name
if there are multiple containers in ECS task definition to allow ECS service discovery to narrow down
the container running Akka application.

## Background Context

As mentioned in issue #542 with multiple containers in ECS task definition the service discovery failed form cluster, by providing `container-name` service discovery can filter container running Akka application.

Issue #542, opened 2019-05-10
Cluster bootstrap failed to start when there are multiple container in a ECS task


sfali:
In our `ECS` (with `Fargate`) task we have two containers running, one is `Akka` application and other is a `sidecar` container. The `ECSServiceDiscovery` successfully resolves the IP addresses of both containers (which is the same IP address) but the cluster eventually failed to start since `ECSServiceDiscovery` is returning resolved targets without port:

https://github.com/akka/akka-management/blob/3ad1d3e1ee2de7c075deecefa35c6fd20a91189f/discovery-aws-api/src/main/scala/akka/discovery/awsapi/ecs/EcsServiceDiscovery.scala#L58

while `BootstrapCoordinator` filters multiple results based on port:

https://github.com/akka/akka-management/blob/3ad1d3e1ee2de7c075deecefa35c6fd20a91189f/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L185

The `Akka` version we are using is `2.5.22` while `Akka Management` version is `1.0.0`

Regards,

Syed

chbatey:
Hi @sfali I'm not very familiar with ECS, would there be a way to get the port from the API? If so could you raise a PR?

sfali:
I looked at different options on how to resolve this, one way is to call `describe-task-definition` on each container and then get a container such that whose port matches `Akka management` port, I think this would be overkill.

I think there would be one and only one container running `Akka`, I am thinking to introduce an optional  property `akka.management.cluster.bootstrap.contact-point-discovery.container-name`, for multiple containers pass the container name and service discovery will filter out based on the container name.

I'll try this and will create a PR.

Regards,

Syed

PR #610, opened 2019-11-12
Fix fallback port behavior on multiple resolved targets

milanvdm:
Related to:
* https://github.com/akka/akka-management/issues/542
* https://github.com/akka/akka-management/pull/552

The behavior/description of setting the fallback-port for cluster-bootstrap is unexpected in the case of having multiple `ResolvedTarget` with the same host but no port defined.
Currently, the fallback-port will actually filter out all the found hosts.

I changed this, in the case of not a single defined port, to use the fallback-port.

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
@@ -182,7 +182,11 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
             case (host, immutable.Seq(singleResult)) =>
               immutable.Seq(singleResult)
             case (host, multipleResults) =>
-              multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
+              if (multipleResults.exists(_.port.isDefined)) {
+                multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
+              } else {
+                multipleResults
+              }
           }
 
       log.info("Located service members based on: [{}]: [{}], filtered to [{}]", lookup, contactPoints.mkString(", "),
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -101,8 +101,9 @@ akka.management {
     contact-point {
 
       # If no port is discovered along with the host/ip of a contact point this port will be used as fallback
-      # Also, when no port-name is used and multiple results are returned for a given service, this port is
-      # used to disambiguate. When set to <fallback-port>, defaults to the value of akka.management.http.port
+      # Also, when no port-name is used and multiple results are returned for a given service with at least one
+      # port defined, this port is used to disambiguate. When set to <fallback-port>, defaults to the value of
+      # akka.management.http.port
       fallback-port = "<fallback-port>"
 
       # If some discovered seed node will keep failing to connect for specified period of time,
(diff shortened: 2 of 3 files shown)


lightbend-cla-validator:
Hi @milanvdm,

chbatey:
Thanks @milanvdm I agree this makes more sense. I think this will also be enough for https://github.com/akka/akka-management/issues/611 as it handles the case that ports aren't named, if they are named fallback port can be used

milanvdm:
@chbatey I personally would have changed it to change every `None` port to the `fallback-port`. As that would be the expected behavior IMHO.
But I didnt want to change too much logic as I am unaware of the reason why the filtering was initially added.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -182,7 +182,11 @@ private[akka] class BootstrapCoordinator(discovery: ServiceDiscovery,
             case (host, immutable.Seq(singleResult)) =>
               immutable.Seq(singleResult)
             case (host, multipleResults) =>
-              multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
+              if (multipleResults.exists(_.port.isDefined)) {
+                multipleResults.filter(_.port.contains(settings.contactPoint.fallbackPort))
+              } else {
+                multipleResults
```
chbatey: > I personally would have changed it to change every None port to the fallback-port. As that would be the expected behavior IMHO.

I think you're right. I can't think of any reason to return multiple results for the same host with no ports. The fallbackport will be added to each of them and we'll probe the same node multiple times.

First there's the easy case:

* The user has defined a port in their query so it is up to service discovery impl to return only those with that port

Then for when a user does not include a port in the query:

* Discovery returns multiple targets for the same host, without ports. We should fill in the fallback port
* Discovery returns multiple targets with different ports: right now we filter based on the fallback port as we're assuming that they are mutiple ports for the same service e.g. remoting port, management port, HTTP port. This makes bootstrap easier to use with less configuration (not having to specify the port name). But it breaks the case where dynamic ports are being used for management so all the targets should be kept. 

For the last point I can't think of a way to support both without adding another configuration flag saying that management has dynamic ports so don't filter.

WDYT @milanvdm

milanvdm: I fully agree with that approach.

Although Im not sure if your last point needs an extra configuration flag.

We get the following list:
```
host1:None
host1:1234
host2:None
host2:8558
host2:1234
```

You add the `fallback-port` to the `None`. Would you expect the `host1` not to show up in the final list of hosts?
For me, here again, whenever you have a `None`, the `fallback-port` is used. This would cause the `host1` to get into the final list of hosts.

chbatey: For your example I think there two behaviors we want to support.

* Replace the Nones with fallback ports and return only the results that either contained None or are the same as the fallback port. This is the scenario where all akka management ports are the same across your cluster. This is nearly always the case in Kubernetess for example.
* Replace the Nones with fallback ports and don't do any filtering apart from maybe duplicates.  This is the case where the akka management port is dynamic and we can't assume that results with the same host and a different port are the same Akka nodes as they could be different Akka nodes on the same host with different akka management ports.

It is the latter I think is broken atm.

milanvdm: Understood. 
If that is indeed something that needs to be supported, then you need the extra configuration.

Do you add this to this PR or is this something I need to do as an extra?

chbatey: I can follow up with another PR

chbatey: Note that we already fill in ports that are None with the fallback port: https://github.com/akka/akka-management/blob/master/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala#L286

PR #614, opened 2019-11-18
Scaladoc: make use of new flags

ennru:
## Purpose

* Upgrades to Scala 2.12.10
* Uses the new Scaladoc flags introduced in 2.12.9 and 2.13.0 for better links to source and adding canonical URLs

## References

References https://github.com/akka/akka.github.com/issues/623


chbatey:
The build failure was during the docker build for a test, likely transient, re-running.

PR #615, opened 2019-11-18
Use bintray API key rather than password

chbatey:
My mistake from last time, using pass rather than api key


chbatey:
Validation/review won't show anything here, only trying a release with it will

Issue #611, opened 2019-11-15
Bootstrap: Handle multiple akka nodes on the same host with different ports


chbatey:
Bootstrap assumes that nodes have different IPs and filters nodes with the same host unless `port-name` is set. 

`port-name` is meant to distinguish between multiple ports for the same service rather than to differentiate between different Akka nodes on the same host so we should support that explicitly.

PR #616, opened 2019-11-19
Flag for disabling filter on fallback port

chbatey:
Refs #611

### DIFF
--- a/cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
+++ b/cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
@@ -103,6 +103,7 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
   }
 
   "BootstrapCoordinator target filtering" should {
+    // For example when using DNS SRV-record-based discovery
     "not filter when port-name is set" in {
       val beforeFiltering = List(
         ResolvedTarget("host1", Some(1), None),
@@ -115,6 +116,7 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
         Lookup.create("service").withPortName("cats"), 8558, filterOnFallbackPort = true, beforeFiltering) shouldEqual beforeFiltering
     }
 
+    // For example when using DNS A-record-based discovery in K8s
     "filter when port-name is not set" in {
       val beforeFiltering = List(
         ResolvedTarget("host1", Some(8558), None),
@@ -130,11 +132,12 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
       )
     }
 
+    // For example when using ECS service discovery
     "not filter when port-name is not set but filtering disabled" in {
       val beforeFiltering = List(
-        ResolvedTarget("host1", Some(8558), None),
+        ResolvedTarget("host1", Some(1), None),
         ResolvedTarget("host1", Some(2), None),
-        ResolvedTarget("host2", Some(8558), None),
+        ResolvedTarget("host2", Some(3), None),
         ResolvedTarget("host2", Some(4), None)
       )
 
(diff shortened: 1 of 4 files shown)


cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -102,6 +102,59 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
     }
   }
 
+  "BootstrapCoordinator target filtering" should {
+    "not filter when port-name is set" in {
```
raboof: Perhaps we should add comments clarifying in which situations the different behaviors are expected? I guess:

```suggestion
    // For example when using DNS SRV-record-based discovery
    "not filter when port-name is set" in {
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -102,6 +102,59 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
+        ResolvedTarget("host2", Some(2), None),
+        ResolvedTarget("host3", Some(3), None),
+        ResolvedTarget("host4", Some(4), None)
+      )
+
+      BootstrapCoordinator.selectHosts(
+        Lookup.create("service").withPortName("cats"), 8558, filterOnFallbackPort = true, beforeFiltering) shouldEqual beforeFiltering
+    }
+
+    "filter when port-name is not set" in {
```
raboof:
```suggestion
    // For example when using DNS A-record-based discovery in K8s
    "filter when port-name is not set" in {
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -102,6 +102,59 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
+      )
+
+      BootstrapCoordinator.selectHosts(
+        Lookup.create("service"), 8558, filterOnFallbackPort = true, beforeFiltering) shouldEqual List(
+        ResolvedTarget("host1", Some(8558), None),
+        ResolvedTarget("host2", Some(8558), None)
+      )
+    }
+
+    "not filter when port-name is not set but filtering disabled" in {
```
raboof:
```suggestion
    // For example when using ECS service discovery
    "not filter when port-name is not set but filtering disabled" in {
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -102,6 +102,59 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
+      val beforeFiltering = List(
+        ResolvedTarget("host1", Some(8558), None),
```
raboof: I guess this example would be more realistic when it doesn't contain 8558 on both hosts, since in this case ports are arbitrary so would rarely happen to be 8558?

```suggestion
        ResolvedTarget("host1", Some(1), None),
```

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinatorSpec.scala
```diff
@@ -102,6 +102,59 @@ class BootstrapCoordinatorSpec extends WordSpec with Matchers with BeforeAndAfte
...
+        ResolvedTarget("host1", Some(2), None),
+        ResolvedTarget("host2", Some(8558), None),
```
raboof:
```suggestion
        ResolvedTarget("host2", Some(3), None),
```

Issue #618, opened 2019-11-19
Drop Scala 2.11 


chbatey:
Do we need a minor release for that? Seems surprising to do it in a patch release

ennru:
Fixed with #619 and released with https://github.com/akka/akka-management/releases/tag/v1.1.0

PR #619, opened 2019-11-19
Drop 2.11 support

chbatey:
This will require a 1.1 release. 

Refs #618


patriknw:
I don't mind dropping 2.11 but I don't see that we need a 1.1 release for that. It will increase our maintenance burden to keep two branches alive.

Could we drop 2.11 for upcoming 1.0.x releases without starting to use 2.12 only features? Waiting some for feedback to see if there are any important complaints?

chbatey:
Yep let’s do that

patriknw:
Given what we heard from the Play team we need to reconsider.

What was the main reason for dropping 2.11? Is that worth the alternative to maintain two branches?

chbatey:
It was to reduce overhead, having just had a build fail due to 2.11, so if we need 2 branches then that defeats the purpose :)

Issue #612, opened 2019-11-15
Dynamically set log levels via HTTP


chbatey:
Initially support log level

API should take in a log level and logger to change

Will need to use reflection or provided dependency to avoid putting logback on user classpath

ihostage:
Hi, @chbatey! 👋 
Why don't use an out-of-box [reloadable configuration feature](http://logback.qos.ch/manual/configuration.html#autoScan) of Logback?
It's difficult for many nodes?

chbatey:
Thanks for the link. This is mainly for environments where we don’t have
access to JMX or the file system e.g in Kubernetes

ihostage:
@chbatey 
Our cases:
1. We successfully use this on Kubernates. Logback configuration (with `scan` feature) stored on `ConfigMap` and mount as a volume in pods. One note that after change a ConfigMap needs to delete/create pods because K8S can't hot-change a volume. 😞 
2. We expose JMX port on K8S, but don't use it to change a log level. Thanks for idea 👍

chbatey:
A use case that comes up often is doing it without deleting pods as it is
to debug issues that will go away with a restart

ihostage:
Ohhhh... Sorry, Christopher, my mistake 😞 
I messed up with HOCON configuration, that not reloadable 😄 
Reloadable Logback configuration is all fine. When I change a ConfigMap the file in pods also changed and Logback reloads self-configuration on all pods without a problem.

chbatey:
I think it would be great to document that pattern (not akka specific) and hold off on this until more demand comes then

patriknw:
I think there is already enough demand. I have heard or wanted to recommend this many times. 

Changing the full logback configuration file is a good complement but I guess that changing files in some production deployments are not that easy.

I can probably take a look at implementing this next week.

PR #622, opened 2019-11-28
Dynamic loglevel through management routes

johanandren:
Refs #612

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ for the following modules.
 * akka-discovery-aws-api
 * akka-discovery-aws-api-async
 * akka-discovery-consul
+* loglevels-logback
 
 
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,6 @@ for the following modules.
 * akka-discovery-aws-api
 * akka-discovery-aws-api-async
 * akka-discovery-consul
-* loglevels-logback
 
 
 
--- a/docs/src/main/paradox/loglevels.md
+++ b/docs/src/main/paradox/loglevels.md
@@ -21,7 +21,7 @@ With Akka Management started and this module on the classpath the module is auto
 
 ### Reading Logger Levels
 
-A HTTP `GET` request to `loglevel?logger=[logger name]` will return the log level of that logger.
+A HTTP `GET` request to `loglevel/logback?logger=[logger name]` will return the log level of that logger.
 
 ### Changing Logger Levels
 
@@ -33,12 +33,12 @@ If enabling this make sure to properly secure your endpoint with HTTPS and authe
 
 @@@
 
-A HTTP `PUT` request to `loglevel?logger=[logger name]&level=[level name]` will change the level of that logger on the JVM the `ActorSystem` runs on.
+A HTTP `PUT` request to `loglevel/logback?logger=[logger name]&level=[level name]` will change the level of that logger on the JVM the `ActorSystem` runs on.
 
 For example using curl:
 
 ```
-curl -X PUT "http://127.0.0.1:8558/loglevel?logger=com.example.MyActor&level=DEBUG"
+curl -X PUT "http://127.0.0.1:8558/loglevel/logback?logger=com.example.MyActor&level=DEBUG"
 ```
 
 #### Classic Akka Logger Level
@@ -46,13 +46,13 @@ curl -X PUT "http://127.0.0.1:8558/loglevel?logger=com.example.MyActor&level=DEB
 When using classic Akka and logging through the built in API there is an [additional level of filtering](https://doc.akka.io/docs/akka/current/logging.html#slf4j) using the
 `akka.loglevel` setting. If you have not set `akka.loglevel` to `DEBUG` (recommended) log entries from the classic logging API may never reach the logger backend at all.
 
-The current level configured with `akka.loglevel` can be inspected with a GET request to `akka/classic/loglevel`.
+The current level configured with `akka.loglevel` can be inspected with a GET request to `loglevel/akka`.
 
-If management `read-only` is set to `false` PUT requests to `akka/classic/loglevel?level=[level name]` will dynamically change that.
+If management `read-only` is set to `false` PUT requests to `loglevel/akka?level=[level name]` will dynamically change that.
 Note that the allowed level for Akka Classic logging is a subset of the loglevels supported by SLF4j, valid values are `OFF`, `DEBUG`, `INFO`, `WARNING` and `ERROR`.
 
 For example using curl:
 
 ```
-curl -X PUT "http://127.0.0.1:8558/akka/classic/loglevel?level=DEBUG"
+curl -X PUT "http://127.0.0.1:8558/loglevel/akka?level=DEBUG"
 ```
--- a/docs/src/main/paradox/loglevels.md
+++ b/docs/src/main/paradox/loglevels.md
@@ -1,0 +1,38 @@
+# Dynamic Log Levels
+
+Dynamic Log Levels for Logback hooks into Akka Management and provides a route where log levels can be read and set over HTTP.
+
+## Project Info
+
+@@project-info{ projectId="loglevels-logback" }
+
+Requires @ref:[Akka Management](akka-management.md) and that the application uses [Logback](http://logback.qos.ch) as logging backend.
+
+@@dependency[sbt,Gradle,Maven] {
+  group=com.lightbend.akka.management
+  artifact=loglevel-logback_$scala.binary_version$
+  version=$project.version$
+  group2=com.lightbend.akka.management
+  artifact2=akka-management_$scala.binary.version$
+  version2=$project.version$
+}
+
+With Akka Management started and this module on the classpath the module is automatically picked up and provides the following two HTTP routes:
+
+### Reading Logger Levels
+
+A HTTP `GET` request to `loglevel?logger=[logger name]` will return the log level of that logger.
+
+### Changing Logger Levels
+
+Only enabled if `akka.management.http.route-providers-read-only` is set to true. 
+
+@@@ Warning
+
+If enabling this make sure to properly secure your endpoint with HTTPS and authentication or else anyone with access to the system could change logger levels and potentially do a DoS attack by setting all loggers to `TRACE`.
+
+@@@
+
+A HTTP `POST` request to `loglevel?logger=[logger name]&level=[level name]` will change the logger level of that logger.
+
+
--- a/docs/src/main/paradox/loglevels.md
+++ b/docs/src/main/paradox/loglevels.md
@@ -27,7 +27,7 @@ A HTTP `GET` request to `loglevel?logger=[logger name]` will return the log leve
 
 Only enabled if `akka.management.http.route-providers-read-only` is set to true. 
 
-@@@ Warning
+@@@ warning
 
 If enabling this make sure to properly secure your endpoint with HTTPS and authentication or else anyone with access to the system could change logger levels and potentially do a DoS attack by setting all loggers to `TRACE`.
 
--- a/docs/src/main/paradox/loglevels.md
+++ b/docs/src/main/paradox/loglevels.md
@@ -33,7 +33,7 @@ If enabling this make sure to properly secure your endpoint with HTTPS and authe
 
 @@@
 
-A HTTP `PUT` request to `loglevel?logger=[logger name]&level=[level name]` will change the level of that logger.
+A HTTP `PUT` request to `loglevel?logger=[logger name]&level=[level name]` will change the level of that logger on the JVM the `ActorSystem` runs on.
 
 For example using curl:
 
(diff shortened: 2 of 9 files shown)


loglevels-logback/src/main/scala/akka/management/loglevels/logback/LogLevelRoutes.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  private def getLogger(name: String) = {
+    val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+    context.getLogger(name)
+  }
+  override def routes(settings: ManagementRouteProviderSettings): Route =
+    path("loglevel") {
+      parameter("logger") { loggerName =>
+        concat(
+          post {
```
johanandren: Should only be enabled when not `readOnly` in settings.

ihostage:
@johanandren Do you plan to implement logic with changing log level on all cluster nodes by one HTTP request? Or I need to call all pods directly?

johanandren:
Plan was to do it on the local node, given that management will run on all nodes anyway. But I can definitely see how it would be useful if it did that for the entire cluster. I'll discuss it a bit with the team and see what they think.

README.md
```diff
@@ -34,6 +34,7 @@ for the following modules.
 * akka-discovery-aws-api
 * akka-discovery-aws-api-async
 * akka-discovery-consul
+* loglevels-logback
```
johanandren: Not sure if we want to mark it as supported right away rather?

loglevels-logback/src/main/scala/akka/management/loglevels/logback/LogLevelRoutes.scala
```diff
@@ -0,0 +1,70 @@
...
+import ch.qos.logback.classic.LoggerContext
+import org.slf4j.LoggerFactory
+
+object LogLevelRoutes extends ExtensionId[LogLevelRoutes] {
+  override def createExtension(system: ExtendedActorSystem): LogLevelRoutes =
+    new LogLevelRoutes
+
+  private implicit val levelFromStringUnmarshaller: Unmarshaller[String, Level] =
+    Unmarshaller.strict { string =>
+      Level.valueOf(string)
```
chbatey: maybe validate that this is valid? Otherwise it will silently set it to DEBUG if someone typos ERROR which may blow up their logging infra

johanandren: Ah you are right, I expected it to throw for invalid value for some reason.

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,38 @@
...
+
+With Akka Management started and this module on the classpath the module is automatically picked up and provides the following two HTTP routes:
+
+### Reading Logger Levels
+
+A HTTP `GET` request to `loglevel?logger=[logger name]` will return the log level of that logger.
+
+### Changing Logger Levels
+
+Only enabled if `akka.management.http.route-providers-read-only` is set to true.
```
patriknw: read-only set to false?

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,38 @@
...
+
+Only enabled if `akka.management.http.route-providers-read-only` is set to true. 
+
+@@@ warning
+
+If enabling this make sure to properly secure your endpoint with HTTPS and authentication or else anyone with access to the system could change logger levels and potentially do a DoS attack by setting all loggers to `TRACE`.
+
+@@@
+
+A HTTP `POST` request to `loglevel?logger=[logger name]&level=[level name]` will change the logger level of that logger.
```
patriknw: Isn't POST with request params an anti-pattern?
Since it's a modification PUT (or POST) with json or form data might would be more RESTish.

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,38 @@
...
+
+
```
patriknw: would also be good to have read and update of `akka.loglevel`. It's recommended to set that to DEBUG when using logback, but since our default is INFO I guess many leave it at INFO. It can be changed with `ActorSystem.eventStream.setLogLevel`

project/project-info.conf
```diff
@@ -102,5 +102,14 @@ project-info {
       }
     ]
   }
-
+  loglevels-logback: ${project-info.shared-info} {
+    title: "Dynamic Log Levels"
+    levels: [
+      { // keep in sync with README.md
+        readiness: CommunityDriven
```
patriknw: I think we can take responsibility for maintaining this

README.md
patriknw: yeah, not much to incubate

project/Dependencies.scala
```diff
@@ -138,6 +138,14 @@ object Dependencies {
       )
   )
 
+  val LoglevelsLogback = Seq(
+    libraryDependencies ++=
+      DependencyGroups.Logging ++
```
patriknw: there is a comment "// For demos/tests" that can be removed now

docs/src/main/paradox/loglevels.md
johanandren: I don't that matters much, but PUT would perhaps be more appropriate since this is an update operation rather than creating something and the operation is idempotent.

I think making this more rest style just makes things complicated both to implement and use. It would mean we should represent each logger with a path and have to make `.` in the logger names into `/` in the path etc. and then accept the logger state as body of the request.

docs/src/main/paradox/loglevels.md
johanandren: In what circumstance should that be read and updated, we can't know if a given logger is used by a classic actor, but perhaps we could always change it if the loglevel is changed to DEBUG? Could be a bit surprising, no?

docs/src/main/paradox/loglevels.md
johanandren: Ups

docs/src/main/paradox/loglevels.md
patriknw: I wasn't thinking of making each logger a resource. It could be pragmatic rest with a single loglevel resource that can be changed, but if I'm downvoted on this I don't have a strong opinion about it.

docs/src/main/paradox/loglevels.md
patriknw: I was thinking of having a separate route path for changing the global `akka.loglevel`. Useful in case where something is running "accidentally" in production with INFO level and one want to change to DEBUG to see more about an ongoing incident.

johanandren: Ah, ok that sounds more sensible indeed. I'll add that.

docs/src/main/paradox/loglevels.md
johanandren: Changed to PUT now so I met you halfways ;)

docs/src/main/paradox/loglevels.md
```diff
@@ -40,3 +40,19 @@ For example using curl:
 ```
 curl -X PUT "http://127.0.0.1:8558/loglevel?logger=com.example.MyActor&level=DEBUG"
 ```
+
+#### Classic Akka Logger Level
+
+When using classic Akka and logging through the built in API there is an [additional level of filtering](https://doc.akka.io/docs/akka/current/logging.html#slf4j) using the
+`akka.loglevel` setting. If you have not set `akka.loglevel` to `DEBUG` (recommended) log entries from the classic logging API may never reach the logger backend att all.
```
patriknw: typo: att

docs/src/main/paradox/loglevels.md
```diff
@@ -40,3 +40,19 @@ For example using curl:
...
+
+The current level configured with `akka.loglevel` can be inspected with a GET request to `akka/classic/loglevel`.
+
+If management `read-only` is set to `false` PUT requests to `akka/classic/loglevel?level=[level name]` will dynamically change that.
```
patriknw: I'm not fully convinced that we should call this Classic, since it's also used for internal logging (also when using Typed): https://doc.akka.io/docs/akka/current/typed/logging.html#internal-logging-by-akka

loglevels-logback/src/test/scala/akka/management/loglevels/logback/LogLevelRoutesSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+      val readOnlyRoutes = LogLevelRoutes
+        .createExtension(system.asInstanceOf[ExtendedActorSystem])
+        .routes(ManagementRouteProviderSettings(Uri("https://example.com"), readOnly = true))
+      Put("/loglevel?logger=LogLevelRoutesSpec&level=DEBUG") ~> readOnlyRoutes ~> check {
+        response.status should ===(StatusCodes.Forbidden)
+      }
+    }
+
+    "allow inspecting classic Akka loglevel" in {
+      Get("/akka/classic/loglevel") ~> routes ~> check {
```
patriknw: if the ordinary (logback) is `/loglevel` this could be `/loglevel/akka` ?

johanandren: I thought it was good to put all things Akka under `akka` (did so with the enhancement ones as well) since any number of route providers could add their own routes. The other one `loglevel` is JVM-global though and not strictly related to Akka so that's why I gave that a toplevel path.

Not that I feel very strongly about it...

docs/src/main/paradox/loglevels.md
johanandren: Ah, good point. So perhaps just skip classic from that path then?

loglevels-logback/src/test/scala/akka/management/loglevels/logback/LogLevelRoutesSpec.scala
patriknw: true that is good to use somewhat unique paths. I was thinking that it was a sub-feature of `/loglevel`

The existing cluster management is under `/cluster/members/` so we wouldn't place everything Akka under /akka anyway.

johanandren: Good point, I had forgotten we already had some stuff in here, I think it should be leaf-paths though, so wdyt about `/loglevel/logback` and `/loglevel/akka` in that case?

johanandren: Pushed a change of paths to that last suggestion

patriknw: good, thanks

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,58 @@
...
+  group2=com.lightbend.akka.management
+  artifact2=akka-management_$scala.binary.version$
+  version2=$project.version$
+}
+
+With Akka Management started and this module on the classpath the module is automatically picked up and provides the following two HTTP routes:
+
+### Reading Logger Levels
+
+A HTTP `GET` request to `loglevel?logger=[logger name]` will return the log level of that logger.
```
patriknw: path changed?

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,58 @@
...
+
+Only enabled if `akka.management.http.route-providers-read-only` is set to `false`. 
+
+@@@ warning
+
+If enabling this make sure to properly secure your endpoint with HTTPS and authentication or else anyone with access to the system could change logger levels and potentially do a DoS attack by setting all loggers to `TRACE`.
+
+@@@
+
+A HTTP `PUT` request to `loglevel?logger=[logger name]&level=[level name]` will change the level of that logger on the JVM the `ActorSystem` runs on.
```
patriknw: path

docs/src/main/paradox/loglevels.md
```diff
@@ -0,0 +1,58 @@
...
+```
+curl -X PUT "http://127.0.0.1:8558/loglevel?logger=com.example.MyActor&level=DEBUG"
+```
+
+#### Classic Akka Logger Level
+
+When using classic Akka and logging through the built in API there is an [additional level of filtering](https://doc.akka.io/docs/akka/current/logging.html#slf4j) using the
+`akka.loglevel` setting. If you have not set `akka.loglevel` to `DEBUG` (recommended) log entries from the classic logging API may never reach the logger backend at all.
+
+The current level configured with `akka.loglevel` can be inspected with a GET request to `akka/classic/loglevel`.
```
patriknw: path

docs/src/main/paradox/loglevels.md
patriknw: update path

Issue #629, opened 2019-12-17
Structured log events for bootstrap process


patriknw:
Make use of https://github.com/akka/akka/issues/28207 from Akka Management and Bootstrap in particular.

PR #630, opened 2019-12-17
Structured log events for bootstrap, requires Akka 2.5.27, #629

patriknw:
Refs #629

### DIFF
--- a/integration-test/kubernetes-api-java/pom.xml
+++ b/integration-test/kubernetes-api-java/pom.xml
@@ -18,9 +18,9 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <encoding>UTF-8</encoding>
-    <akka.version>2.5.19</akka.version>
+    <akka.version>2.5.27</akka.version>
     <akka.http.version>10.1.6</akka.http.version>
-    <akka-management.version>0.20.0</akka-management.version>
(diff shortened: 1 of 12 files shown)


integration-test/kubernetes-api-java/pom.xml
```diff
@@ -18,9 +18,9 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <encoding>UTF-8</encoding>
-    <akka.version>2.5.19</akka.version>
+    <akka.version>2.5.27</akka.version>
```
patriknw: travis failed with "Current version of Akka is [2.5.19], but cluster-bootstrap requires version [2.5.27]", so updated this. Not sure if there are more places but didn't find anything more when searching.

integration-test/kubernetes-api-java/pom.xml
```diff
@@ -18,9 +18,9 @@
...
     <akka.http.version>10.1.6</akka.http.version>
-    <akka-management.version>0.20.0</akka-management.version>
+    <akka-management.version>1.0.5</akka-management.version>
```
patriknw: should this be updated? 0.20.0 seems very old?

raboof: yes, ideally we should pass this in via a property instead of having it as a constant in the pom.xml

patriknw:
rebased

PR #632, opened 2019-12-17
update scalafmt, same as in akka/akka

patriknw:
* enforce formatting by checking in travis

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,13 +1,51 @@
-version = 2.0.0-RC5
+version = 2.3.0
 
 style = defaultWithAlign
 
-align.tokens               = [off]
-danglingParentheses        = true
 docstrings                 = JavaDoc
 indentOperator             = spray
 maxColumn                  = 120
-rewrite.rules              = [RedundantParens, SortImports]
-spaces.inImportCurlyBraces = true
+rewrite.rules              = [RedundantParens, SortImports, AvoidInfix]
 unindentTopLevelOperators  = true
-//binPack.callSite           = true
+align.tokens               = [{code = "=>", owner = "Case"}]
+align.openParenDefnSite    = false
+align.openParenCallSite    = false
+optIn.breakChainOnFirstMethodDot = false
+danglingParentheses = false
+spaces.inImportCurlyBraces = true
+rewrite.neverInfix.excludeFilters = [
+  and
+  min
+  max
+  until
+  to
+  by
+  eq
+  ne
+  "should.*"
+  "contain.*"
+  "must.*"
+  in
+  ignore
+  be
+  taggedAs
+  thrownBy
+  synchronized
+  have
+  when
+  size
+  only
+  noneOf
+  oneElementOf
+  noElementsOf
+  atLeastOneElementOf
+  atMostOneElementOf
+  allElementsOf
+  inOrderElementsOf
+  theSameElementsAs
+]
+rewriteTokens = {
+  "⇒": "=>"
+  "→": "->"
+  "←": "<-"
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,7 @@ jobs:
     script:
     - jabba use "adopt@1.8-0"
     - git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH"
-    - sbt -jvm-opts .jvmopts-travis ";+test:compile;docs/paradox"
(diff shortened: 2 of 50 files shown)


project/plugins.sbt
```diff
@@ -1,7 +1,7 @@
 addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.3.1")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.2.0")
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.0.0-RC5")
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.0.6")
```
raboof:
```suggestion
addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.0")
```
on this version I didn't see 'search state exploded'

.scalafmt.conf
```diff
@@ -1,13 +1,52 @@
...
-rewrite.rules              = [RedundantParens, SortImports]
-spaces.inImportCurlyBraces = true
+rewrite.rules              = [RedundantParens, SortImports, AvoidInfix]
 unindentTopLevelOperators  = true
-//binPack.callSite           = true
+align.tokens               = [{code = "=>", owner = "Case"}]
+align.openParenDefnSite    = false
+align.openParenCallSite    = false
+optIn.breakChainOnFirstMethodDot = false
+optIn.configStyleArguments = false
```
raboof: TBQH I'm not sure if I like this one - this is what forces
```
      val checks = im.Seq(
        SlowCheck,
        OkCheck
      )
```
to be rewritten as
```
      val checks = im.Seq(SlowCheck, OkCheck)
```
While setting this to 'true' seems to allow either (does not touch it)

.travis.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     script:
     - jabba use "adopt@1.8-0"
     - git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH"
-    - sbt -jvm-opts .jvmopts-travis ";+test:compile;docs/paradox"
+    - sbt -jvm-opts .jvmopts-travis ";scalafmtCheckAll;+test:compile;docs/paradox"
     - git diff --exit-code --color || { echo "[error] Unformatted code found. Please run 'test:compile' and commit the reformatted code."; false; }
```
raboof: This line is now no longer necessary
```suggestion
```

patriknw: removed (would be needed if we add javafmt)

.scalafmt.conf
patriknw: I agree, it seems to depend on where the `)` is placed

project/plugins.sbt
patriknw: good, updated

.scalafmt.conf
raboof: you could
```
git reset HEAD^
git add .scalafmt.conf project/plugins.sbt
git restore .
sbt scalafmtAll
git commit -a
```
for a smaller diff now ;)

patriknw: I think I did something like that to reformat all with the new version/settings. Do you see anything unexpected?

raboof: I would not expect https://github.com/akka/akka-management/pull/632/files#diff-b7dade6df5a2268bddde162140416a7fR161 anymore

patriknw: ah, I tried to change those manually. If placing the trailing `)` on a new line (as it was) then it keeps them as it. I was thinking that it looked better with one-liners for these, and misunderstood that you had the opposite opinion. I'll revert those manual changes.

raboof: Ahh gotcha. I don't have a strong opinion in these particular cases, but in general I like scalafmt not to touch those when I do :smile:

.travis.yml
```diff
@@ -36,8 +36,7 @@ jobs:
     script:
     - jabba use "adopt@1.8-0"
     - git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH"
-    - sbt -jvm-opts .jvmopts-travis ";+test:compile;docs/paradox"
-    - git diff --exit-code --color || { echo "[error] Unformatted code found. Please run 'test:compile' and commit the reformatted code."; false; }
+    - sbt -jvm-opts .jvmopts-travis ";scalafmtCheckAll;+test:compile;docs/paradox"
```
ignasi35: Consider `+scalafmtCheckAll scalafmtSbtCheck` in case you have code dependant on the scala version and/or sbt code.

PR #634, opened 2019-12-18
AWS: Add tag-based filtering to ecs-service-discovery-async

milanvdm:
## Purpose

Extends the `aws-ecs-async` discovery method to also be able to filter nodes based on tags.

## Changes

* Adds new `tags` configuration key 
* Filter based on `tags` configuration

## Background Context

Be able to only form a cluster with specific tasks using certain tags


discovery-aws-api-async/src/main/resources/reference.conf
```diff
@@ -12,5 +12,7 @@ akka.discovery {
 
     cluster = "default"
 
+    tags = []
```
chbatey: Add a comment for the format?

patriknw:
There is a merge conflict here now

milanvdm:
Ill fix it when Im back from holidays

PR #636, opened 2019-12-30
Use TaskSets to discover nodes in ECS

milanvdm:
## Purpose

Add a new discovery method for Akka in ECS Fargate using TaskSets.

## Changes

1. Add `aws-api-ecs-task-set-async`
2. Update aws docs
3. Tested on our own instances using AWS CodeDeploy

## Background Context

ECS can be used together with AWS CodeDeploy to support blue/green deployments.
These new nodes are spawned in the same service-group. Therefore, we need a new approach to make sure nodes only form a cluster with only newly spawned nodes.
Otherwise, the new nodes will join the existing AkkaCluster forming 1 large cluster instead of 2 smaller ones.

### DIFF
--- a/discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
+++ b/discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
@@ -4,7 +4,7 @@
 
 package akka.discovery.awsapi.ecs
 
-import java.net.{ InetAddress, NetworkInterface }
+import java.net.InetAddress
 import java.util.concurrent.TimeoutException
 
 import scala.collection.JavaConverters._
@@ -59,7 +59,7 @@ class AsyncEcsTaskSetDiscovery(system: ActorSystem) extends ServiceDiscovery {
     Future.firstCompletedOf(
       Seq(
         after(resolveTimeout, using = system.scheduler)(
-          Future.failed(new TimeoutException("Future timed out!"))
+          Future.failed(new TimeoutException(s"$lookup timed out after $resolveTimeout"))
         ),
         resolveTasks(ecsClient, cluster, httpClient).map(
           tasks =>
@@ -89,27 +89,6 @@ object AsyncEcsTaskSetDiscovery {
   private[this] implicit val orderFormat = jsonFormat1(TaskMetadata)
 
   private val ECS_CONTAINER_METADATA_URI_PATH = "ECS_CONTAINER_METADATA_URI"
-
-  // InetAddress.getLocalHost.getHostAddress throws an exception when running
-  // in awsvpc mode because the container name cannot be resolved.
-  // ECS provides a metadata file
-  // (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-metadata.html)
-  // that we ought to be able to use instead to find our IP address, but the
-  // metadata file does not get set when running on Fargate. So this is our
-  // only option for determining what the canonical Akka and akka-management
-  // hostname values should be set to.
-  def getContainerAddress: Either[String, InetAddress] =
-    NetworkInterface.getNetworkInterfaces.asScala
-      .flatMap(_.getInetAddresses.asScala)
-      .filterNot(_.isLoopbackAddress)
-      .filter(_.isSiteLocalAddress)
-      .toList match {
-      case List(value) =>
-        Right(value)
-
-      case other =>
-        Left(s"Exactly one private address must be configured (found: $other).")
-    }
 
   private def resolveTasks(ecsClient: EcsAsyncClient, cluster: String, httpClient: HttpExt)(
       implicit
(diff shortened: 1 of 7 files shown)


discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+  private def resolveTasks(ecsClient: EcsAsyncClient, cluster: String, httpClient: HttpExt)(
+      implicit
+      ec: ExecutionContext,
+      mat: ActorMaterializer
+  ): Future[Seq[Task]] =
+    for {
+      taskArn <- resolveTaskMetadata(httpClient).map(_.map(_.TaskARN))
+      taskSet <- taskArn match {
```
milanvdm: Didn't want to pull in cats to have `OptionT`, so I came up with this. Suggestions are welcome to make this cleaner.

milanvdm:
@chbatey @sebastianharko Could someone have a look at this PR?

chbatey:
Hi @milanvdm i'll look at this now, sorry for the delay.

Is this typically how you want to do blue/green deployments with Akka Cluster, particularly when using Cluster Sharding and/or Singletons? If you have two clusters it'll be up to external load balancing to make sure messages for any entity only go to one cluster.

We've been discussing in the team how better to support canary and blue/green deployments within a single Cluster by supporting a concept of application version and making sure that once entities are started on a new version they aren't then rebalanced to an old version.

As this is a new method I think we can merge it even if users disagree on how they want to do blue green deployments.

milanvdm:
@chbatey As you mentioned, AkkaClustering/Sharding does not support blue-green deployments atm.
So you will have to rely on some mechanism to send traffic to the correct cluster out of both sides.

There are of course quite some edge-cases when relying on an external system. For example, what if one side processes a message after the sides have switched (2 clusters trying to write to the same entity).

For now, I don't see a better way (at least when using ECS Fargate).

Does this answer your question?

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
```diff
@@ -0,0 +1,217 @@
...
+  DescribeTasksRequest,
+  DesiredStatus,
+  ListTasksRequest,
+  Task,
+  TaskField,
+  Tag => _
+}
+import spray.json.DefaultJsonProtocol._
+
+@ApiMayChange
```
chbatey: 👍

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
```diff
@@ -0,0 +1,217 @@
...
+class AsyncEcsTaskSetDiscovery(system: ActorSystem) extends ServiceDiscovery {
+
+  private[this] val config = system.settings.config.getConfig("akka.discovery.aws-api-ecs-task-set-async")
```
chbatey: just private?

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
```diff
@@ -0,0 +1,217 @@
...
+  private[this] implicit val materializer: ActorMaterializer = ActorMaterializer()
+  private[this] implicit val ec: ExecutionContext = system.dispatcher
+
+  private[this] val httpClient: HttpExt = Http()
+
+  override def lookup(lookup: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] =
+    Future.firstCompletedOf(
+      Seq(
+        after(resolveTimeout, using = system.scheduler)(
+          Future.failed(new TimeoutException("Future timed out!"))
```
chbatey: Useful message? s"$lookup timed out after x"?

discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsTaskSetDiscovery.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  // InetAddress.getLocalHost.getHostAddress throws an exception when running
+  // in awsvpc mode because the container name cannot be resolved.
+  // ECS provides a metadata file
+  // (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-metadata.html)
+  // that we ought to be able to use instead to find our IP address, but the
+  // metadata file does not get set when running on Fargate. So this is our
+  // only option for determining what the canonical Akka and akka-management
+  // hostname values should be set to.
+  def getContainerAddress: Either[String, InetAddress] =
```
chbatey: Share with the other async ecs impl?

docs/src/main/paradox/discovery/aws.md
```diff
@@ -139,10 +139,8 @@ Screenshot of two ECS task instances (the service name is
 
 #### Dependencies and usage (ECS Discovery)
 
-There are two "flavours" of the ECS Discovery module. Functionally they are
-identical; the difference is in which version of the AWS SDK they use. They are
-both provided so that you can choose which set of AWS SDK dependencies you're
-most comfortable with bringing in to your project.
+There are two "flavours" of the ECS Discovery module. One offers service-discovery using
```
chbatey: three now?

milanvdm: In this context, there are still only 2 since it refers to blocking vs non-blocking AWS SDK usage.

milanvdm:
@chbatey Could you have another look?

PR #638, opened 2020-02-05
Include failing/timed out check name in health/readiness check response

longshorej:
Include failing check name in health/readiness check messages/responses

This modifies the readiness/health check internals to ensure that when a check fails, for any reason, that the check's class name is included in the message to aid in debugging. Additionally, timeouts are now tracked on a per-check basis so that it's clear which check timed out.

~I haven't updated the tests yet -- if I could get some validation/assurance that such a chance would be accepted, then I'd be happy to get the tests passing.~

*Ready for a review.*

Example failure:

```
$ curl  192.168.1.20:8558/ready

Health Check Failed: Check [com.myco.GrpcReadinessCheck] failed: UNAVAILABLE: io exception
```

### DIFF
--- a/management/src/main/scala/akka/management/javadsl/HealthChecks.scala
+++ b/management/src/main/scala/akka/management/javadsl/HealthChecks.scala
@@ -7,14 +7,13 @@ package akka.management.javadsl
 import java.util.concurrent.CompletionStage
 import java.util.function.Supplier
 import java.util.function.{ Function => JFunction }
-import java.util.{ List => JList }
+import java.util.{ Optional, List => JList }
 
+import scala.compat.java8.FunctionConverters._
 import scala.compat.java8.FutureConverters._
-
 import akka.actor.ActorSystem
 import akka.actor.ExtendedActorSystem
 import akka.actor.setup.Setup
-import akka.dispatch.ExecutionContexts
 import akka.management.HealthCheckSettings
 import akka.management.internal.HealthChecksImpl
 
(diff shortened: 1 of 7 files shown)


renatocaval:
I like the idea and I don't see a reason for not improving the message.

longshorej:
Cool, thanks @renatocaval. This is ready for a review.

management/src/main/scala/akka/management/scaladsl/HealthChecks.scala
```diff
@@ -33,14 +33,14 @@ abstract class HealthChecks {
   /**
    * Returns Future(true) if the system is ready to receive user traffic
    */
-  def ready(): Future[Boolean]
+  def ready(): Future[Either[String, Unit]]
```
chbatey: This changes the public API in a non-compatible way. 

We'd have to add new methods and delegate

management/src/main/scala/akka/management/javadsl/HealthChecks.scala
```diff
@@ -30,16 +29,16 @@ final class HealthChecks(system: ExtendedActorSystem, settings: HealthCheckSetti
   /**
    * Returns CompletionStage(true) if the system is ready to receive user traffic
    */
-  def ready(): CompletionStage[java.lang.Boolean] =
-    delegate.ready().map(Boolean.box)(ExecutionContexts.sameThreadExecutionContext).toJava
+  def ready(): CompletionStage[Either[String, Unit]] =
```
patriknw: we shouldn't use `Either` in javadsl

longshorej:
> Can we solve this with additional or improved logging instead of changing API?

I'll look into it, but atm I'm not sure how unfortunately, as I think the majority of the utility is from having the messages directly in the HTTP response. Consider it may be a bit more difficult to obtain host logs than it is to hit an endpoint.

management/src/main/scala/akka/management/scaladsl/HealthChecks.scala
longshorej: Okay, that should be simple enough, I'll update when I have some time.

management/src/main/scala/akka/management/javadsl/HealthChecks.scala
longshorej: Good point - ideas on what a good alternative would be for Java (if this PR goes forward)? Maybe `Optional<String>` (plus delegated methods requested by @chbatey)?

chbatey: Probably a custom type is more idiomatic in Java

chbatey:
I agree it will be very useful to have this in the error message so you can see it from the place doing the pinging of the health checks. If we can do it in an API compatible way I am +1 to doing this

longshorej:
Okay, I've pushed an update to ensure the API remains compatible, let me know what y'all think. Thanks!

management/src/main/scala/akka/management/javadsl/HealthChecks.scala
```diff
@@ -27,19 +26,31 @@ final class HealthChecks(system: ExtendedActorSystem, settings: HealthCheckSetti
...
+   */
+  def readyResult(): CompletionStage[CheckResult] =
+    delegate.readyResult().map(new CheckResult(_))(system.dispatcher).toJava
+
   /**
    * Returns CompletionStage(true) if the system is ready to receive user traffic
    */
   def ready(): CompletionStage[java.lang.Boolean] =
-    delegate.ready().map(Boolean.box)(ExecutionContexts.sameThreadExecutionContext).toJava
+    readyResult().thenApply(((r: CheckResult) => r.isSuccess).asJava)
```
longshorej: Originally had `.thenApply(_.iSuccess).asJava` but that failed to compile on 2.11:

```
[error] /Users/jlongshore/work/tubi/akka-management/management/src/main/scala/akka/management/javadsl/HealthChecks.scala:38:29: missing parameter type for expanded function ((x$2) => x$2.isSuccess)
[error]     readyResult().thenApply(_.isSuccess)
[error]                             ^
[error] /Users/jlongshore/work/tubi/akka-management/management/src/main/scala/akka/management/javadsl/HealthChecks.scala:52:29: missing parameter type for expanded function ((x$4) => x$4.isSuccess)
[error]     aliveResult().thenApply(_.isSuccess)
[error]                             ^
[error] two errors found
```

longshorej:
Hi @patriknw and @chbatey - do you think you'll have some time to look this over in the coming weeks?

Issue #645, opened 2020-03-26
Log level check for INFO excludes WARNING logs in LowestAddressJoinDecider


dope9967:
### Versions used 

Akka version: 2.5.13

### Expected Behavior

https://github.com/akka/akka-management/blob/v0.17.0/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala#L59-L74
Expecting warning log to output if logger level is WARN.

### Actual Behavior

Warning log is wrapped in INFO level condition, this skipping output.

### Suggested Fix

Could produce a MR that removes check on line https://github.com/akka/akka-management/blob/v0.17.0/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala#L59.  Could be fixed both in latest master and backported as v0.17.1.

raboof:
> Could produce a MR that removes check

A MR would be appreciated! I think since collecting the information to log could be heavy it might make sense to have isInfoEnabled/isWarningEnabled there still.

PR #647, opened 2020-03-27
Fix LowestAddressJoinDecider cluster formation warning log to not depend on INFO logging level being set

dope9967:
Issue is quite self explanatory - I can't see a good reason to have an important warning log to be disabled by not setting logging level to INFO.

References https://github.com/akka/akka-management/issues/645

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
@@ -63,8 +63,8 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
         if (isJoinSelfAble && settings.newClusterEnabled)
           JoinSelf.asCompletedFuture
         else {
-          if (log.isInfoEnabled) {
-            if (settings.newClusterEnabled)
+          if (settings.newClusterEnabled) {
+            if (log.isInfoEnabled)
               log.info(
                 BootstrapLogMarker.inProgress(info.contactPoints.map(contactPointString), info.allSeedNodes),
                 "Exceeded stable margins without locating seed-nodes, however this node {} is NOT the lowest address " +
@@ -74,7 +74,8 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
                 lowestAddress.map(contactPointString).getOrElse(""),
                 info.contactPoints.map(contactPointString).mkString(", ")
               )
-            else
+          } else {
+            if (log.isWarningEnabled)
               log.warning(
                 BootstrapLogMarker.inProgress(info.contactPoints.map(contactPointString), info.allSeedNodes),
                 "Exceeded stable margins without locating seed-nodes, however this node {} is configured with " +


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -74,7 +74,7 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
                 lowestAddress.map(contactPointString).getOrElse(""),
                 info.contactPoints.map(contactPointString).mkString(", ")
               )
-            else
+            else if (log.isWarningEnabled)
```
renatocaval: In the previous code, this logging was only displaying if `newClusterEnabled` was false. That's not the case anymore.

dope9967: Yeah, `scalafmt` screwed me here. I'll just add curly braces.

renatocaval: See logging below:

```
Exceeded stable margins without locating seed-nodes, however this node {} is configured with 
new-cluster-enabled=off, thus NOT joining self. Expecting existing cluster or node [{}] 
(out of [{}]) to perform the self-join and initiate the cluster.
```

The full log says, that `new-cluster-enabled=off`.

dope9967: Fixed it up

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
```diff
@@ -74,7 +74,8 @@ class LowestAddressJoinDecider(system: ActorSystem, settings: ClusterBootstrapSe
                 lowestAddress.map(contactPointString).getOrElse(""),
                 info.contactPoints.map(contactPointString).mkString(", ")
               )
-            else
+          } else {
+            if (log.isWarningEnabled)
```
renatocaval: Strictly speaking, checking if warning logging is enabled is not that useful. Warn logging is often enable so the chances of optimisation are minimal here. That's much different that `info` and `debug`. Those are the levels that are in practice tweaked by users.

dope9967: Makes sense.

renatocaval: Anyway, not a big issue and I think it's ok to leave like that. That was just a side commend.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/LowestAddressJoinDecider.scala
renatocaval: Thanks for the quick reaction. 👍

chbatey:
The build failure is because we test against the latest version of k8s with minikube and the install for that no longer works. I'll fix that under a separate issue.

chbatey:
The failure is unrelated and fixed on master

PR #648, opened 2020-04-01
sbt 1.3.9; Akka Paradox 32

ennru:
Was sbt 1.2.8


ennru:
The build fails with "The none driver requires conntrack to be installed for kubernetes version 1.18.0"?

https://travis-ci.org/github/akka/akka-management/jobs/669566360#L279

raboof:
> The none driver requires conntrack to be installed for kubernetes version 1.18.0

I think travis is using 'the latest minikube' and that changed behaviour, will PR the fix.

raboof:
#649

Issue #651, opened 2020-04-06
Master Whitesource fails on jiffi


ennru:
https://travis-ci.org/github/akka/akka-management/jobs/669618428#L336

Native and non-native.

```
ModuleInfo(com.github.jnr,jffi,1.2.16,Some((Artifact(jffi, jar, jar, None, Vector(), Some(https://repo1.maven.org/maven2/com/github/jnr/jffi/1.2.16/jffi-1.2.16.jar), Map(), None, false),/home/travis/.cache/coursier/v1/https/repo1.maven.org/maven2/com/github/jnr/jffi/1.2.16/jffi-1.2.16.jar)))

ModuleInfo(com.github.jnr,jffi,1.2.16,Some((Artifact(jffi, jar, jar, Some(native), Vector(), Some(https://repo1.maven.org/maven2/com/github/jnr/jffi/1.2.16/jffi-1.2.16-native.jar), Map(), None, false),/home/travis/.cache/coursier/v1/https/repo1.maven.org/maven2/com/github/jnr/jffi/1.2.16/jffi-1.2.16-native.jar)))


```

ennru:
Most likely related to https://github.com/sbt/sbt/issues/5325

ennru:
This showed up as #648 switcht to sbt 1.3.9 with Coursier.

PR #652, opened 2020-04-06
Build: use ivy to circumvent classifier problem

ennru:
References #651


ennru:
Let's see if #655 solves it better.

ennru:
Fixed by #655

Issue #662, opened 2020-05-05
akka-management working with akka 2.6?


olijaun:
Hi

Is the current akka-management version supposed to work with Akka 2.6? Where can I find the compatibility? On https://doc.akka.io/docs/akka-management/current/akka-management.html# it is mentioning scala Versions but not Akka versions. I could not find this information anywhere. I'm sorry if this is a stupid question :-)

Regards
Oliver

patriknw:
It's compatible with Akka 2.6. Sorry that we haven't clarified that in the documentation. You may have to explicitly define more Akka dependencies in your build so that all Akka dependencies are on the same patch version, 2.6.5. Otherwise you will see a runtime exception about that.

patriknw:
Clarified in PR #662

olijaun:
great! thanks a lot!

ennru:
Fixed with #663

PR #663, opened 2020-05-05
Docs: Clarify that it can be used with Akka 2.6

patriknw:
References #662

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -258,6 +258,7 @@ lazy val docs = project
         "canonical.base_url" -> "https://doc.akka.io/docs/akka-management/current",
         "scala.binary_version" -> scalaBinaryVersion.value,
         "akka.version" -> Dependencies.AkkaVersion,
+        "akka.version26" -> "2.6.5",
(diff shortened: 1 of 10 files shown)


patriknw:
I had missed scala.binary_version and also adjusted to use `symbol=AkkaVersion`

build.sbt
```diff
@@ -258,6 +258,7 @@ lazy val docs = project
         "canonical.base_url" -> "https://doc.akka.io/docs/akka-management/current",
         "scala.binary_version" -> scalaBinaryVersion.value,
```
ennru, line 259: This needs to become `scala.binary.version`

patriknw: Strange, it actually works with `scala.binary.version` in the dependency snippets without changing this. Anyway, I changed all usages of `scala.binary.version` to `"scala.binary_version"` since that was what used previously.

ennru: Hm, I need to check why that worked...

ennru: Ah, it is the new feature in the Paradox that requires that... I'll fix once this is merged.

PR #666, opened 2020-05-08
Kubernetes: Opt in to ip instead of host

sebastian-alfers:
Enables the usage of the dotted IP address string as the host name, rather than relying on Kube's pod DNS scheme.

### DIFF
--- a/discovery-kubernetes-api/src/main/resources/reference.conf
+++ b/discovery-kubernetes-api/src/main/resources/reference.conf
@@ -31,5 +31,8 @@ akka.discovery {
     # Selector value to query pod API with.
     # `%s` will be replaced with the configured effective name, which defaults to the actor system name
     pod-label-selector = "app=%s"
+
+    # Enables the usage of the raw IP instead of the composed value for the resolved target host
+    raw-ip-enabled = false
   }
 }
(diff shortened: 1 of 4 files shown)


discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -31,5 +31,8 @@ akka.discovery {
     # Selector value to query pod API with.
     # `%s` will be replaced with the configured effective name, which defaults to the actor system name
     pod-label-selector = "app=%s"
+
+    # Enables the usage of the raw IP instead of the composed value for the resolved target host
+    raw-ip-enabled = false
```
ennru, line 36: Maybe `use-raw-ip`?

sebastian-alfers: Sure, that should also work 👍

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala
```diff
@@ -63,9 +64,9 @@ object KubernetesApiServiceDiscovery {
           } yield Some(port.containerPort)
       }
     } yield {
-      val host = s"${ip.replace('.', '-')}.$podNamespace.pod.$podDomain"
+      val hostOrIp = if (rawIp) ip else s"${ip.replace('.', '-')}.$podNamespace.pod.$podDomain"
```
raboof, line 67: does this need `[]` brackets when it's an IPv6 IP?

sebastian-alfers: I don`t know, but according to the request for this change, this was not mentioned.

PR #670, opened 2020-05-08
Simplify releasing

ennru:
* revise release checklist and add script to create the issue
* configure release drafter via Github actions
* add PR autolabeler config
* introduce "integration" stage for Travis

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,23 @@
+# Config for https://github.com/toolmantim/release-drafter
+name-template: 'Akka Management $NEXT_PATCH_VERSION'
+tag-template: 'v$NEXT_PATCH_VERSION'
+categories:
+  - title: 'Cluster Bootstrap'
+    label: 't:bootstrap'
+  - title: 'Cluster HTTP Management'
+    label: 't:cluster-http'
+  - title: 'Discovery: AWS'
+    label: 't:aws'
+  - title: 'Discovery: Consul'
+    label: 't:consul'
+  - title: 'Discovery: Kubernetes'
+    label: 't:kubernetes'
+  - title: 'Discovery: Marathon'
+    label: 't:marathon'
+
+
+change-template: '- $TITLE [#$NUMBER](https://github.com/akka/alpakka/issues/$NUMBER) by [@$AUTHOR](https://github.com/$AUTHOR)'
+template: |
+  ## Changes
+
+  $CHANGES
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -16,7 +16,7 @@ categories:
     label: 't:marathon'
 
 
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -39,21 +37,20 @@ jobs:
...
     - sbt -jvm-opts .jvmopts-travis ";scalafmtCheckAll;+test:compile;docs/paradox"
     name: "Compile on all Scala versions, check code style (fixed with `sbt Test/compile`)"
 
-  - stage: test
-    env: CMD="+test"
-    name: "Run tests on AdoptOpenJDK 8 (all Scala versions)"
-
   - env:
-    - JDK="adopt@1.11-0"
+    - ADOPTOPENJDK=8
```
patriknw, line 45: so this is built into travis? is this an experiment to see if it works better than Jabba or do we already know that this is what we shall use everywhere?

raboof: It is imported from https://github.com/scala/scala-dev/blob/scala-dev/travis/default.yml (see also https://twitter.com/SethTisue/status/1239988439199580160).

I introduced it in https://github.com/akka/akka-http/pull/3137 because jabba was lagging in supported jdk version, but I don't have actual experience with it myself yet

patriknw: ok, the goal should be that we use same everywhere so we don't have to trouble shoot several things, but maybe let this mature in akka-http and here before rolling it out in all places

ennru: We've seen our use of jabba failing quite often recently. But maybe I should remove it again so that we see how sdkman behaves in the more active Akka HTTP repository instead.

.github/release-drafter.yml
```diff
@@ -0,0 +1,23 @@
...
+    label: 't:aws'
+  - title: 'Discovery: Consul'
+    label: 't:consul'
+  - title: 'Discovery: Kubernetes'
+    label: 't:kubernetes'
+  - title: 'Discovery: Marathon'
+    label: 't:marathon'
+
+
+change-template: '- $TITLE [#$NUMBER](https://github.com/akka/alpakka/issues/$NUMBER) by [@$AUTHOR](https://github.com/$AUTHOR)'
```
ihostage, line 19: `$NUMBER` is a number of PR, not an issue 😞 
We use a rule, that subject of pull request  **MUST** start from the number of related issue.
Then the result of Release Drafter looks very informative.(example https://github.com/taymyr/lagom-openapi/releases/tag/v1.1.0)

ihostage: And `alpakka` looks like a copy&paste 😄

ennru: PRs and issues shared the same sequence. You can refer to a PR that way.

Thanks for the copy&waste detection.

.travis.yml
ennru: Kept jabba for now.

Issue #673, opened 2020-05-18
Failed: minikube install


chbatey:
CI uses the latest k8s/minikube version. That has broken recently.

```
$ java -Xmx32m -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
$ javac -J-Xmx32m -version
javac 11.0.2
Using Scala 2.12.8
before_install.1
0.74s$ git remote set-branches --add origin master && git fetch
before_install.2
2.79s$ git fetch --unshallow
before_install.3
1.23s$ curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.1/install.sh | bash && . ~/.jabba/jabba.sh
before_install.4
1.14s$ jabba install adopt@1.8-0
before_install.5
0.46s$ jabba install adopt@1.11-0
before_script
0.00s$ unset _JAVA_OPTIONS
9.35s$ ./scripts/setup-minikube-for-linux.sh && ./integration-test/kubernetes-api/test.sh
+MINIKUBE_VERSION=latest
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 52.7M  100 52.7M    0     0   106M      0 --:--:-- --:--:-- --:--:--  106M
+chmod +x minikube
+sudo cp minikube /usr/local/bin/
+rm minikube
++curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.18.2/bin/linux/amd64/kubectl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 41.9M  100 41.9M    0     0   153M      0 --:--:-- --:--:-- --:--:--  152M
+chmod +x kubectl
+sudo cp kubectl /usr/local/bin/
+rm kubectl
+export MINIKUBE_WANTUPDATENOTIFICATION=false
+MINIKUBE_WANTUPDATENOTIFICATION=false
+export MINIKUBE_WANTREPORTERRORPROMPT=false
+MINIKUBE_WANTREPORTERRORPROMPT=false
[...]

PR #674, opened 2020-05-18
Install contrack

chbatey:
Refs #673

### DIFF
--- a/scripts/setup-minikube-for-linux.sh
+++ b/scripts/setup-minikube-for-linux.sh
@@ -2,6 +2,10 @@
 
 set -exu
 
+# Required by minikube
+sudo apt update
+sudo apt install conntrack
+
 # MINIKUBE_VERSION="latest" -- https://github.com/kubernetes/minikube/issues/2704
 MINIKUBE_VERSION="latest"
 


chbatey:
Looks to work. I'll do the same for akka-upgrade-testing

scripts/setup-minikube-for-linux.sh
```diff
@@ -2,6 +2,10 @@
 
 set -exu
 
+# Required by minikube
+sudo apt update
+sudo apt install conntrack
```
raboof, line 7: doing it like https://github.com/akka/akka-sample-cluster-kubernetes-java/blob/master/.travis.yml#L8 might save some time 'update'-ing, but this is :+1: as well

chbatey: Hrmm we already have that in our travis and doesn't seem to work for minikube

raboof: Huh, so #649 worked for a while and then stopped working again? Weird.

chbatey: Current builds are now passing so i'll close this

PR #692, opened 2020-06-09
minor typo in doc forming-a-cluster.md

aldenml:


aldenml:
already two PRs with the same fix

Issue #679, opened 2020-06-02
CI for Kubernetes lease integration tests


chbatey:
Setup minikube for being able to test the lease

PR #694, opened 2020-06-15
Lease tests in travis minikube

chbatey:
https://github.com/akka/akka-management/issues/679

I expected that to be more work but easy given we already run other tests in minikube


chbatey:
I am not sure what is going on with travis reporting, CI has run and passed

PR #721, opened 2020-07-09
Update jackson-databind to 2.10.4

raboof:
To keep consistent with Akka

Let's see if this conflicts with the AWS SDK

If this works, it superseded #708 and #709

Otherwise we might first want to update to awssdk2...


ennru:
Jackson databind 2.10.5 is available now.

chbatey:
We upgraded to this else where

PR #723, opened 2020-07-09
ScalaTest 3.1.1 (was 3.0.8) with rewrites

ennru:
This is required to enable the upgrade to 3.2 in #720

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -2,6 +2,8 @@ pullRequests.frequency = "@monthly"
 
 updates.ignore = [
   { groupId = "com.typesafe.akka" }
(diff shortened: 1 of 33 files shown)


.scala-steward.conf
```diff
@@ -2,6 +2,8 @@ pullRequests.frequency = "@monthly"
 
 updates.ignore = [
   { groupId = "com.typesafe.akka" }
+  // Keep 2.12 until ScalaTest+ Junit upgrades https://github.com/scalatest/scalatestplus-junit/releases
+  { groupId = "junit", artifactId = "junit" }
```
raboof, line 6: should we depend on junit at all? presumably scalatestplus-junit brings it in?

PR #726, opened 2020-08-03
Akka HTTP 10.2 RC2 in the scheduled build

ennru:
https://doc.akka.io/docs/akka-http/10.2/release-notes/10.2.x.html


jrudolph:
Scheduled another cron job to run right now, so we'll have quick results.

ennru:
Thanks, it showed the obvious so I created https://github.com/akka/akka-management/pull/727

PR #727, opened 2020-08-04
build: Scala 2.12 and 2.13 in scheduled build

ennru:
The scheduled build uses Akka 2.6 and Akka HTTP 10.2 which are not available for Scala 2.11.


jrudolph:
Now failing with test problems: https://travis-ci.org/github/akka/akka-management/jobs/714783468

PR #731, opened 2020-08-06
Akka HTTP 10.1.11 in Integration test Kubernetes API with maven

ennru:
Use the same versions for compilation and the tests in the Maven project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ cache:
   directories:
     - $HOME/.cache/coursier
     - $HOME/.ivy2/cache
+    - $HOME/.m2/repository # for Maven-based integration tests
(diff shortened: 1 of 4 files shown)


ennru:
The Maven-based project failed in the scheduled build.
It uses an Akka Management version compiled against Akka HTTP 10.2.0 which implements the extension getters with `ClassicActorSystemProvider`, but used Akka HTTP 10.1.5 for running the tests which doesn't have this method.
A full-blown solution might be to make both the Akka version and the Akka HTTP version dynamic in the scheduled build even for the Maven project.

ennru:
This wasn't enough to make the dependencies work. https://travis-ci.org/github/akka/akka-management/jobs/715813085#L718

Issue #746, opened 2020-08-13
Failed: Travis integration tests fail because ingress cannot be enabled


jrudolph:
https://travis-ci.org/github/akka/akka-management/jobs/717561226

```
+sudo -E minikube addons enable ingress

* Due to networking limitations of driver none, ingress addon is not supported. Try using a different driver.

The command "./scripts/setup-minikube-for-linux.sh && ./integration-test/kubernetes-api/test.sh" exited with 64.
```

seglo:
https://travis-ci.org/github/akka/akka-management/jobs/720619844#L284

PR #747, opened 2020-08-17
Use docker minikube driver

chbatey:


scripts/setup-minikube-for-linux.sh
```diff
@@ -17,8 +17,8 @@ mkdir -p $HOME/.kube
 touch $HOME/.kube/config
 
 export KUBECONFIG=$HOME/.kube/config
-sudo -E minikube start --vm-driver=none
-sudo -E minikube addons enable ingress
+minikube start --driver=docker
```
patriknw, line 20: do we have to update deployment guide and samples also?

chbatey: i'll check. we shouldn't be recommending the any driver but maybe we do

chbatey: We just assume users have minikube in the akka management one: https://doc.akka.io/docs/akka-management/current/kubernetes-deployment/deploying.html#deploying-to-minikube
and we link to the docs from the Lagom one

raboof:
Should fix #746

chbatey:
Merging this as all other PR validation is broken

PR #749, opened 2020-08-25
Setting for all contact points being up

chbatey:


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -280,9 +278,6 @@ private[akka] class BootstrapCoordinator(
 
             val seedNodesList = (seedNodes - cluster.selfAddress).toList // order doesn't matter
             cluster.joinSeedNodes(seedNodesList)
-
-            // once we issued a join bootstrapping is completed
-            replyTo ! BootstrappingCompleted
```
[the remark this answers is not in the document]
chbatey: 👍 re-added

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
```diff
@@ -45,8 +45,6 @@ private[akka] object BootstrapCoordinator {
 
   object Protocol {
     final case class InitiateBootstrapping(selfContactPoint: Uri)
-    sealed trait BootstrappingCompleted
-    final case object BootstrappingCompleted extends BootstrappingCompleted
```
johanandren, line 49: I don't quite get how adding that all contact points setting makes this response superflous.

cluster-bootstrap/src/main/resources/reference.conf
```diff
@@ -95,6 +95,13 @@ akka.management {
 
       # Timeout for getting a reply from the service-discovery subsystem
       resolve-timeout = 3 seconds
+
+      # Does a succcessful response have to be received by all contact points.
+      # Used by the LowestAddressJoinDecider
+      # Can be set to false in environments where old contact points may still be in service discovery
+      # or when using local discovery and cluster formation is desired without starting all the nodes
+      # Required-contact-point-nr still needs to be met
```
johanandren, line 103: So, with this changed to `false`, say it would discover 10 nodes, but not get response from 5, then as long as `required-contact-points = 2` is fulfilled it will go ahead and form a cluster?

What is a scenario where "old contact points may still be in service discovery"?

patriknw (approved):
LGTM

cluster-bootstrap/src/main/resources/reference.conf
chbatey: That's right.

>What is a scenario where "old contact points may still be in service discovery"?

For instance when they are in a DNS cache

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/internal/BootstrapCoordinator.scala
chbatey: It was an unrelated change. This always went to dead letters

PR #750, opened 2020-08-28
Access Akka HTTP more explicitly to achieve some backwards compatiblity

### LINKED ISSUE #744: Failed: Exception in thread "main" java.lang.NoSuchMethodError: akka.http.scaladsl.Http$.apply(Lakka/actor/ClassicActorSystemProvider;)Lakka/http/scaladsl/HttpExt;
This is the maven integration test:

https://travis-ci.org/github/akka/akka-management/jobs/716528114

ennru:
This should fix #744.

### DIFF
--- a/management/src/main/scala/akka/management/scaladsl/AkkaManagement.scala
+++ b/management/src/main/scala/akka/management/scaladsl/AkkaManagement.scala
@@ -145,11 +145,12 @@ final class AkkaManagement(implicit private[akka] val system: ExtendedActorSyste
 
         val combinedRoutes = prepareCombinedRoutes(effectiveProviderSettings)
 
+        val http = Http.get(system)
         val connectionContext =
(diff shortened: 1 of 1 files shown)


raboof (commented):
If I understand correctly this fixes compiling Akka Management against Akka HTTP 10.2.0 and then using those artifacts in a project with Akka HTTP 10.1.x.

The Akka Managment artifacts we publish are built against Akka HTTP 10.1.x.

I guess building against Akka HTTP 10.2 on CI can sometimes catch things, but then making non-idiomatic changes to allow testing that artifact with Akka HTTP 10.1 on the classpath seems like a slippery slope. I'd rather build with 10.1.x and test with 10.2.x, for example.

Or, if we want to stick to building with 10.2, perhaps just skip the maven test in that case.

ennru:
I agree. Let's just exclude that validation step in the scheduled build.

PR #766, opened 2020-09-16
small fixes in deployment guide

patriknw:


raboof (approved):
Not sure if it really *needs* to be in `kubernetes/akka-cluster.yaml`, but good to give direction.

patriknw:
> Not sure if it really needs to be in kubernetes/akka-cluster.yaml

That is what is used later in the text and what is in the example project.

PR #770, opened 2020-09-23
cluster-http: support IPv6 addresses in cluster http routes

### LINKED ISSUE #769: Cluster Management Routes do not parse members with IPv6 address correctly
If you have a member whose address contains an IPv6 address enclosed in brackets like `[::1]`, you will have to use percent encoding to pass those brackets inside of the path like this:

```
http://[::1]:25261/cluster/members/ClusterSystem%40%5B%3A%3A1%5D%3A25251
```

However, an existing node will not be found in this case and this error will be shown:

```
{"message":"Member [ClusterSystem@%5B::1%5D:25251] not found"}
```

So, it looks like some percent encodings are decoded but others are not. However, that's not the reason. All of the percent encodings actually are decoded but the directive used to extract the path is `path(Remaining)` in

https://github.com/akka/akka-management/blob/af0aae0f3f03abb6f0631d01132a75cffaf6f31d/cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterHttpManagementRoutes.scala#L98

An that PathMatcher uses `path.toString` which tries to make a well-formed path again containing percent-encodings but only where needed. Brackets need to be encoded so the member is extracted as seen above.

The easiest would be to replace `path(Remaining)` with `path(Segment)`, however, that might fail for addresses containing schemes and slashes (if not percent encoded).

I guess the least confusing option might be to try to keep the current behavior for compatibility but also add the possibility to use `/cluster/members?address=` syntax which doesn't have the confusing problem of how to interpret slashes.

jrudolph:
Fixes #769

### DIFF
--- a/cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
+++ b/cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
@@ -228,7 +228,7 @@ class ClusterHttpManagementRoutesSpec
         doNothing().when(mockedCluster).down(any[Address])
 
         Put(s"/cluster/members/$address", urlEncodedForm) ~> ClusterHttpManagementRoutes(mockedCluster) ~> check {
-          status shouldEqual StatusCodes.OK
+          status shouldEqual StatusCodes.NotFound
           responseAs[ClusterHttpManagementMessage] shouldEqual ClusterHttpManagementMessage(
             s"Member [$address] not found")
         }
@@ -285,7 +285,7 @@ class ClusterHttpManagementRoutesSpec
 
         Seq("akka://Main2@hostname.com:3311", "Main2@hostname.com:3311").foreach(address => {
           Put(s"/cluster/members/$address", urlEncodedForm) ~> ClusterHttpManagementRoutes(mockedCluster) ~> check {
-            status shouldEqual StatusCodes.OK
+            status shouldEqual StatusCodes.NotFound
             responseAs[ClusterHttpManagementMessage] shouldEqual ClusterHttpManagementMessage(
               s"Member [$address] not found")
           }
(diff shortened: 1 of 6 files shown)


cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -225,9 +228,9 @@ class ClusterHttpManagementRoutesSpec
         doNothing().when(mockedCluster).down(any[Address])
 
         Put(s"/cluster/members/$address", urlEncodedForm) ~> ClusterHttpManagementRoutes(mockedCluster) ~> check {
+          status shouldEqual StatusCodes.OK
```
raboof, line 231: Should this now return OK for addresses that are not found? That's not obvious to me.

jrudolph: Oops, bad copy pasta.

cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -311,8 +314,8 @@ class ClusterHttpManagementRoutesSpec
 
         Seq("akka://Main@hostname.com:3311", "Main@hostname.com:3311").foreach(address => {
           Put(s"/cluster/members/$address", urlEncodedForm) ~> ClusterHttpManagementRoutes(mockedCluster) ~> check {
+            status shouldEqual StatusCodes.BadRequest
```
jrudolph, line 317: All these checks did not work before, and in this case it was even wrong...

cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
jrudolph: Thanks for catching. Now fixed.

johanandren (approved):

ennru (approved):
LGTM.

PR #790, opened 2020-10-19
Restructure dependencies declaration

ennru:
Maybe it is just me but the structure in `Dependencies` encrypts what dependencies are actually used/available/exported in each module.

This makes the dependencies explicit by less indirection.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ lazy val `akka-discovery-kubernetes-api` = project
   .settings(
     name := "akka-discovery-kubernetes-api",
     organization := "com.lightbend.akka.discovery",
-    Dependencies.DiscoveryKubernetesApi
+    libraryDependencies := Dependencies.DiscoveryKubernetesApi
   )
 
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -68,7 +68,7 @@ lazy val `akka-discovery-aws-api-async` = project
   .settings(
     name := "akka-discovery-aws-api-async",
     organization := "com.lightbend.akka.discovery",
-    Dependencies.DiscoveryAwsApiAsync
+    libraryDependencies := Dependencies.DiscoveryAwsApiAsync
```
raboof, line 71: I prefer `++=` as a general style (since it makes builds more composable), though in this case it probably doesn't really matter.

raboof (approved):
a bit more duplication, but the reduced indirection seems worth it.

build.sbt
ennru: I like the WYSIWYG of `:=`. The old setup injecting libraries from `Common` was a bit surprising to me.

PR #792, opened 2020-10-20
Allow sbt to run in parallel (but not in tests)

ennru:
[`32de629` (#205)](https://github.com/akka/akka-management/pull/205/commits/32de6296bb5858f8a59f4937bfd3de767422a76a) switched off all parallel execution in sbt.

With this only tests run sequentially.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,7 +31,6 @@ lazy val `akka-management-root` = project
     docs
   )
   .settings(
-    parallelExecution in GlobalScope := false,
     publish / skip := true
   )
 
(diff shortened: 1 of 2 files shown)


raboof (approved):

ennru:
Interesting, with this the different projects run their tests in parallel which seems to trigger https://travis-ci.org/github/akka/akka-management/jobs/737347515#L830 

Not sure if it's worth to investigate this further.

patriknw:
For the record, we had some surprising moments in Akka Projections also. https://github.com/akka/akka-projection/pull/380
That involved moving some tests to integration test folder and such to be able to run some in parallel for local dev. I'm not convinced that complexity was worth it, and it's difficult to troubleshoot when something fails.

Issue #795, opened 2020-10-30
SelfAwareJoinDecider does not work for IPv6 hosts


djm1329:
### Versions used 

Akka Management version 1.0.9

Akka version: 2.6.10

### Expected Behavior

Using Akka Management on an IPv6 host, Akka Cluster Bootstrap SelfAwareJoinDecider should allow lowest address to join itself, and initiate cluster formation. (Observed when using akka-dns contact point discovery)

### Actual Behavior

No cluster is formed since the node with lowest address never joins itself.

### Relevant logs

Here are log snippets, one from each of two nodes:

```
WARN  LowestAddressJoinDecider LowestAddressJoinDecider(akka://ccs) ccs-akka.actor.default-dispatcher-5 - Self contact point [[240b:c0e0:202:5e2b:b424:2:0:450]:9231] not found in targets ResolvedTarget(240b:c0e0:202:5e2b:b424:2:0:cc4,None,Some(/240b:c0e0:202:5e2b:b424:2:0:cc4)), ResolvedTarget(240b:c0e0:202:5e2b:b424:2:0:450,None,Some(/240b:c0e0:202:5e2b:b424:2:0:450))
2020-10-27T17:03:56.060 INFO  LowestAddressJoinDecider LowestAddressJoinDecider(akka://ccs) ccs-akka.actor.default-dispatcher-5 - Exceeded stable margins without locating seed-nodes, however this node [240b:c0e0:202:5e2b:b424:2:0:450]:9231 is NOT the lowest address out of the discovered endpoints in this deployment, thus NOT joining self. Expecting node [240b:c0e0:202:5e2b:b424:2:0:450:0] (out of [240b:c0e0:202:5e2b:b424:2:0:cc4:0, 240b:c0e0:202:5e2b:b424:2:0:450:0]) to perform the self-join and initiate the cluster.


WARN  LowestAddressJoinDecider LowestAddressJoinDecider(akka://ccs) ccs-akka.actor.default-dispatcher-27 - Self contact point [[240b:c0e0:202:5e2b:b424:2:0:cc4]:9231] not found in targets ResolvedTarget(240b:c0e0:202:5e2b:b424:2:0:cc4,None,Some(/240b:c0e0:202:5e2b:b424:2:0:cc4)), ResolvedTarget(240b:c0e0:202:5e2b:b424:2:0:450,None,Some(/240b:c0e0:202:5e2b:b424:2:0:450))
[...]

PR #796, opened 2020-10-30
ignore [ and ] in host name when checking if a node can join itself

djm1329:
References https://github.com/akka/akka-management/issues/795


raboof (approved):

chbatey (approved):
LGTM, thanks!

PR #812, opened 2020-11-25
Update KubernetesApiImpl.scala to prevent uppcase causing exception in k8s

nathanleyton:
issue: https://github.com/akka/akka-management/issues/811

### DIFF
--- a/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
+++ b/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
@@ -4,7 +4,7 @@
 
 package akka.coordination.lease.kubernetes.internal
 
-import java.nio.file.{Files, Paths}
+import java.nio.file.{ Files, Paths }
 
 import akka.Done
 import akka.actor.ActorSystem
@@ -13,11 +13,11 @@ import akka.event.Logging
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshalling.Marshal
 import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken}
+import akka.http.scaladsl.model.headers.{ Authorization, OAuth2BearerToken }
 import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.pattern.after
-import akka.coordination.lease.kubernetes.{KubernetesApi, KubernetesSettings, LeaseResource}
-import akka.coordination.lease.{LeaseException, LeaseTimeoutException}
+import akka.coordination.lease.kubernetes.{ KubernetesApi, KubernetesSettings, LeaseResource }
+import akka.coordination.lease.{ LeaseException, LeaseTimeoutException }
 import akka.stream.ActorMaterializer
 import com.typesafe.sslconfig.akka.AkkaSSLConfig
 import com.typesafe.sslconfig.ssl.TrustStoreConfig
@@ -226,7 +226,9 @@ PUTs must contain resourceVersions. Response:
   }
 
   private def pathForLease(name: String): Uri.Path =
-    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name.replaceAll("[^\\d|\\w|\\-|\\.]", "").toLowerCase
+    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name
+      .replaceAll("[^\\d|\\w|\\-|\\.]", "")
+      .toLowerCase
 
   private def requestForPath(
       path: Uri.Path,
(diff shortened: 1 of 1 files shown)


nathanleyton:
Test fails with or without my commit for this issue. I will try and see what the cause of the failed test is.

raboof:
> Test fails with or without my commit for this issue.

That's weird, it succeeded for master relatively recently (https://travis-ci.com/github/akka/akka-management/jobs/446315936)

> I will try and see what the cause of the failed test is.

Thanks!

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
```diff
@@ -227,7 +227,7 @@ PUTs must contain resourceVersions. Response:
   }
 
   private def pathForLease(name: String): Uri.Path =
-    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name
+    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name.toLowerCase
```
raboof, line 230: Aside from lower-casing, perhaps we should remove more invalid characters from the name?

nathanleyton: Hows that?

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
```diff
@@ -227,7 +227,7 @@ PUTs must contain resourceVersions. Response:
   }
 
   private def pathForLease(name: String): Uri.Path =
-    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name
+    Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name.toLowerCase.filter(c => "\\d|\\w|-|\\.".r.findAllIn(c.toString).nonEmpty)
```
raboof, line 230: Seems a bit 'heavy' to use a regex for this (and construct it for every single character), but this is probably not called often so might not be a problem.

I like the new behavior :).

nathanleyton: Yeah I couldn't think of another way to do it easily without having a list of characters either allowed or not allowed. Perhaps a list of allowed is better?

raboof: maybe something like `name.toLowerCase.filter(c => c.isLetterOrDigit || c == '-' || c == '.')`?

nathanleyton: I do it better tomorrow. I left my computer now. I also noticed the test failure was due to upping version of Akka Http and now it wants us to specify the content type. I managed to get the test to pass before I left . I push tomorrow.

raboof: Great, looking forward to it!

nathanleyton: After various performance tests, the one I just pushed was by far the most performant. 

with 10K iteration 
0.77 seconds for the new one 
vs
16.61 sec for your one (which, by the way was more performant that my initial attempt) 

Screenshot from 2020-11-26 09-51-55

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
```diff
@@ -228,6 +227,8 @@ PUTs must contain resourceVersions. Response:
 
   private def pathForLease(name: String): Uri.Path =
     Uri.Path.Empty / "apis" / "akka.io" / "v1" / "namespaces" / namespace / "leases" / name
+      .replaceAll("[^\\d|\\w|\\-|\\.]", "")
```
raboof, line 230: You don't need `|` characters in a regex
```suggestion
      .replaceAll("[^\\d\\w\\-\\.]", "")
```

raboof (commented):
Almost there :D

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/internal/KubernetesApiImpl.scala
nathanleyton: Oh right , I thought the or works more like an if else so is mutually exclusive and in this case should be more performant. You don't need but unless I have not understood something without it will match like a bunch of ifs

raboof: No, in this case the `|` character is just one of the characters in the `[]`, so this would be wrong for entities with `|` characters in their name:

```
@ "asdf|.-!000".replaceAll("[^\\d|\\w|\\-|\\.]", "") 
res0: String = "asdf|.-000"
```

nathanleyton: Maybe it needs to be escaped but I really doubt it will make much difference at all to the performance so let's run with your last one 🙂 thanks for your time on it.

raboof (approved):

chbatey (approved):

Issue #768, opened 2020-09-18
Failed: KubernetesApiSpec => LeaseTimeoutException, with message: Timed out removing lease [lease-1]


octonato:
https://travis-ci.org/github/akka/akka-management/jobs/728309081#L1161

```scala
[info] - should be able to be created *** FAILED *** (2 seconds, 39 milliseconds)
[info]   The future returned an exception of type: akka.coordination.lease.LeaseTimeoutException, with message: Timed out removing lease [lease-1]. It is not known if the remove happened. Is the API server up?. (KubernetesApiSpec.scala:83)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:531)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.coordination.lease.kubernetes.KubernetesApiSpec.$anonfun$new$2(KubernetesApiSpec.scala:83)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1076)
[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)

```

raboof:
https://travis-ci.org/github/akka/akka-management/jobs/733642654

ennru:
[...]

Issue #821, opened 2020-12-11
KubernetesApiIntegrationTest: 7 or 8 tests failed


ennru:
```
Kubernetes lease resource
- should be able to be created *** FAILED ***
  The future returned an exception of type: akka.coordination.lease.LeaseTimeoutException, with message: Timed out removing lease [lease-1]. It not known if the remove happened. Is the API server up?. (KubernetesApiIntegrationTest.scala:56)

- should be able to read back with same version *** FAILED ***
  "[595]" did not equal "[]" (KubernetesApiIntegrationTest.scala:66)
  Analysis:
  "[595]" -> "[]"

- should be able to take a lease with no owner *** FAILED ***
  The future returned an exception of type: akka.coordination.lease.LeaseException, with message: PUT for lease lease-1 returned unexpectedatus code 422 Unprocessable Entity. Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"leases.akka.io \"lease-1\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update","reason":"Invalid","details":{"name":"lease-1","group":"akka.io","kind":"leases","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: 0x0: must be specified for an update","field":"metadata.resourceVersion"}]},"code":422}
  . (KubernetesApiIntegrationTest.scala:70)

- should be able to update a lease if resource version is correct *** FAILED ***
  The future returned an exception of type: akka.coordination.lease.LeaseException, with message: PUT for lease lease-1 returned unexpectedatus code 422 Unprocessable Entity. Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"leases.akka.io \"lease-1\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update","reason":"Invalid","details":{"name":"lease-1","group":"akka.io","kind":"leases","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: 0x0: must be specified for an update","field":"metadata.resourceVersion"}]},"code":422}
  . (KubernetesApiIntegrationTest.scala:81)

- should not be able to update a lease if resource version is correct *** FAILED ***
  "[595]" did not equal "[]" (KubernetesApiIntegrationTest.scala:97)
  Analysis:
  "[595]" -> "[]"

- should be able to remove ownership *** FAILED ***
[...]

PR #822, opened 2020-12-14
Harden KubernetesApiIntegrationSpec

chbatey:
The test fails when the first request to the API server times out. So:

* Cancel remaining tests on failure, they are dependent on each other
* Check API server connectivity before starting the tests

Refs #821


chbatey:
Also fixes https://github.com/akka/akka-management/issues/768

johanandren (approved):

Issue #540, opened 2019-05-09
SSE (Server-Sent-Events) endpoint for cluster events


dbrinegar:
### Short description

An option under Akka Management HTTP for SSE to expose various events, so an external http client could observe a stream of Cluster Events, for example. This could be extended with various real time diagnostics, in future.

### Details

An HTTP event interface to Akka Management events would facilitate external integrations for publishing, reporting, responding to various Cluster Events. For example, an external cluster status report could listen for updates on a long lived connection instead of polling the http endpoint. Such an integration point for low frequency events could not only improve resource utilization (only gather status when it changes, instead of every N seconds) but also make external response time to cluster change more lively.

references:

https://doc.akka.io/docs/akka-http/current/sse-support.html

https://doc.akka.io/docs/akka/current/cluster-usage.html#subscribe-to-cluster-events

longshorej:
Hey everyone - I'm working on this, hope to have a PR in a couple days :)

patriknw:
Nice, thanks.

PR #836, opened 2021-01-07
cluster-http: Add a route for observing cluster events

longshorej:
This PR adds a new route to cluster-http, `/cluster/domain-events`, that streams cluster events as Server Sent Events, JSON encoded. Please see the included documentation for route parameters, response format, etc.

ref: #540

### DIFF
--- a/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
+++ b/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
@@ -61,12 +61,12 @@ object ClusterDomainEventServerSentEventEncoder extends SprayJsonSupport with De
         roleLeaderChanged.leader match {
           case Some(address) =>
             sse(
-              "LeaderChanged",
+              "RoleLeaderChanged",
               JsObject("role" -> JsString(roleLeaderChanged.role), "address" -> encode(address))
             )
 
           case None =>
-            sse("LeaderChanged", JsObject("role" -> JsString(roleLeaderChanged.role)))
+            sse("RoleLeaderChanged", JsObject("role" -> JsString(roleLeaderChanged.role)))
         }
 
       case ClusterEvent.ClusterShuttingDown =>
--- a/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
+++ b/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
@@ -46,6 +46,9 @@ object ClusterDomainEventServerSentEventEncoder extends SprayJsonSupport with De
 
           case ClusterEvent.MemberRemoved(member, previousStatus) =>
             sse("MemberRemoved", JsObject("member" -> encode(member), "previousStatus" -> encode(previousStatus)))
+
+          case _ =>
+            None
         }
 
       case leaderChanged: ClusterEvent.LeaderChanged =>
@@ -90,8 +93,8 @@ object ClusterDomainEventServerSentEventEncoder extends SprayJsonSupport with De
         }
 
       case _ =>
-        // @TODO these are either internal, or added but forgot to update this module
-        // @TODO how should we handle them? maybe leave this as FIXME for now?
+        // these are either internal events we don't want to expose, or new events
+        // that this code is unaware of. either way, let's ignore them
 
         None
     }
@@ -113,6 +116,7 @@ object ClusterDomainEventServerSentEventEncoder extends SprayJsonSupport with De
       case MemberStatus.Exiting  => "Exiting"
       case MemberStatus.Down     => "Down"
       case MemberStatus.Removed  => "Removed"
+      case other                 => other.toString
     }
   )
 
--- a/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
+++ b/cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2017-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.management.cluster
+
+import akka.actor.Address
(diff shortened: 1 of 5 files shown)


longshorej:
demo project is here: https://github.com/longshorej/akka-cluster-membership-events

longshorej:
Hi team - this is ready for a review, let me know what y'all think - happy to discuss/change as required.

cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
```diff
@@ -0,0 +1,125 @@
...
+
+          case ClusterEvent.MemberExited(member) =>
+            sse("MemberExited", JsObject("member" -> encode(member)))
+
+          case ClusterEvent.MemberDowned(member) =>
+            sse("MemberDowned", JsObject("member" -> encode(member)))
+
+          case ClusterEvent.MemberRemoved(member, previousStatus) =>
+            sse("MemberRemoved", JsObject("member" -> encode(member), "previousStatus" -> encode(previousStatus)))
+        }
```
raboof, line 49: Maybe just to future-proof we should add a `case _ =>` here?

cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
```diff
@@ -0,0 +1,125 @@
...
+        dataCenterReachabilityEvent match {
+          case ClusterEvent.UnreachableDataCenter(dataCenter) =>
+            sse("UnreachableDataCenter", JsObject("dataCenter" -> JsString(dataCenter)))
+          case ClusterEvent.ReachableDataCenter(dataCenter) =>
+            sse("ReachableDataCenter", JsObject("dataCenter" -> JsString(dataCenter)))
+        }
+
+      case _ =>
+        // @TODO these are either internal, or added but forgot to update this module
+        // @TODO how should we handle them? maybe leave this as FIXME for now?
```
raboof, line 94: I guess our options are to either crash, silently ignore them, or log a message. Silently ignoring them might be fine...

cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  private def encode(memberStatus: MemberStatus): JsValue = JsString(
+    memberStatus match {
+      case MemberStatus.Joining  => "Joining"
+      case MemberStatus.WeaklyUp => "WeaklyUp"
+      case MemberStatus.Up       => "Up"
+      case MemberStatus.Leaving  => "Leaving"
+      case MemberStatus.Exiting  => "Exiting"
+      case MemberStatus.Down     => "Down"
+      case MemberStatus.Removed  => "Removed"
```
raboof, line 115: maybe `case other => other.toString`?

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterHttpManagementRoutes.scala
```diff
@@ -113,6 +113,108 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
...
+          val classes =
+            if (providedEventTypes.nonEmpty)
+              providedEventTypes.foldLeft(List.empty[Class[_]]) {
+                case (accum, eventType) =>
+                  eventClasses.get(eventType).toList ::: accum
+              }
+            else
+              List(classOf[ClusterEvent.ClusterDomainEvent])
+
+          val eventualActorRef = Promise[Option[ActorRef]]
```
raboof, line 160: I think this could also be a `Promise[ActorRef]`, which we could `tryFailure` when the stream completes before successfully completing the promise. This is okay as well, though.

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterHttpManagementRoutes.scala
```diff
@@ -113,6 +113,108 @@ object ClusterHttpManagementRoutes extends ClusterHttpManagementJsonProtocol {
...
+              case Some(serverSentEvent) => serverSentEvent
+            }
+            .keepAlive(10.seconds, () => ServerSentEvent.heartbeat)
+            .mapMaterializedValue { actorRef =>
+              eventualActorRef.success(Some(actorRef))
+              ()
+            }
+            .watchTermination() {
+              case (_, eventualDone) =>
+                eventualDone.onComplete { _ =>
```
raboof, line 180: :+1:

raboof (approved):
A couple of small comments, but overall LGTM - and a useful feature!

cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
patriknw: I think some more are internal, SeenChanged.
Not often added but we actually have one coming soon. Let's ignore them here.

cluster-http/src/main/scala/akka/management/cluster/ClusterDomainEventServerSentEventEncoder.scala
patriknw: sounds good to handle that gracefully, another is coming soon https://github.com/akka/akka/pull/29838

patriknw (commented):
@longshorej  looking great, something minor and then we can merge

cluster-http/src/main/scala/akka/management/cluster/scaladsl/ClusterHttpManagementRoutes.scala
longshorej: Yeah, I see that's nicer in a way then since success doesn't have the option to deal with. I don't like having to pick an exception type though (probably should be `NoStackTrace`), so rather leave it as-is.

longshorej:
Thanks for the review @raboof and @patriknw. I think I've addressed your feedback, ready to merge when y'all are :)

patriknw (approved):
LGTM

PR #841, opened 2021-02-04
fix missing SIGTERM: use exec format in docker-maven-plugin

patriknw:
See https://github.com/akka/akka-platform-guide/pull/543

### DIFF
--- a/docs/src/main/paradox/kubernetes-deployment/building-using-maven.md
+++ b/docs/src/main/paradox/kubernetes-deployment/building-using-maven.md
@@ -60,8 +60,13 @@ The version number will only change when you create a new git commit. If you are
                         <tag>${version.number}</tag>
                     </tags>
                     <entryPoint>
-                       java $JAVA_OPTS -cp '/maven/*' akka.cluster.bootstrap.demo.DemoApp
-                    </entryPoint> 
+                        <exec>
+                            <arg>java</arg>
+                            <arg>-cp</arg>
+                            <arg>/maven/*</arg>
+                            <arg>akka.cluster.bootstrap.demo.DemoApp</arg>
+                        </exec>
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/kubernetes-deployment/building-using-maven.md
```diff
@@ -60,8 +60,13 @@ The version number will only change when you create a new git commit. If you are
...
                     </tags>
                     <entryPoint>
-                       java $JAVA_OPTS -cp '/maven/*' akka.cluster.bootstrap.demo.DemoApp
-                    </entryPoint> 
+                        <exec>
+                            <arg>java</arg>
+                            <arg>-cp</arg>
+                            <arg>/maven/*</arg>
+                            <arg>akka.cluster.bootstrap.demo.DemoApp</arg>
+                        </exec>
```
octonato, line 68: what happened with the $JAVA_OPTS? 

Was that a conscious removal? If not, why is it gone?

octonato: hmm, I think it was useless, correct?

Since this is the docker entry point, it doesn't make sense to pass the java opts as a env var. Who would have filled the var?

octonato (approved):

docs/src/main/paradox/kubernetes-deployment/building-using-maven.md
patriknw: I had trouble adding $JAVA_OPTS as an arg. It thinks it is the main class.
That makes sense because it's not a `sh` any more.

Probably not important to include JAVA_OPTS since JAVA_TOOL_OPTIONS can be used.

Issue #850, opened 2021-02-12
Auto-start of AkkaManagement and Bootstrap


patriknw:
Some config to enable automatic start of management and bootstrap, replacing the two `start` lines of code.

ignasi35:
Done in #851

PR #851, opened 2021-02-12
Autostart cluster bootstrap if the extension is loaded through config

johanandren:
References #850

### DIFF
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -68,9 +68,9 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
 
   private[this] val _selfContactPointUri: Promise[Uri] = Promise()
 
-
   // autostart if the extension is loaded through the config extension list
-  private val autostart = system.settings.config.getStringList("akka.extensions").contains(classOf[ClusterBootstrap].getName)
+  private val autostart =
+    system.settings.config.getStringList("akka.extensions").contains(classOf[ClusterBootstrap].getName)
 
   if (autostart) {
     log.info("ClusterBootstrap loaded through 'akka.extensions' auto starting management and bootstrap.")
@@ -85,7 +85,7 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
         AkkaManagement(system).start().failed.foreach(autostartFailed)
         ClusterBootstrap(system).start()
       } catch {
-        case NonFatal(ex)=> autostartFailed(ex)
+        case NonFatal(ex) => autostartFailed(ex)
       }
     }
   }
--- a/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
+++ b/cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
@@ -5,11 +5,10 @@
 package akka.management.cluster.bootstrap
 
 import java.util.concurrent.atomic.AtomicReference
-
 import akka.AkkaVersion
-import scala.concurrent.{ Future, Promise, TimeoutException }
-import scala.concurrent.duration._
 
+import scala.concurrent.{Future, Promise, TimeoutException}
+import scala.concurrent.duration._
 import akka.actor.ActorSystem
 import akka.actor.ClassicActorSystemProvider
 import akka.actor.ExtendedActorSystem
@@ -18,12 +17,13 @@ import akka.actor.ExtensionId
 import akka.actor.ExtensionIdProvider
 import akka.annotation.InternalApi
 import akka.cluster.Cluster
-import akka.discovery.{ Discovery, ServiceDiscovery }
+import akka.discovery.{Discovery, ServiceDiscovery}
 import akka.event.Logging
 import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.server.Route
 import akka.management.cluster.bootstrap.contactpoint.HttpClusterBootstrapRoutes
 import akka.management.cluster.bootstrap.internal.BootstrapCoordinator
+import akka.management.scaladsl.AkkaManagement
 import akka.management.scaladsl.ManagementRouteProviderSettings
 import akka.management.scaladsl.ManagementRouteProvider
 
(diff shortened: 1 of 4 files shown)


cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -64,6 +64,26 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
 
   private[this] val _selfContactPointUri: Promise[Uri] = Promise()
 
+
+  private val autostart = system.settings.config.getStringList("akka.extensions")
+    .indexOf(classOf[ClusterBootstrap].getName) > -1
```
johanandren, line 69: A little bit hackish to look at the `akka.extension` list, but there's no public API way I can think of to identify when its loaded through config.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -64,6 +64,26 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
...
+    // Akka Management hosts the HTTP routes used by bootstrap
+    Future {
+      // we can't let it block extension init
+      try {
+        AkkaManagement(system).start()
+        ClusterBootstrap(system).start()
+      } catch {
+        case ex =>
+          log.error("Failed to autostart cluster bootstrap, terminating system", ex)
+          system.terminate()
```
johanandren, line 82: Actually a failure would probably be async, bind port failure for example, so we'd need some way to catch that and terminate.

cluster-bootstrap/src/test/scala/akka/management/cluster/bootstrap/contactpoint/ClusterBootstrapAutostartIntegrationSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+      info(s"System [$id]:   remoting port: $remotingPort")
+
+      contactPointPorts = contactPointPorts.updated(id, managementPort)
+      remotingPorts = remotingPorts.updated(id, remotingPort)
+
+      ConfigFactory.parseString(s"""
+        akka {
+          loglevel = INFO
+          # trigger autostart by loading the extension through config
+          extensions = ["akka.management.cluster.bootstrap.ClusterBootstrap"]
```
johanandren, line 45: This is what's needed to enable autoload on the user side

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
octonato: It may look a little less hackish if we use `contains`, no?

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
octonato: AkkaManagement returns a Future so we could already call the ClusterBootstrap in a `map`. 

I will assume that a port binding error i AkkaManagement would result in a failed future. 

For the Bootstrap, if it is a remote port issue, it will crash immediately, no? That would fail the Future

octonato (approved):
That will be a good addition. Simple and good.

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
johanandren: I think not mapping the two of them is fine, but we should capture the Future returned by `AkkaManagement#start()` and terminate system if it fails.`ClusterBootstrap#start()` sadly is `Unit` and starts some actor but not sure anything that should lead to termination can really go wrong there.

patriknw (commented):
looking good, this will be useful

cluster-bootstrap/src/main/scala/akka/management/cluster/bootstrap/ClusterBootstrap.scala
```diff
@@ -64,6 +67,28 @@ final class ClusterBootstrap(implicit system: ExtendedActorSystem) extends Exten
 
   private[this] val _selfContactPointUri: Promise[Uri] = Promise()
 
+
+  // autostart if the extension is loaded through the config extension list
+  private val autostart = system.settings.config.getStringList("akka.extensions").contains(classOf[ClusterBootstrap].getName)
```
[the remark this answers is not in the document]
johanandren: I'm thinking it should be extensions and added by user in application.conf, having it in library-extensions auto-added by library would be too magic for this IMO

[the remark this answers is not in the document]
patriknw: yes, I didn't mean to add it to any reference.conf. Just want to be sure that it can be added with `+=` by the operator in `main.conf` which includes `application.conf`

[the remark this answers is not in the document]
patriknw: and to be more precise
the user may have
```
akka.extensions = [foo]
```

and from main.conf we want to have

```
include application.conf
akka.extensions += [ClusterBootstrap]
```

I think that should work, but config has surprised me previously in this area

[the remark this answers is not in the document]
johanandren: Hmmm. I'd have to double check that, not sure what Config does when there is a `+=` for a key that is not defined yet.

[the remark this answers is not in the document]
johanandren: Verified and such an include+append works fine 👍 
```
conf.getStringList("akka.extensions")
val res1: java.util.List[String] = [foo, ClusterBootstrap]
```

ignasi35:
@johanandren I did the header upgrade on https://github.com/akka/akka-management/pull/855 so we can rebase this.

Alternatively, we can do a full `sbt test:headerCreate` since I think this PR doesn't update the headers of all modules (only 29/74 files modified).

ignasi35 (approved):

Issue #852, opened 2021-02-12
ClusterSingleton does not work with KubernetesLease


danischroeter:
since ClusterSingletonManager uses his self.path in the lease name.

ClusterSingletonManager
```
  private val singletonLeaseName = s"${context.system.name}-singleton-${self.path}"

  override val log = Logging.withMarker(context.system, this)

  val lease: Option[Lease] = settings.leaseSettings.map(
    settings =>
      LeaseProvider(context.system)
        .getLease(singletonLeaseName, settings.leaseImplementation, cluster.selfAddress.hostPort))
```
I guess this should be `settings.singletonName`
I can create a pull request if my assumption is true...

```
akka.actor.InvalidActorNameException: Invalid actor path element [kubernetesLease5-im-singleton-akka://im/user/identityManager-singleton-im@10.52.105.235:9999], illegal character [/] at position: 35. Actor paths MUST: not start with `$`, include only ASCII letters and can only contain these special characters: -_.*$+:@&=,!~';.
	at akka.actor.ActorPath$.validatePathElement(ActorPath.scala:100)
	at akka.actor.ActorPath$.validatePathElement(ActorPath.scala:78)
	at akka.actor.dungeon.Children.checkName(Children.scala:250)
	at akka.actor.dungeon.Children.attachChild(Children.scala:52)
	at akka.actor.dungeon.Children.attachChild$(Children.scala:51)
	at akka.actor.ActorCell.attachChild(ActorCell.scala:410)
	at akka.actor.ActorSystemImpl.systemActorOf(ActorSystem.scala:889)
	at akka.coordination.lease.kubernetes.KubernetesLease.<init>(KubernetesLease.scala:32)
	at akka.coordination.lease.kubernetes.KubernetesLease.<init>(KubernetesLease.scala:36)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
[...]

PR #853, opened 2021-02-12
Avoid using arbitrary string in lease actor name

johanandren:
References #852

Unsafe to assume lease name and owner are always valid strings to include in actor name.


johanandren:
Updated with sanitization of the lease name.

chbatey (approved):

Issue #848, opened 2021-02-08
[Failed] ClusterHttpManagementRoutesSpec


ignasi35:
https://travis-ci.com/github/akka/akka-management/jobs/480307808#L4064

```
[info]   - when calling GET /cluster/domain-events *** FAILED *** (838 milliseconds)
[info]     A timeout occurred waiting for a future to complete. Waited 150 milliseconds. (ClusterHttpManagementRoutesSpec.scala:449)
[info]     org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2:
```

patriknw:
150 millis looks like a small ScalaTest patience config.

@longshorej can you follow up here?

longshorej:
Ah - I'll look into this, should have some time over the next week.

ignasi35:
Again https://travis-ci.com/github/akka/akka-management/jobs/486017783#L6071

```
[info]   - when calling GET /cluster/domain-events *** FAILED *** (770 milliseconds)
[info]     A timeout occurred waiting for a future to complete. Waited 150 milliseconds. (ClusterHttpManagementRoutesSpec.scala:449)
[info]     org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2:
[info]     at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:339)
[info]     at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]     at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]     at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
[info]     at akka.cluster.http.management.scaladsl.ClusterHttpManagementRoutesSpec.$anonfun$new$34(ClusterHttpManagementRoutesSpec.scala:449)
[info]     at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]     at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]     at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]     at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]     at org.scalatest.Transformer.apply(Transformer.scala:20)
```

PR #860, opened 2021-02-25
Increase patience config

ignasi35:
A blind shot at fixing #848

### DIFF
--- a/cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
+++ b/cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
@@ -36,6 +36,7 @@ import akka.util.Timeout
 import com.typesafe.config.ConfigFactory
 import org.mockito.Matchers._
 import org.mockito.Mockito._
+import org.scalatest.concurrent.Eventually
 import org.scalatest.concurrent.PatienceConfiguration.{ Timeout => ScalatestTimeout }
 import org.scalatest.concurrent.ScalaFutures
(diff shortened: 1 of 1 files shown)


cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -400,6 +404,8 @@ class ClusterHttpManagementRoutesSpec
 
         import scala.concurrent.duration._
 
+        implicit val patience: PatienceConfig = PatienceConfig(timeout = Span(3, Seconds), interval = Span(50, Millis))
```
ignasi35, line 407: The flaky test in #848 doesn't seem to do anything fancier than any other test in that Spec so I suspect this PatienceConfig is not the fix.

cluster-http/src/test/scala/akka/cluster/http/management/scaladsl/ClusterHttpManagementRoutesSpec.scala
```diff
@@ -443,7 +449,7 @@ class ClusterHttpManagementRoutesSpec
             .futureValue(t)
         responseGetDomainEvents.status shouldEqual StatusCodes.OK
         val responseGetDomainEventsData = responseGetDomainEvents.entity.dataBytes
-          .takeWithin(100.millis)
+          .takeWithin(500.millis)
```
ignasi35, line 452: I suspect this is the actual fix since taking events during a longer period is likely to obtain the one we're looking for.

ignasi35:
Rebased over #861

chbatey (approved):

PR #863, opened 2021-02-26
Fix compilation errors on authors.scala

ignasi35:
fixed while releasing #862


ignasi35:
Fails because `1.0.10` artifacts are not in maven central yet.

octonato (approved):
👍

Issue #890, opened 2021-04-23
Ignore 'waiting' containers 


raboof:
### Short description

When using k8s API-based discovery, multiple containers per pods, and the Akka container is failing, it will also prevent other nodes from successfully forming a cluster.

### Details

When using Kubernetes API-based discovery, Cluster Bootstrap will only take into account pods that are in the `Running` phase.

When using multiple containers per pod, the pod goes from `Pending` to `Running` as soon as the first container is started (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/).

In that case, when the container that is running Akka is not running, it will prevent Cluster Bootstrap from forming a cluster, as the pod will be discovered but connecting to it will fail. When the container was still starting up, waiting for it to become available is a good, conservative choice.

However, when the container is not starting up successfully at all (for example due to a configuration error), and going in `CrashLoopBackOff`, it now also prevents other nodes (where the configuration error might be fixed) from forming a cluster.

We don't know which container in a pod is the Akka container. I think it would be reasonable to introduce a configuration option to specify the name of the Akka container, and to disregard pods where the container with that name is in the `waiting` state.

raboof:
Fixed by #891

PR #889, opened 2021-04-19
release to sonatype using GH Actions

octonato:
https://github.com/akka/akka-meta/issues/149

TODO
- [x] port integration tests
- [x] configure cron job
- [x] configure to run on release branch
- [ ] configure email notification
- [ ] configure required checks in settings page

PR #891, opened 2021-04-23
Filter pods where the Akka container has 'waiting' status

raboof:
We wait until pods become 'Running', but when using pods with multiple
containers, the pod becomes 'Running' as soon as the first primary
container starts.

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

In that case, when the Akka container fails to start, it also prevents
other pods (where the issue might be resolved) from starting. For that
reason, allow configuring which container is the Akka container, so we
can ignore it from discovery when it is 'waiting'.

Refs #890


raboof:
Travis failure is #892

patriknw (approved):
LGTM

raboof:
Remaining failures are due to the docker pull limit, which should be fixed with #889

PR #892, opened 2021-04-26
Use the latest jabba

raboof:
This uses the jabba script from master. This is needed because the
latest release, 0.11.2, has a problem.


raboof:
Not sure why GitHub is waiting for checks from #889 as that PR isn't merged yet.

PR #919, opened 2021-06-12
missing port name, it's necessary to pass Kubernetes Ready probe

franciscolopezsancho:

### DIFF
--- a/integration-test/kubernetes-api/kubernetes/akka-cluster.yml
+++ b/integration-test/kubernetes-api/kubernetes/akka-cluster.yml
@@ -41,6 +41,7 @@ spec:
         # akka-management bootstrap
         - containerPort: 8558
           protocol: TCP
+          name: management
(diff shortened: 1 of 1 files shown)


johanandren:
Hmm, but this test was passing already before, so the port name is not really necessary?

franciscolopezsancho:
OK. I did a test in local and didn't work for me (on my laptop : ) until I added it so I thought. But if you have tests that pass with this I'll close it, np. I didn't know about those.

franciscolopezsancho:
BTW when I said testing in local is not with a Spec but actually deploying in a Kubernetes Cluster.

johanandren:
Hmm, ok, well, the test passes with it as well, but would be good to understand why it would fail in one case but not in the integration tests.

franciscolopezsancho:
I'm looking into it. When I find what's the difference I'll open it again.

PR #921, opened 2021-07-01
Remove may-change marker on aws module

johanandren:
AWS module has been stable for quite a while and most if not all actual API is configuration,
by marking a few classed final/methods private I think we can remove the may change marker and
see it as stable.


RayRoestenburg (approved):
LGTM

Issue #926, opened 2021-07-08
System Materializer shutdown on stop


fredfp:
Akka-management shuts down the system materializer when stopping.

PR #927, opened 2021-07-08
Do not shutdown system materializer

fredfp:
References #926


raboof (approved):
That indeed looks like a bug!

patriknw (approved):
LGTM

Issue #920, opened 2021-06-16
Failed to unmarshal Kubernetes API response after update to Akka Management 1.1.0


thjaeckle:
### Versions used 

Akka version: 2.6.13
Akka Management version: 1.1.0

We are running Kubernetes 1.20 on managed Azure AKS.

### Expected Behavior

After updating from Akka Management 1.0.10 to 1.1.0 we would have expected to see no warnings in the logs.

### Actual Behavior

Actually, those WARNINGS are logged by `akka.discovery.kubernetes.KubernetesApiServiceDiscovery`:
```
Failed to unmarshal Kubernetes API response. 
Status code: [200 OK]; 
Response body: [HttpEntity.Strict(application/json,298268 bytes total)]. 
Ex: [Object is missing required member 'containerStatuses']

Resolve attempt failed! 
Cause: spray.json.DeserializationException: Object is missing required member 'containerStatuses'
```

The cluster formed itself, even with the WARNINGS.
Could it be that the `"containerStatuses"` is an optional field in the K8S API?

We don't use several containers per pod - as I understood the `"containerStatuses"` was added to support such a scenario.

DennisVDB:
We had the same issue at work. In our case it was due to a pod not starting up (ready 0/2), when akka-management queries that pod there would no `containerStatuses` field, decoding would fail as expects at least an empty list. This could be solved by defaulting to an empty list of container statuses.

thjaeckle:
Will the fix for this issue done in the merged PR #928 be released?

octonato:
v1.1.2 is released and #928 is included: https://github.com/akka/akka-management/commit/bad23b9508ea67d9e553caf2a1a274da82446826

PR #928, opened 2021-07-28
Ignore pending pods

DennisVDB:
References #920

Fixes the issue where pending pods (no `containerStatuses` field return by the k8s api) would fail the retrieval of the pod list. The json decoder always expected the `containerStatuses` field in the response.


octonato (approved):
LGTM

We will need some mima exclusions because of change in PodStatus (see https://github.com/akka/akka-management/pull/928/checks?check_run_id=3180624670#step:6:555)

Alternative would be to return `Seq.empty` if the field is not available. That would avoid the mima exclusion, but I guess it's more idiomatic to use `Option` for missing json fields.

octonato:
@DennisVDB, for the mima exclusion, you will need to add a file under src/main/mima-filters with the previous version in the name. See https://github.com/akka/akka-management/blob/master/lease-kubernetes/src/main/mima-filters/1.0.9.backwards.excludes as example.

raboof (approved):
sorry

thjaeckle:
@raboof will this be part of a 1.1.x bugfix release?

PR #929, opened 2021-07-29
Remove misleading check of k8s lease

andreaTP:


sebastian-alfers (approved):

docs/src/main/paradox/kubernetes-lease.md
```diff
@@ -113,11 +113,11 @@ spec:
 
 ```
 
-To get the current leases the fully qualified name must be uses as Kubernetes has since added a Lease type.
+@@@ note
 
-```
-kubectl get leases.akka.io
-```
```
octonato, line 120: What was exactly the issue? 
That calling `kubectl get leases.akka.io` would only return values during a split?

octonato (approved):
LGTM, but I have a question... :-)

docs/src/main/paradox/kubernetes-lease.md
andreaTP: correct

PR #930, opened 2021-07-30
Fix integration tests

andreaTP:
Use minikube action instead of a custom script.


andreaTP:
Currently integration tests are failing on master (e.g.: https://github.com/akka/akka-management/pull/929 )

raboof (approved):
balancing trusting that action against maintaining our own minikube script, OK I guess :)

andreaTP:
@raboof correct and the developer is a RedHat employee working on projects that we use (such as fabric8 kubernetes client)

PR #939, opened 2021-08-09
Don't auto-update jackson-datatype

raboof:
Instead of #932

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -8,6 +8,7 @@ updates.ignore = [
 updates.pin = [
   # To be updated in tandem with upstream Akka
   {groupId = "com.fasterxml.jackson.core", version = "2.10."}
+  {groupId = "com.fasterxml.jackson.datatype", version = "2.10."}
   {groupId = "org.scalatest", artifactId = "scalatest", version = "3.1."}
 ]
 
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,15 +1,18 @@
 pullRequests.frequency = "@monthly"
 
+updates.pin = [
+  # To be updated in tandem with upstream Akka
+  {groupId = "com.fasterxml.jackson.core", version = "2.11."}
+  {groupId = "org.scalatest", artifactId = "scalatest", version = "3.1."}
+]
+
 updates.ignore = [
   {groupId = "com.typesafe.akka"}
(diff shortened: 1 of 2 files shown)


.scala-steward.conf
```diff
@@ -8,6 +8,7 @@ updates.ignore = [
 updates.pin = [
   # To be updated in tandem with upstream Akka
   {groupId = "com.fasterxml.jackson.core", version = "2.10."}
+  {groupId = "com.fasterxml.jackson.datatype", version = "2.10."}
```
patriknw, line 11: Isn't `master` for Akka 2.6.x only? Then it should align the Jackson version (in Dependencies.scala) with what we have in Akka, which is 2.11.4

Also, see https://github.com/akka/akka/blob/master/.scala-steward.conf for pin and ignore

project/Dependencies.scala
```diff
@@ -18,19 +18,18 @@ object Dependencies {
...
-    "com.fasterxml.jackson.core" % "jackson-databind" % JacksonDatabindVersion
+    "com.fasterxml.jackson.core" % "jackson-databind" % JacksonVersion
   )
 
   private val JacksonDatatype = Seq(
     "com.fasterxml.jackson.datatype" % "jackson-datatype-guava" % JacksonVersion,
     "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8" % JacksonVersion,
     // Specifying guava dependency because older transitive dependency has security vulnerability
-    "com.google.guava" % "guava" % "27.1-jre"
+    "com.google.guava" % "guava" % "29.0-jre"
```
patriknw, line 32: Could we remove this now and rely on the transitive dependency instead?

raboof: I don't think so: https://repo1.maven.org/maven2/com/fasterxml/jackson/datatype/jackson-datatype-guava/2.11.4/jackson-datatype-guava-2.11.4.pom still specifies 0.20

patriknw: ok, hopefully switching from 27 to 29 doesn't break something else

patriknw (approved):
LGTM

project/Dependencies.scala
raboof: `com.orbitz.consul:consul-client:1.5.3` was already bringing in version 29.0-jre, so sbt users would already get that one. I updated this one here so maven users are in sync again.

Issue #947, opened 2021-09-07
SunCertPathBuilderException when using Kubernetes api discovery in OpenShift


khujo:
### Versions used

Akka version: 2.6.15  
Akka-management version: 1.1.1  
Akka-http version: 10.2.6

### Expected Behavior

We run an Akka cluster using Kubernetes API as discovery mechanism. Discovery is configured
something like this.

```
akka.discovery {
  kubernetes-api {
    pod-namespace = "some-namespace"
    pod-label-selector = "actorSystemName=someActorSystem"
  }
}
```

Discovery should work this way in a normal Kubernetes distribution, e.g. AKS, as well as in OpenShift. (It was working in OpenShift in 1.0.10)

### Actual Behavior

After upgrading to akka-management 1.1.1 the discovery throws an exception when running in OpenShift CodeReady Containers.

```
  sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
  	at java.base/sun.security.validator.PKIXValidator.doBuild(Unknown Source)
  	at java.base/sun.security.validator.PKIXValidator.engineValidate(Unknown Source)
  	at java.base/sun.security.validator.Validator.validate(Unknown Source)
  	at java.base/sun.security.ssl.X509TrustManagerImpl.validate(Unknown Source)
  	at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(Unknown Source)
  	at java.base/sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(Unknown Source)
  	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.checkServerCerts(Unknown Source)
  	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.onCertificate(Unknown Source)
  	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.consume(Unknown Source)
  	at java.base/sun.security.ssl.SSLHandshake.consume(Unknown Source)
  	at java.base/sun.security.ssl.HandshakeContext.dispatch(Unknown Source)
  	at java.base/sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(Unknown Source)
  	at java.base/sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(Unknown Source)
  	at java.base/java.security.AccessController.doPrivileged(Native Method)
[...]

PR #954, opened 2021-09-15
Load multiple ca certificates from file for kubernetes api discovery

khujo:
This PR will load multiple CA certificates from a file provided to the kubernetes api discovery. This is necessary because in case of OpenShift, the cluster will have multiple CAs and the ca.crt file will thus contain multiple certificates.

There are two things I would like to bring up for discussion.

1. Every certificate needs a unique key when added to the trust store. Before this PR the key was the atomic string "cacert". To create a unique key I now append a random string of 6 chars. I don't know if this is the best solution. 
2. The PemManagersProvider is copied from Akka-HTTP. So this PR will introduce a difference to the source in Akka-HTTP. Maybe this issue should also be fixed in Akka-HTTP.

References #947

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -110,8 +110,7 @@ lazy val `akka-management-pki` = project
   .settings(
     name := "akka-management-pki",
     libraryDependencies := Dependencies.ManagementPki,
-    // Don't enable mima until 1.1.1
-    mimaPreviousArtifacts := Set.empty
+    mimaPreviousArtifactsSet
   )
 
 lazy val `loglevels-logback` = project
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @cpqngbot,

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

octonato (commented):
Thanks @khujo 

The build will fail because of bin-compat issues. Please, check the MiMa output. You will need to add MiMa exclusions.

khujo:
Hi @octonato,
the MiMa output did not report any issues. However, it failed because mimaPreviousArtifact was not set for akka-management-pki after the 1.1.1 release. I pushed a new commit to set this value.

octonato (approved):
Thanks @khujo. 

You are right, we won't need MiMa because the `PemManagersProvider` object is private. All fine in that case.

jvdgeest:
I'm just curious why this PR has not been merged yet? I would really appreciate this fix.

octonato:
@jvdgeest, no special reason. We lost if from sight. I will merge it.

jvdgeest:
Thanks a lot @octonato!

PR #982, opened 2021-12-10
Swap order of directives in Cluster Bootstrap

zackman0010:
Swapping the order of the `get` and `path` directives in the Cluster Bootstrap routes so that a rejection will prefer 404 Not Found over 405 Method Not Allowed if both method and path are wrong.


johanandren (approved):
Sounds like it makes sense.

Issue #975, opened 2021-12-03
Unable to use akka-discovery-aws-api-async with multiple AWS SDK async http clients on the classpath


fgrutsch:
### Versions used 

Akka version: transitive dependency of akka-management
Akka management version: 1.1.1

### Expected Behavior

I should be able to use `akka-discovery-aws-api-async` even if I have more than one AWS SDK async HTTP client implementation on the classpath.

### Actual Behavior

Right now I am unable to use `akka-discovery-aws-api-async` if I have more than one AWS SDK HTTP client implementation on the classpath, because the `EcsAsyncClient` doesn't explicitly define an httpClient and therefore does a lookup on the classpath (see description [here](https://github.com/matsluni/aws-spi-akka-http#usage)). 

In my production application, I came across that issue because I am also using [alpakka sqs](https://doc.akka.io/docs/alpakka/current/sqs.html), which transitively depends on https://github.com/matsluni/aws-spi-akka-http, therefore pulling in another AWS SDK async HTTP client.

My workaround right now is to use the sync `akka-discovery-aws-api` discovery method, as in my production app I only use AWS SDK async clients and therefore don't have any other AWS SDK sync HTTP clients on the classpath.

### Relevant logs

```
[ERROR] [12/02/2021 17:50:42.326] [ecsBootstrapDemoApp-akka.actor.internal-dispatcher-2] [akka://ecsBootstrapDemoApp/system/bootstrapCoordinator] Multiple HTTP implementations were found on the classpath. To avoid non-deterministic loading implementations, please explicitly provide an HTTP client via the client builders, set the software.amazon.awssdk.http.async.service.impl system property with the FQCN of the HTTP service to use as the default, or remove all but one HTTP implementation from the classpath
software.amazon.awssdk.core.exception.SdkClientException: Multiple HTTP implementations were found on the classpath. To avoid non-deterministic loading implementations, please explicitly provide an HTTP client via the client builders, set the software.amazon.awssdk.http.async.service.impl system property with the FQCN of the HTTP service to use as the default, or remove all but one HTTP implementation from the classpath
        at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:98)
        at software.amazon.awssdk.core.internal.http.loader.ClasspathSdkHttpServiceProvider.loadService(ClasspathSdkHttpServiceProvider.java:62)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
[...]

PR #986, opened 2021-12-23
explicitly set httpClient in AsyncEcsServiceDiscovery

fgrutsch:
References #975


discovery-aws-api-async/src/main/scala/akka/discovery/awsapi/ecs/AsyncEcsServiceDiscovery.scala
```diff
@@ -43,7 +44,8 @@ class AsyncEcsServiceDiscovery(system: ActorSystem) extends ServiceDiscovery {
 
   private[this] lazy val ecsClient = {
     val conf = ClientOverrideConfiguration.builder().retryPolicy(RetryPolicy.none).build()
-    EcsAsyncClient.builder().overrideConfiguration(conf).build()
+    val httpClient = NettyNioAsyncHttpClient.create()
```
johanandren, line 47: Does this one also come in transitively, I don't see an explicit dependency on Netty?

patriknw: Isn't this in conflict with the description you linked to in the issue?
https://github.com/matsluni/aws-spi-akka-http#usage

Doesn't this always require the `netty-nio-client` but the linked description talks about excluding it.

Do we need some configuration to select client implementation?

fgrutsch: @johanandren Yes, both the [Apache Http Client](https://mvnrepository.com/artifact/software.amazon.awssdk/apache-client) (for sync clients) and the [Netty Nio Http Client](https://mvnrepository.com/artifact/software.amazon.awssdk/netty-nio-client) (for async clients) are transitive dependencies of the [ECS AWS SDK dependency](https://mvnrepository.com/artifact/software.amazon.awssdk/ecs) (this is true for all SDK services).

As this is the `AsyncEcsServiceDiscovery` we don't need the apache client, I added this exclusion to reduce the size of pulled-in transitive dependencies as it's not needed in this case.


@patriknw Yes if you mean this

> When you use this library and a specific AWS service (e.g. S3, SQS, etc...) you may want to exclude the transitive Netty dependency netty-nio-client like this ...

then it refers to https://github.com/matsluni/aws-spi-akka-http itself as this library is an [AWS Http Client SPI](https://mvnrepository.com/artifact/software.amazon.awssdk/http-client-spi) implementation. So you if you pull in any other AWS SDK dependency you most likely don't want the transitive HTTP clients (to reduce the size again).

In the end this means I just set the NIO client explicitly to avoid the runtime exception if there are multiple implementations of [AWS Http Client SPI](https://mvnrepository.com/artifact/software.amazon.awssdk/http-client-spi) on the classpath.

> Do we need some configuration to select client implementation?

This would be the preferable solution I assume. So if someone already has an instance of https://github.com/matsluni/aws-spi-akka-http in their project to use this as the underlying HTTP client of the ECS client. However, I am not sure how you would pass an instance to this discovery method, as everything seems to be initialized via extensions. Maybe you have some ideas?

patriknw: With configuration I mean a way to not use the NettyNioAsyncHttpClient and then use how it was done before, but maybe that doesn't make sense at all?

fgrutsch: Hmm, I understand what you mean now, but for my use case, it doesn't make too much sense as mostly the classpath exception is a problem that would be avoided with the current solution and be backward compatible.

Even though we would have the configuration and use the old approach, then if you would like to use https://github.com/matsluni/aws-spi-akka-http you are still stuck because the [default builder](https://github.com/matsluni/aws-spi-akka-http/blob/master/src/main/scala/com/github/matsluni/akkahttpspi/AkkaHttpClient.scala#L144) would spawn an actor system again and you run into startup problems (e.g. port 2551 already used) as its not using an existing system.

johanandren (approved):
LGMT, sorry for the long delay @fgrutsch

patriknw (approved):
LGTM

patriknw:
looks like all checks are green, just not reporting back

PR #1001, opened 2022-01-11
fix: ipv6 match host

WayneWang12:
host match will fail in ipv6 environment. Fix it by replace all `[` and `]`.


johanandren (approved):
LGTM

PR #1019, opened 2022-03-08
cluster-bootstrap: Add AKKA_CLUSTER_BOOTSTRAP_PORT_NAME env var

longshorej:
Adds env var `AKKA_CLUSTER_BOOTSTRAP_PORT_NAME` to specify port name
used in bootstrap.

To be used alongside the existing env var, `AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME`

### DIFF
--- a/cluster-bootstrap/src/main/resources/reference.conf
+++ b/cluster-bootstrap/src/main/resources/reference.conf
@@ -34,6 +34,7 @@ akka.management {
       # If set to "", `None` is passed to the discovery mechanism and
       # ${akka.management.http.port} is assumed.
       port-name = ""
+      port-name= ${?AKKA_CLUSTER_BOOTSTRAP_PORT_NAME}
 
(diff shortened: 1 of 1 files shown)


johanandren (approved):
LGTM

johanandren:
Noted, that makes sense and I agree. Went ahead and Ok:ed/merged this since we already had one for the `AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME`

longshorej:
Would it be possible to get a release soon? Is that something I can help do?

patriknw:
We could cut a release, but I don't see this as a critical issue that motivates a release in itself. Unless I'm mistaken you can add the same thing in you `application.conf` until it has been released?

longshorej:
You're right - not critical, we can mirror config until next release.

patriknw:
thanks

PR #1099, opened 2022-11-20
chore: renovate GH workflows

ennru:
Update to current versions of GH actions before Node 12 is discontinued.
* Replace `olafurpg/setup-scala` with `coursier/setup-action`
* Locking "privately owned" actions to their git hash 
* Current Fossa CLI install URL
* Explicit permissions (see https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)
* Upgrade Mockito to 4.9.0 to become JDK17 friendly

Refs https://github.com/akka/akka/pull/31730

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -8,13 +8,16 @@ on:
       - release-*
     tags-ignore: [ v.* ]
 
+permissions:
+  contents: read
+
 jobs:
   check-code-style:
     name: Check Code Style
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3.1.0
         with:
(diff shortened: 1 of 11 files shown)


ennru:
Mockito on JDK17 requires some module openers:

[error] Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @5a9a0ced

johanandren:
Looks like it is mockito wanting to define classes, so may be worth trying to bump mockito to a newer version that get along with JDK 17 better instead of opening up.

johanandren (approved):
LGTM

Issue #12, opened 2017-02-27
decouple "backend" from frontend (akka-http)


sebastianharko:
It would be nice to have a better decoupling before the "backend" logic and the akka-http frontend.  Some people may want to use a different web front-end (http4s, Spring MVC etc.) but get the same info about members etc. Like all problems in CS, this can be solved by another level of indirection. In this case, introducing a controller class ... 

EDIT: I offer to do a PR for this, if it's considered a decent idea

patriknw:
Is it really worth it? How many lines of code would it actually save? 

It's not strange that we are opinionated about using Akka HTTP for this utility.

ktoso:
I don't think we should spend the time on this, sorry @sebastianharko.
We're the Akka team, this is the Akka organisation, and it uses Akka HTTP to manage Akka Cluster...
It makes sense. 

If someone indeed from-the-field will really need this we can reconsider, but let's not complicate more than we need to.

PR #1113, opened 2023-02-22
feat: add pod deletion cost extension

efgpinto:
References https://github.com/akka/akka/issues/31383

Opening for visibility and feeback.

Missing:
- [x] integration tests
- [x] unit tests for the annotator actor
- [x] retry strategy on annotation failure
- docs in separate PR https://github.com/akka/akka-management/pull/1123


rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/CostStrategy.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ *  INTERNAL API
+ *  Defines a trait for calculating the cost of removing a member from the akka cluster,
+ *  given said member and the list of the members of the cluster from oldest to newest.
+ */
+@InternalApi private[akka] trait CostStrategy {
+  def costOf(member: Member, membersByAgeDesc: SortedSet[Member]): Option[Int]
+}
+
+object OlderCostsMore extends CostStrategy {
```
patriknw, line 21: internal api markers

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/CostStrategy.scala
```diff
@@ -0,0 +1,32 @@
...
+  def costOf(member: Member, membersByAgeDesc: SortedSet[Member]): Option[Int] = {
+    val nrMembersToAnnotate = 2
```
patriknw, line 23: config property, and let's use 3 as default

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCost.scala
```diff
@@ -0,0 +1,85 @@
...
+  private val settings = KubernetesSettings(system)
+  log.debug("Settings {}", settings)
+
+  private final val startStep = new AtomicReference[BootstrapStep](NotRunning)
+
+  def start(): Unit = {
+    if (settings.podName.isEmpty) {
+      log.warning(
+        "No configuration found to extract the pod name from. " +
+        "Be sure to provide the pod name with `akka.management.pod-deletion-cost.pod-name` " +
```
patriknw, line 40: wrong config path?

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCost.scala
```diff
@@ -0,0 +1,85 @@
...
+        "or by setting ENV variable `KUBERNETES_POD_NAME`.")
+    } else if (startStep.compareAndSet(NotRunning, Initializing)) {
+      val listener = system.systemActorOf(Props[PodDeletionCostAnnotator](), "podDeletionCostAnnotator")
+      Cluster(system).subscribe(listener, classOf[ClusterEvent.MemberUp])
+      Cluster(system).subscribe(listener, classOf[ClusterEvent.MemberRemoved])
```
patriknw, line 45: I think it's better to move these subscribe to the constructor of the actor so that it subscribes to itself `context.self`

patriknw: that would also work in case the actor is restarted

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,134 @@
...
+  def receive = onMessage(0, SortedSet.empty)
+
+  private def onMessage(deletionCost: Int, membersByAgeDesc: SortedSet[Member]): Receive = {
+    case ClusterEvent.MemberUp(m) =>
+      log.debug("Received MemberUp {}", m)
+      updateDeletionCost(deletionCost, membersByAgeDesc + m)
+    case ClusterEvent.MemberRemoved(m, _) =>
+      log.debug("Received MemberRemoved {}", m)
+      updateDeletionCost(deletionCost, membersByAgeDesc - m)
+    case _ =>
```
patriknw, line 91: There is an `initialStateMode` parameter to the subscribe method, but here you have used the default InitialStateAsSnapshot. Then you must also handle `akka.cluster.ClusterEvent.CurrentClusterState`, and populate the initial membersByAgeDesc from that.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,134 @@
...
+
+      val request = ApiRequests.podDeletionCost(k8sSettings, apiToken, podNamespace, newCost)
+
+      val responseFuture: Future[HttpResponse] =
+        if (k8sSettings.secure)
+          http.singleRequest(request, clientSslContext)
+        else
+          http.singleRequest(request)
+
+      val res = Await.result(responseFuture, 2.seconds)
```
patriknw, line 122: Don't use blocking here, you can pipe the result of the future back to `self` https://doc.akka.io/docs/akka/current/actors.html#ask-send-and-receive-future

I think it can become the new state immediately without waiting for the response of the kubernetes request.

If the kubernetes request fails you can try again, after a scheduled backoff delay

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,134 @@
...
+import scala.collection.immutable.SortedSet
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.Await
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
+@InternalApi
+final class PodDeletionCostAnnotator extends Actor with ActorLogging {
+  private val cluster = Cluster(context.system)
+  private val k8sSettings = KubernetesSettings(context.system)
```
patriknw, line 37: An alternative would be to pass in KubernetesSettings as a constructor parameter. Then no need to make KubernetesSettings an extension since it's anyway only created once by the other extension.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,134 @@
...
+  private val http = Http()(context.system)
+  private lazy val apiToken = readConfigVarFromFilesystem(k8sSettings.apiTokenPath, "api-token").getOrElse("")
+  private val podNamespace = k8sSettings.podNamespace
+    .orElse(readConfigVarFromFilesystem(k8sSettings.podNamespacePath, "namespace"))
+    .getOrElse("default")
+
+  /**
+   * This uses blocking IO, and so should only be used to read configuration at startup.
+   */
+  private def readConfigVarFromFilesystem(path: String, name: String): Option[String] = {
```
patriknw, line 48: I think this should be passed in as a constructor parameter instead. The blocking IO isn't that nice. We can move this reading to the PodDeletionCost extension and run it from a Future with Akka's default `blocking-dispatcher`. When it has read the file it can start the actor. No need to start it immediately.

patriknw (commented):
looking good

rolling-update-kubernetes/src/main/resources/reference.conf
```diff
@@ -0,0 +1,35 @@
...
+
+    # Port for the Kubernetes API server. Typically this will be set via an environment
+    # variable that is set when running inside Kubernetes
+    api-service-port = 8080
+    api-service-port = ${?KUBERNETES_SERVICE_PORT}
+
+    # Namespace file path. The namespace is to create the lock in. Can be overridden by "namespace"
+    #
+    # If this path doesn't exist, the namespace will default to "default".
+    namespace-path = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```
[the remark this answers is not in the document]
efgpinto: yeah, the problem is that for the lease it's like this, see [here](https://github.com/akka/akka-management/blob/main/lease-kubernetes/src/main/resources/reference.conf#L21). basically there's a differnet setup between the lease and the discovery api modules. happy to follow the one you find better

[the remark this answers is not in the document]
patriknw: ok, let's use what you have then

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCost.scala
```diff
@@ -0,0 +1,117 @@
...
+  def start(): Unit = {
+    if (k8sSettings.podName.isEmpty) {
+      log.warning(
+        "No configuration found to extract the pod name from. " +
+        s"Be sure to provide the pod name with `$configPath.pod-name` " +
+        "or by setting ENV variable `KUBERNETES_POD_NAME`.")
+    } else if (startStep.compareAndSet(NotRunning, Initializing)) {
+      log.debug("Starting PodDeletionCost for podName={} with settings={}", k8sSettings.podName, costSettings)
+
+      implicit val blockingDispatcher: MessageDispatcher = system.dispatchers.lookup("blocking-io-dispatcher")
```
johanandren, line 47:
```suggestion
      implicit val blockingDispatcher: ExecutionContext = system.dispatchers.lookup("blocking-io-dispatcher")
```
Is what we need it for here

johanandren (commented):
Looks good, only one nitpick. Some form of integration tests would be good though.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/CostStrategy.scala
efgpinto: thanks, done in https://github.com/akka/akka-management/pull/1113/commits/e037463e5789c85bc091750b6e3590800a74c968

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCost.scala
efgpinto: tks, done in https://github.com/akka/akka-management/pull/1113/commits/e037463e5789c85bc091750b6e3590800a74c968

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
efgpinto: Thanks, done in https://github.com/akka/akka-management/pull/1113/commits/e037463e5789c85bc091750b6e3590800a74c968

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
efgpinto: using the `blocking-io-dispatcher` now as suggested. tks

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,193 @@
...
+      log.error(
+        "There was a client error when trying to set pod-deletion-cost annotation. " +
+        "Not retrying, check configuration. Error: {}",
+        er)
+
+    case RetryAnnotate =>
+      costRequestToResult(deletionCost).pipeTo(self)
+
+    case es => log.debug("Ignoring message {}", es)
+  }
```
johanandren, line 118: If a call fails, we schedule a retry in N seconds, but if before that backoff completes, cluster topology changes, that will trigger a new update, before the retry arrives.

Should we perhaps keep a revision number, increase on each topology update to be able to discard retries if they were for a different revision of the cluster?

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,193 @@
...
+    if (settings.secure)
+      http.singleRequest(request, clientSslContext)
+    else
+      http.singleRequest(request)
+  }.map {
+      case HttpResponse(status, _, e, _) if status.isSuccess() =>
+        e.discardBytes()(Materializer(context)); PodAnnotated()
+      case HttpResponse(s @ ClientError(_), _, _, _) => GiveUp(s.toString())
+      case HttpResponse(status, _, _, _)             => ScheduleRetry(s"Request failed with status=$status")
+      case _                                         => GiveUp("Unknown")
```
johanandren, line 137: Is this even reachable? The above case should capture all. 

I think you should make sure to discard bodies for the non success status cases as well, in case k8 emits some error details (then we'd be clogging up that connection pool).

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,193 @@
...
+    keyStore.load(null)
+    factory.init(keyStore, Array.empty)
+    val km: Array[KeyManager] = factory.getKeyManagers
+    val tm: Array[TrustManager] =
+      PemManagersProvider.buildTrustManagers(certificates)
+    val random: SecureRandom = new SecureRandom
+    val sslContext = SSLContext.getInstance("TLSv1.2")
+    sslContext.init(km, tm, random)
+    sslContext
+  }
```
johanandren, line 81: Perhaps better to do this up front to fail early setup is broken rather than on first access? It is currently anyway always touched by the val above.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,193 @@
...
+  }.map {
+      case HttpResponse(status, _, e, _) if status.isSuccess() =>
+        e.discardBytes()(Materializer(context)); PodAnnotated()
+      case HttpResponse(s @ ClientError(_), _, _, _) => GiveUp(s.toString())
+      case HttpResponse(status, _, _, _)             => ScheduleRetry(s"Request failed with status=$status")
+      case _                                         => GiveUp("Unknown")
+    }
+    .recover {
+      case NonFatal(e) => ScheduleRetry(e.getMessage)
+    }
```
johanandren, line 141: This looks good but one good pattern for future composition in actors is to pull it out to the companion to completely make sure we don't accidentally touch mutable state/actor context from the callbacks.

johanandren (commented):
A bit more detailed review of the actor.

rolling-update-kubernetes/src/main/resources/reference.conf
```diff
@@ -0,0 +1,35 @@
...
+
+    # Namespace to query for pods.
+    #
+    # Set this value to a specific string to override discovering the namespace using namespace-path.
+    namespace = "<namespace>"
```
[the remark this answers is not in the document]
efgpinto: Made it possible to set with ENV var as welll now https://github.com/akka/akka-management/pull/1113/commits/5f1a77475aea2e9cab7b29c6cc92aa1c0eba2920

build.sbt
```diff
@@ -154,6 +156,16 @@ lazy val `cluster-bootstrap` = project
...
+lazy val `rolling-update-kubernetes` = project
+  .in(file("rolling-update-kubernetes"))
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(
+    name := "akka-rolling-update-kubernetes",
+    libraryDependencies := Dependencies.RollingUpdateKubernetes,
+    mimaPreviousArtifactsSet
+  )
+  .dependsOn(`akka-management`, `akka-management-pki`, `cluster-bootstrap`)
+
```
efgpinto, line 168: I was wondering if maybe this could be only `rolling-update`.. not sure if we need to tie it to kubernetes on the depedency name. In the sense that maybe other things related with rolling update in general can be fit here in the future? kubernetes specific things are under a specific `package` anyway. just opening up the discussion...

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,221 @@
...
+    val sslContext = SSLContext.getInstance("TLSv1.2")
+    sslContext.init(km, tm, random)
+    sslContext
+  }
+  private val clientSslContext: Option[HttpsConnectionContext] = if (settings.secure) Some(ConnectionContext.httpsClient(sslContext)) else None
+
+  implicit val dispatcher: ExecutionContextExecutor = context.system.dispatcher
+  def receive = onMessage(0, SortedSet.empty(Member.ageOrdering))
+
+  private def onMessage(deletionCost: Int, membersByAgeDesc: SortedSet[Member]): Receive = {
```
johanandren, line 86: Maybe give this a more clear name to hint about the state machine, I think we often use `idle` for such cases.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,221 @@
...
+@InternalApi private[kubernetes] object PodDeletionCostAnnotator {
+  case object RetryTimerId
+  case object RetryAnnotate
+  sealed trait RequestResult
+  case object PodAnnotated extends RequestResult
+  case class ScheduleRetry(cause: String) extends RequestResult
+  case class GiveUp(cause: String) extends RequestResult
+
+  private def toResult(futResponse: Future[HttpResponse])(implicit context: ActorContext): Future[RequestResult] = {
+    implicit val mat: Materializer = Materializer(context)
```
johanandren, line 205: This is a resource leak, the materialiser has a lifecycle and each call to `toResult` creates a new one that leaks.

Easiest is to instead take an implicit system paramameter, instead of context, that way the system materialiser is implicitly available.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,221 @@
...
+      log.error(
+        "There was a client error when trying to set pod-deletion-cost annotation. " +
+        "Not retrying, check configuration. Error: {}",
+        er)
+
+    case es => log.debug("Ignoring message {}", es)
+  }
+
+  private def underRetryBackoff(deletionCost: Int, membersByAgeDesc: SortedSet[Member]): Receive = {
+    case cs@ClusterEvent.CurrentClusterState(members, _, _, _, _) =>
```
patriknw, line 119: I don't think CurrentClusterState can happen here. It will be the first message, before any MemberUp or MemberRemoved. It's only sent once.

integration-test/rollingupdate-kubernetes/src/main/resources/application.conf
```diff
@@ -0,0 +1,33 @@
...
+
+#FIXME how to avoid having this explicitly?
+blocking-io-dispatcher {
+  type = Dispatcher
+  executor = "thread-pool-executor"
+  thread-pool-executor {
+    fixed-pool-size = 32
+  }
+  throughput = 1
+}
```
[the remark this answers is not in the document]
johanandren: There is a blocking dispatcher defined for core Akka, maybe just use that? `akka.actor.default-blocking-io-dispatcher`/ constant `akka.dispatch.Dispatchers.DefaultBlockingDispatcherId`

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,221 @@
...
+      log.debug("Received while on retry backoff CurrentClusterState {}", cs)
+      context.become(underRetryBackoff(deletionCost, membersByAgeDesc ++ members)) // ordering used is from the first operand (so, by age)
+
+    case ClusterEvent.MemberUp(m) =>
+      log.debug("Received while on retry backoff MemberUp {}", m)
+      context.become(underRetryBackoff(deletionCost, membersByAgeDesc + m))
```
patriknw, line 125: good, this member will also be included in the next update from RetryAnnotate

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,221 @@
...
+    log.debug(
+      "Calculated cost={} (previously {}) for member={} in members by age (desc): {}",
+      newCost,
+      existingCost,
+      cluster.selfMember,
+      membersByAgeDesc)
+
+    if (newCost != existingCost) {
+      costRequestToResult(newCost).pipeTo(self)
+      context.become(onMessage(newCost, membersByAgeDesc))
```
patriknw, line 166: I haven't checked in detail. Does it still handle several subsequent failures and continue retrying until successful? Thinking about newCost vs existingCost for the failure retry scenario

rolling-update-kubernetes/src/test/resources/reference.conf
patriknw, line 1: remove this empty file?

patriknw (commented):
looking good

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCost.scala
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (C) 2017-2021 Lightbend Inc. <https://www.lightbend.com>
```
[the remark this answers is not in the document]
johanandren: We haven't bumped the currentYear in this project for a long while, leave it like this and I'll bump before we release.

efgpinto (commented):
Another round complete with unit tests for the actor and incorporated latest feedback.

project/project-info.conf
```diff
@@ -122,4 +122,14 @@ project-info {
       }
     ]
   }
+  rolling-update-kubernetes: ${project-info.shared-info} {
+    title: "Akka Rolling Update Kubernetes"
+    levels: [
+      {
+        readiness: Supported
+        since: "2023-03-01" #FIXME
+        since-version: "1.2.0" #FIXME
```
efgpinto, line 131: Still not sure what to do here...

efgpinto: I suppose the new version will be 1.2.1? or 1.3.0?
And support date probably from april? 2023-04-01 ?

patriknw: Yes, it will be a minor bump, 1.3.0, because we changed some dependencies with minor bump (Akka HTTP for Scala 3).

2023-04-01 sounds good, not important to set that exactly right to the release date

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
johanandren: Fair enough!

johanandren (commented):
Looking good

rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotatorSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+    "apiToken",
+    namespace,
+    PodDeletionCostSettings(system.settings.config.getConfig("akka.rollingupdate.kubernetes"))
+  )
+
+  override implicit val patienceConfig =
+    PatienceConfig(timeout = Span(5, Seconds), interval = Span(100, Millis))
+
+  override protected def afterAll(): Unit = {
+    TestKit.shutdownActorSystem(system)
```
patriknw, line 95: there is a shortcut for this, just `shutdown()` (in super TestKit)

rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotatorSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+
+  private val annotatorActorProps = Props(
+    classOf[PodDeletionCostAnnotator],
+    settings,
+    "apiToken",
+    namespace,
+    PodDeletionCostSettings(system.settings.config.getConfig("akka.rollingupdate.kubernetes"))
+  )
+
+  override implicit val patienceConfig =
```
patriknw, line 91: probably need explicit type, otherwise Scala 3 will complain

rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotatorSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+    """)
+      ))
+    with ImplicitSender
+    with AnyWordSpecLike
+    with Matchers
+    with BeforeAndAfterAll
+    with BeforeAndAfterEach
+    with Eventually {
+
+  private val wireMockServer = new WireMockServer(wireMockConfig().port(0))
```
patriknw, line 66: nice to be able to fake the api server

patriknw (commented):
good test

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotator.scala
```diff
@@ -0,0 +1,211 @@
...
+      log.debug("Received while on retry backoff MemberRemoved {}", m)
+      context.become(underRetryBackoff(membersByAgeDesc - m))
+
+    case RetryAnnotate =>
+      updateIfNewCost(Int.MinValue, membersByAgeDesc)
+
+    case es => log.debug("Under retry backoff, ignoring message {}", es)
+  }
+
+  private def updateIfNewCost(existingCost: Int, membersByAgeDesc: immutable.SortedSet[Member]): Unit = {
```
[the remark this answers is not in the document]
efgpinto: Making sure I got your idea right.. 

Currently, it logs an error every time it fails to update the annotation and schedules a retry.. we could downgrade the severity to warning after say 3 atttempts and also not log it on every attempt but maybe every 10th retry attempt for instance? This is would avoid clogging the logs with the error too much..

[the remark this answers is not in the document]
johanandren: We talked but to get the clarification in print: the idea was to log at debug or info first N tries as that is just resilience/retrying, but after N tries it might be a sign of misconfiguration, so start log at warn. This is probably not critical enough to ever be an error though the deployed app will work even if this would not work.

.github/workflows/integration-tests-rollingupdate.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - release-*
+    tags-ignore: [ v.* ]
+  schedule:
+    - cron: '0 2 * * *'  # every day 2am
```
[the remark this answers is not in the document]
efgpinto: tks, added in line `#12`

rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/PodDeletionCostAnnotatorSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+      underTest ! dummyNewMember
+      underTest ! dummyNewMember
+      underTest ! dummyNewMember
+
+      // no other interactions should have occurred while on backoff regardless of updates to the cluster
+      verify(1, patchRequestedFor(podK8sPath))
+
+      assertState(scenarioName, "OK")
+      verify(2, patchRequestedFor(podK8sPath))
+    }
```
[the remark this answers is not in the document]
efgpinto: added now, see lines 185-200

efgpinto (commented):
Updated with latest feedback. Docs have a different PR https://github.com/akka/akka-management/pull/1123 so marking this as ready for review as well.

johanandren (approved):
LGTM!

patriknw (approved):
LGTM

PR #1112, opened 2023-02-02
feat: Cross build for Scala 3

patriknw:

PR #1122, opened 2023-02-27
fix: Make api doc generation pass with Scala 3

johanandren:
Follow up to #1112

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,6 +2,9 @@ import com.typesafe.sbt.packager.docker.{ Cmd, ExecCmd }
 import sbt.Keys.parallelExecution
 
 ThisBuild / resolvers += Resolver.jcenterRepo
+Global / excludeLintKeys += autoAPIMappings
+Global / excludeLintKeys += projectInfoVersion
+Global / excludeLintKeys += previewPath
 
 // root
 lazy val `akka-management-root` = project
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -51,10 +51,10 @@ object Common extends AutoPlugin {
         else
           scalacOptionsBase
       },
-      javacOptions ++= Seq(
+      Compile / javacOptions ++= Seq(
           "-Xlint:unchecked"
         ),
-      javacOptions ++= (
+      Compile / javacOptions ++= (
           if (isJdk8) Seq.empty
           else Seq("--release", "8")
         ),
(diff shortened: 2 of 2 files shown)


project/Common.scala
```diff
@@ -51,25 +51,27 @@ object Common extends AutoPlugin {
...
-      javacOptions ++= (
+      Compile / javacOptions ++= (
           if (isJdk8) Seq.empty
           else Seq("--release", "8")
         ),
       scalacOptions ++= (
           if (isJdk8 || scalaVersion.value == Dependencies.Scala212) Seq.empty
           else Seq("--release", "8")
         ),
+      Compile / doc / javacOptions := Nil,
```
johanandren, line 65: Not sure why this one was failing as well, but it was on my machine (javadoc on JDK 17 issue?). Anyway, no need for Xlint during javadoc generation.

project/Common.scala
```diff
@@ -51,25 +51,27 @@ object Common extends AutoPlugin {
...
           "-doc-version",
-          version.value,
-          "-skip-packages",
-          "akka.pattern" // for some reason Scaladoc creates this
-        ),
+          version.value
+        ) ++
+        // for some reason Scaladoc creates this
+        (if (scalaVersion.value.startsWith("3")) Seq.empty
+         else Seq("-skip-packages", "akka.pattern")),
```
johanandren, line 74: This was the problem, flag doesn't seem to exist with Scala 3 (or works differently) and would stop building artifacts for publishing.

build.sbt
```diff
@@ -2,6 +2,9 @@ import com.typesafe.sbt.packager.docker.{ Cmd, ExecCmd }
 import sbt.Keys.parallelExecution
 
 ThisBuild / resolvers += Resolver.jcenterRepo
+Global / excludeLintKeys += autoAPIMappings
+Global / excludeLintKeys += projectInfoVersion
+Global / excludeLintKeys += previewPath
```
johanandren, line 7: Silencing annoying noise from sbt on start

project/Common.scala
```diff
@@ -62,14 +62,16 @@ object Common extends AutoPlugin {
           if (isJdk8 || scalaVersion.value == Dependencies.Scala212) Seq.empty
           else Seq("--release", "8")
         ),
+      doc / javacOptions := Nil,
```
johanandren, line 65: Not sure why this was needed, failing on my machine but not CI, I don't think we need that -X:lint when compiling apidocs anwyay.

efgpinto (approved):
Makes sense to me!

PR #1134, opened 2023-04-19
feat: Write pod cost to CRD

patriknw:
* alternative to adding the Kubernetes annotation directly to the pod resource in case allowing PATCH verb is a security concern
* a separate Kubernetes operator with more privilages would watch the CR and update the pod resource with the annotation

TODO:
- [x] add test similar to PodDeletionCostAnnotatorSpec
- [x] add integration test to integration-test/rollingupdate-kubernetes
- [x] docs
- [x] FIXMEs
- [x] maybe some more config for the CRD to be able to change the names


rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
```diff
@@ -0,0 +1,321 @@
...
+    apiToken: String,
+    clientHttpsConnectionContext: Option[HttpsConnectionContext])
+    extends KubernetesApi
+    with KubernetesJsonSupport {
+
+  import system.dispatcher
+
+  private implicit val sys: ActorSystem = system
+  // FIXME the asInstanceOf is because Scala3 complains
+  private val log = Logging(system, getClass.asInstanceOf[Class[Any]])
```
johanandren, line 46: In Akka we do `Logging(system, classOf[ThisClass])` to make Scala 3 happy

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
```diff
@@ -0,0 +1,321 @@
...
+            Future.successful(found)
+          case None =>
+            log.info("PodCost {} does not exist, creating", crName)
+            createPodCostResource(crName).flatMap {
+              case Some(created) => Future.successful(created)
+              case None =>
+                if (tries < maxTries) loop(tries + 1)
+                else Future.failed(new PodCostException(s"Unable to create or read PodCost after $maxTries tries"))
+            }
+        }
```
johanandren, line 117: Could be the yield block instead and return/throw directly instead of wrapping all return values in successful/failed

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
```diff
@@ -0,0 +1,321 @@
...
+              Right(toPodCostResource(updatedCr))
+            })
+        case StatusCodes.Conflict =>
+          getPodCostResource(crName).flatMap {
+            case None =>
+              Future.failed(new PodCostException(s"GET after PUT conflict did not return a PodCost [$crName]"))
+            case Some(cr) =>
+              log.debug("PodCostResource read after conflict: {}", cr)
+              Future.successful(Left(cr))
+          }
```
johanandren, line 171: Just map and throw/return instead of flatMap?

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
```diff
@@ -0,0 +1,321 @@
...
+        case StatusCodes.Unauthorized =>
+          handleUnauthorized(response)
+        case unexpected =>
+          Unmarshal(response.entity)
+            .to[String]
+            .flatMap(body => {
+              Future.failed(
+                new PodCostException(
+                  s"PUT for PodCost [$crName] returned unexpected status code $unexpected. Body: $body"))
+            })
```
johanandren, line 181: map + throw?

johanandren (commented):
Looks good so far

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
patriknw: ah, that seems obvious and better. I recognize this hack and I could have been guilty of introducing it at the first place and then it has been copied to many places. @sebastian-alfers would you be able to look for this pattern in the various repos and PR Johan's suggestion?

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
patriknw: @johanandren I incorporated some of your suggestions in aa4a6e81, but not sure I understand how to change this to yield and not flatMap. Note that it calls createPodCostResource and loop (recursively) and those return Future

johanandren: I think I simply missed the async-recursive loop.

johanandren (approved):
LGTM (but CI unhappy about some formatting and rolling upgrade integration test is failing)

efgpinto (approved):
LGTM

PR #1142, opened 2023-04-28
chore: Blocking IO at startup

patriknw:
Run blocking on DefaultBlockingDispatcherId, similar to the rolling updates initialisation.


johanandren (approved):
LGTM

Issue #966, opened 2021-10-20
@Scala and @Java tags don't seem to work in the documentation


dragisak:
[This](https://github.com/akka/akka-management/blob/v1.1.1/docs/src/main/paradox/bootstrap/index.md?plain=1#L103) doesn't seem to work

<img width="524" alt="image" src="https://user-images.githubusercontent.com/800802/138134542-828269b4-6cb6-42f5-9779-a323de0a9bd2.png">

Tried on both Firefox and Chrome

PR #1149, opened 2023-05-10
doc: Scala/Java toggle not working #966

johanandren:
Is case sensitive lowercase


johanandren:
Tried to make the group directive case insensitive in paradox but gave up.

johanandren:
Yeah, I figured always lc:ing it rather than silently eating would have been nice. Also didn't figure how to make it fail with an error for an unknown group.

PR #1154, opened 2023-05-19
doc: Correction of rbac api group for AppVersionRevision

patriknw:
* pods is in core "" api group
* replicasets is in "apps" api group

### DIFF
--- a/integration-test/rollingupdate-kubernetes/kubernetes/akka-cluster-app-value-revision.yml
+++ b/integration-test/rollingupdate-kubernetes/kubernetes/akka-cluster-app-value-revision.yml
@@ -65,8 +65,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: pod-reader
 rules:
-- apiGroups: ["apps", ""] # additional access to api to get revision
-  resources: ["pods", "replicasets"]
+- apiGroups: [""]
+  resources: ["pods"] # access to get replicaset
+  verbs: ["get", "list"]
+- apiGroups: ["apps"]
+  resources: ["replicasets"] # access to get revision annotation
   verbs: ["get", "list"]
 ---
 kind: RoleBinding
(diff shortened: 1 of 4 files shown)


integration-test/rollingupdate-kubernetes/kubernetes/akka-cluster-app-value-revision.yml
```diff
@@ -65,8 +65,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: pod-reader
 rules:
-- apiGroups: ["apps", ""] # additional access to api to get revision
-  resources: ["pods", "replicasets"]
+- apiGroups: [""]
+  resources: ["pods"] # access to get replicaset
+  verbs: ["get", "list"]
+- apiGroups: ["apps"]
+  resources: ["replicasets"] # access to get revision annotation
```
patriknw, line 72: it works to define it as `["apps", ""]` but it's rather confusing

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala
```diff
@@ -296,7 +296,7 @@ PUTs must contain resourceVersions. Response:
     Unmarshal(response.entity)
       .to[String]
       .map(body =>
-        throw new PodCostException(
+        throw new UnauthorizedException(
```
patriknw, line 299: doesn't matter what exception is used here, but since `handleUnauthorized` is also used for the revision

integration-test/rollingupdate-kubernetes/kubernetes/akka-cluster-app-value-revision.yml
efgpinto: indeed, better now

efgpinto (approved):
LGTM

PR #1155, opened 2023-05-24
fix: ensure versionPromise always completes

octonato:
Noticed today with @franciscolopezsancho, that the versionPromise may never complete forcing users to do some weird future gymnastics to work it around.

### DIFF
--- a/rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/AppVersionRevision.scala
+++ b/rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/AppVersionRevision.scala
@@ -34,21 +34,24 @@ final class AppVersionRevision(implicit system: ExtendedActorSystem) extends Ext
   def getRevision(): Future[Version] = versionPromise.future
 
   def start(): Unit = {
-    if (k8sSettings.podName.isEmpty) {
-      log.error(
-        "Not able to read the app version from the revision of the current ReplicaSet. Reason:" +
-        "No configuration found to extract the pod name from. " +
-        s"Be sure to provide the pod name with `$configPath.pod-name` " +
-        "or by setting ENV variable `KUBERNETES_POD_NAME`.")
-    } else {
-      if (isInitialized.compareAndSet(false, true)) {
+    if (isInitialized.compareAndSet(false, true)) {
+      if (k8sSettings.podName.isEmpty) {
+        val msg = "Not able to read the app version from the revision of the current ReplicaSet. Reason:" +
+          "No configuration found to extract the pod name from. " +
+          s"Be sure to provide the pod name with `$configPath.pod-name` " +
+          "or by setting ENV variable `KUBERNETES_POD_NAME`."
+        log.error(msg)
+        versionPromise.failure(new IllegalStateException(msg))
+      } else {
         Cluster(system).setAppVersionLater(getRevision())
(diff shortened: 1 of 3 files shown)


rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/AppVersionRevisionSpec.scala
```diff
@@ -5,34 +5,12 @@
...
 import org.scalatest.matchers.should.Matchers
-import org.scalatest.time.Millis
-import org.scalatest.time.Seconds
-import org.scalatest.time.Span
 import org.scalatest.wordspec.AnyWordSpecLike
-import org.scalatest.BeforeAndAfterAll
-import org.scalatest.BeforeAndAfterEach
-
-import scala.concurrent.duration._
-
```
octonato, line 35: code here moved to `KubernetesApiSpec` because it's only testing `KubernetesApi`. 

`AppVersionRevisionSpec` is not testing the failure case inside `AppVersionRevision.start`

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/AppVersionRevision.scala
```diff
@@ -34,21 +34,24 @@ final class AppVersionRevision(implicit system: ExtendedActorSystem) extends Ext
...
-    } else {
-      if (isInitialized.compareAndSet(false, true)) {
+    if (isInitialized.compareAndSet(false, true)) {
+      if (k8sSettings.podName.isEmpty) {
+        val msg = "Not able to read the app version from the revision of the current ReplicaSet. Reason:" +
+          "No configuration found to extract the pod name from. " +
+          s"Be sure to provide the pod name with `$configPath.pod-name` " +
+          "or by setting ENV variable `KUBERNETES_POD_NAME`."
+        log.error(msg)
+        versionPromise.failure(new IllegalStateException(msg))
```
octonato, line 44: If podname is not set, there is nothing else we can do, but if we try to later use `getRevision` we get a Future that never completes.

patriknw (commented):
good but question about the test

rolling-update-kubernetes/src/test/scala/akka/rollingupdate/kubernetes/AppVersionRevisionSpec.scala
```diff
@@ -45,188 +23,36 @@ object AppVersionRevisionSpec {
...
+      }
     """)
 }
-
 class AppVersionRevisionSpec
     extends TestKit(
       ActorSystem(
         "AppVersionRevisionSpec",
-        AppVersionRevisionSpec.config
+        KubernetesApiSpec.config
```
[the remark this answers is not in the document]
octonato: Fixed. 

It turns out that tests were passing even without using the right config. Which makes sense because podname defaults to empty in the `reference.conf`. 

Anyway, I keep the new config just to make it very explicit, but did some clean-up as well.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/AppVersionRevision.scala
octonato: Hmm, maybe we should not return an `IllegalStateException`, because it's not illegal. It's just missing info causing a degraded state.

patriknw (approved):
LGTM

PR #1189, opened 2023-08-30
docs: Repository tag had a sneaky space

johanandren:


patriknw (commented):
hope this is not something that has been copy-pasted inte many more places?

johanandren:
I'll double check

johanandren:
Actually I was wrong about this being the cause, it's the same everywhere and works.

The reason it was broken is likely that when the published manually backport publish of the docs was made the paradox bump including the directive was missed. I'll republish.

johanandren:
Republished docs with right paradox and everything is fine. Reverted this PR since it wasn't the cause.

Issue #910, opened 2021-05-26
Kubernetes lease name truncation can lead to non-unique lease names


an-tex:
### Versions used 

Versions:
Akka: 2.6.13
Akka Management: 1.0.10

### Expected Behavior

When using a kubernetes lease for cluster sharding (https://doc.akka.io/docs/akka/current/typed/cluster-sharding.html#lease), the kubernetes lease names should be unique.

### Actual Behavior

In my case I've enabled a projection using a sharded daemon process, using a kubernetes lease for coordination. The lease names end up being prefixed with `application-shard-sharded-daemon-process-[...]` (42 characters) and being [truncated](https://github.com/akka/akka-management/blob/master/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala#L48) to 63 characters, this leaves only 22 characters for the actual projection name + tag. When truncated, the tag is the first to disappear, this generates the same lease name for different sharding processes. This in turn leads to an error loop with the following logs:

`Lease application-shard-sharded-daemon-process-[...] requested by client application@10.42.2.68:25520 is already owned by client. Previous lease was not released due to ungraceful shutdown.`

`Conflict during heartbeat to lease LeaseResource(Some(application@10.42.2.68:25520),8609152,1622013422731). Lease assumed to be released`

`sharded-daemon-process-[...]: Shard id [2] lease lost, stopping shard and killing [1] entities.`

an-tex:
About possible solutions:
1. When truncating, add a shortish hash suffix based on the full name
2. Use multiple domain segments, as the 63 character limit only counts for each segment. The whole name can be up to 253 characters. Meaning a lease name like `application.shard-sharded-daemon-process.projectionName.tag` would mitigate the limits.

octonato:
The prefix is a concatenation of different bits of information. As described here:

https://doc.akka.io/docs/akka/current/typed/cluster-sharding.html#lease

> To use a lease for sharding set akka.cluster.sharding.use-lease to the configuration location of the lease to use. Each shard will try and acquire a lease with with the name <actor system name>-shard-<type name>-<shard id> and the owner is set to the Cluster(system).selfAddress.hostPort.

Another alternative solution could be to make it configurable. But I think only the `<actor system name>-shard` can be made configurable. The `<type name>-<shard-id>` portion will need to be kept around. 

What about going for solution 2 as suggested by @an-tex, if we failure in case the final name is larger than 253 chars? 

[...]

PR #1194, opened 2023-09-21
fix: Fail if too long lease name, #910

patriknw:
* for backwards compatibility it's possible to allow old behavior with config, e.g. to support rolling update

Refs #910

Requires a way to define custom lease names, which is added in https://github.com/akka/akka/pull/32122

### DIFF
--- a/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
+++ b/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
@@ -60,10 +60,22 @@ object KubernetesLease {
    * Regex to follow: [a-z]([-a-z0-9]*[a-z0-9])
    * Limit the resulting name to 63 characters
    */
-  private def makeDNS1039Compatible(name: String): String = {
+  private def makeDNS1039Compatible(name: String, allowLeaseNameTruncation: Boolean): String = {
     val normalized =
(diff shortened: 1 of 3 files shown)


lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
```diff
@@ -60,10 +60,22 @@ object KubernetesLease {
...
-  private def makeDNS1039Compatible(name: String): String = {
+  private def makeDNS1039Compatible(name: String, allowLeaseNameTruncation: Boolean): String = {
     val normalized =
       Normalizer.normalize(name, Normalizer.Form.NFKD).toLowerCase.replaceAll("[_.]", "-").replaceAll("[^-a-z0-9]", "")
-    trim(truncateTo63Characters(normalized), List('-'))
+
+    if (allowLeaseNameTruncation)
+      trim(truncateTo63Characters(normalized), List('-'))
+    else {
+      if (normalized.length > 63)
```
patriknw, line 70: this is the same as we have for rolling update: https://github.com/akka/akka-management/blob/d348eff34b4afe4d9bca675b8a7d9edc92f6b79b/rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApi.scala#L91

johanandren (approved):

PR #1219, opened 2023-11-02
build: Publish snapshots

patriknw:


patriknw:
merging to see that it actually works

PR #1232, opened 2023-12-05
test: Harden rollingupdate-kubernetes-cr-test.sh

patriknw:
* failing in CI
* I think it should wait for ready, not only running
* more error diagnostics
* jdk 17 image (seems to work)
* but main issue is the resource requuests with minikube
  - argh, how many times are we going to run into this problem?
* update minikube and kubernetes versions


patriknw:
after much trial and error I have found the real problem and fixed this

johanandren (approved):
LGTM, good detective work!

PR #1241, opened 2023-12-15
build: Move away from it test configuration

johanandren:
Since now deprecated.

Also moves the lease-kubernetes-int-test into the directory with the other integration tests


johanandren:
I would have wanted to not have a integration-test directory but instead keep the module directories a flat structure with `module-it-is-testing-integration` for each (and use that as a standard across akka projects), but I think this is already enough of changing things in one PR, so deferring that.

octonato (approved):
LGTM

Issue #1242, opened 2023-12-15
DNS Resolution in KubernetesApiServiceDiscovery use deprecated pod.namespace 


vcanuel:
Versions Used
Akka version: 1.05

Expected Behavior
The KubernetesApiServiceDiscovery should correctly use service.namespace for DNS resolution instead of pod.namespace. This change aligns with the recent changes in Kubernetes DNS specifications  : https://github.com/kubernetes/dns/commit/a86af5089052d13bedb63b83601be1e12df6a74b

Actual Behaviour
The service discovery currently uses pod.namespace for DNS resolution, specifically in the following code block from 

https://github.com/akka/akka-management/blob/6a49c2c0f29959976d3a6a903db3c7761e8afab1/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala#L82C60-L82C60

```scala
val hostOrIp = if (rawIp) ip else s"${ip.replace('.', '-')}.$podNamespace.pod.$podDomain"
```
This leads to incorrect DNS resolution behavior, especially in environments like GKE Autopilot with Cloud DNS where IP.pod.namespace simply does not exist.

Suggested Fix
Replace pod.namespace with service.namespace in the respective code section for accurate DNS resolution as per the latest Kubernetes DNS specifications

PR #1244, opened 2023-12-18
fix: Default to kubernetes pod IP instead of deprecated pod DNS names

johanandren:
Naive search replace to see what breaks in CI

References #1242


vcanuel:
If I may, we are looking for pods (in  https://github.com/akka/akka-management/blob/6a49c2c0f29959976d3a6a903db3c7761e8afab1/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/KubernetesApiServiceDiscovery.scala#L149C16-L149C16)  but maybe we  need to find the service instead. Thanks for looking at this issue :)

johanandren:
Hmm, then I think I changed too much to start with, the label selection should still be about pods and I'll dig a bit deeper than a search-replace. 

As far as I understand the only thing we want to change here is to not use the pod dns names as those are deprecated.

johanandren:
The problem is actually that the svc name does not resolve unless there is a service collecting the pods and the pods are ready. Usually Akka forming cluster is something you want to wait for before marking the pods as ready, to not accept requests before a node has joined the cluster.

I wonder if the right thing isn't just to switch to `akka.discovery.kubernetes-api.use-raw-ip=true` to avoid using the deprecated dns names for the pods.

vcanuel:
> The problem is actually that the svc name does not resolve unless there is a service collecting the pods and the pods are ready. Usually Akka forming cluster is something you want to wait for before marking the pods as ready, to not accept requests before a node has joined the cluster.
> 
> 
> 
> I wonder if the right thing isn't just to switch to `akka.discovery.kubernetes-api.use-raw-ip=true` to avoid using the deprecated dns names for the pods.

That's how we fix it indeed ;)

johanandren:
Thanks for verifying, then that should probably be the default (and should anyway work with older Kubernetes versions as well)

octonato (approved):
LGTM

PR #1288, opened 2024-05-11
Add dependencies to same version check

Philippus:
Applying this PR will add the (newer) akka-lease-kubernetes and akka-rolling-update-kubernetes dependencies to the same version check. I think maybe this was missed when the modules were introduced.


johanandren (approved):

johanandren:
Thanks!

Issue #1302, opened 2024-07-24
Support configurable annotation for akka rolling update extension


huayangdu:
### Short description

I hope akka rolling update extension to support configurable annotation for AppVersion

### Details

Now we are using Argo Rollout instead of Deployment which can provide customized canary deployment, etc. It will also control the ReplicaSet rollout steps. But it doesn't have annotation like deployment.kubernetes.io/revision on ReplicaSet, it has rollout.argoproj.io/revision.  So now the former one is [hard code here](https://github.com/akka/akka-management/blob/main/rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesApiImpl.scala#L216) to get AppVersion from annotation on ReplicaSet. I wonder if we can support to make it configurable so that it can still work with other k8s controller resource.

PR #1303, opened 2024-07-25
feat: configurable choice of annotations for rolling update revision

leviramsey:
References #1302 

Adds a config setting `akka.rollingupdate.kubernetes.revision-annotation` to allow configurability (default is the prior behavior).  From there it's just dealing with Spray JSON.  Admittedly it's been a while (ever?) since I've done Spray JSON trickery, but (with apologies to Tolstoy) "all happy typeclass-and-AST JSON libraries are alike".

I don't have a k8s cluster handy to test, mind.

### DIFF
--- a/docs/src/main/paradox/rolling-updates.md
+++ b/docs/src/main/paradox/rolling-updates.md
@@ -237,7 +237,7 @@ Java
 
 #### Configuration
 
-The following configuration is required, more details for each and additional configurations can be found in [reference.conf](https://github.com/akka/akka-management/blob/main/rolling-updates-kubernetes/src/main/resources/reference.conf):
+The following configuration is required, more details for each and additional configurations can be found in [reference.conf](https://github.com/akka/akka-management/blob/main/rolling-update-kubernetes/src/main/resources/reference.conf):
 
 * `akka.rollingupdate.kubernetes.pod-name`: this can be provided by setting `KUBERNETES_POD_NAME` environment variable to `metadata.name` on the Kubernetes container spec.
 
--- a/rolling-update-kubernetes/src/main/resources/reference.conf
+++ b/rolling-update-kubernetes/src/main/resources/reference.conf
@@ -34,10 +34,10 @@ akka.rollingupdate.kubernetes {
     pod-name = ""
     pod-name = ${?KUBERNETES_POD_NAME}
 
-	# Annotations to check to determine the revision.  The first one in the list which matches
-	# wins.  The default is suitable for "vanilla" Kubernetes Deployments, but other CI/CD systems
-	# may set a different annotation.
-	revision-annotation = [ "deployment.kubernetes.io/revision" ]
+    # Annotations to check to determine the revision.  The first one in the list which matches
+    # wins.  The default is suitable for "vanilla" Kubernetes Deployments, but other CI/CD systems
+    # may set a different annotation.
+    revision-annotation = [ "deployment.kubernetes.io/revision" ]
 
     secure-api-server = true
 
--- a/rolling-update-kubernetes/src/main/resources/reference.conf
+++ b/rolling-update-kubernetes/src/main/resources/reference.conf
@@ -34,6 +34,11 @@ akka.rollingupdate.kubernetes {
     pod-name = ""
     pod-name = ${?KUBERNETES_POD_NAME}
 
+	# Annotations to check to determine the revision.  The first one in the list which matches
+	# wins.  The default is suitable for "vanilla" Kubernetes Deployments, but other CI/CD systems
+	# may set a different annotation.
+	revision-annotation = [ "deployment.kubernetes.io/revision" ]
+
     secure-api-server = true
 
     # Configuration for the Pod Deletion Cost extension
--- a/rolling-update-kubernetes/src/main/resources/reference.conf
+++ b/rolling-update-kubernetes/src/main/resources/reference.conf
@@ -35,7 +35,7 @@ akka.rollingupdate.kubernetes {
     pod-name = ${?KUBERNETES_POD_NAME}
 
     # Annotations to check to determine the revision.  The default is suitable for "vanilla"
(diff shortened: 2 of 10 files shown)


docs/src/main/paradox/rolling-updates.md
```diff
@@ -237,7 +237,7 @@ Java
 
 #### Configuration
 
-The following configuration is required, more details for each and additional configurations can be found in [reference.conf](https://github.com/akka/akka-management/blob/main/rolling-updates-kubernetes/src/main/resources/reference.conf):
+The following configuration is required, more details for each and additional configurations can be found in [reference.conf](https://github.com/akka/akka-management/blob/main/rolling-update-kubernetes/src/main/resources/reference.conf):
```
leviramsey, line 240: Not sure how long the typo has been there...

I think the comment in `reference.conf` is sufficient documentation for a niche feature.

rolling-update-kubernetes/src/main/resources/reference.conf
```diff
@@ -34,6 +34,11 @@ akka.rollingupdate.kubernetes {
     pod-name = ""
     pod-name = ${?KUBERNETES_POD_NAME}
 
+    # Annotations to check to determine the revision.  The first one in the list which matches
+    # wins.  The default is suitable for "vanilla" Kubernetes Deployments, but other CI/CD systems
+    # may set a different annotation.
+    revision-annotation = [ "deployment.kubernetes.io/revision" ]
```
sebastian-alfers, line 40: In practice, will there actually be multiple configured? If yes, should this be plural, like `revision-annotations`?

leviramsey: Probably not, but it's a somewhat bigger change down the road to go from exactly-one to multiple than to go from nothing to multiple.

huayangdu: if the first matched one wins, why not just keep it as exactly one match and set it with 
revision-annotation = "deployment.kubernetes.io/revision" 
we can override it in our setting with other annotation

johanandren: I agree, skip the list, and make it a single value.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesJsonSupport.scala
```diff
@@ -86,7 +91,41 @@ trait KubernetesJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
...
+    }
+
   implicit val replicaSetMedatataFormat: JsonFormat[ReplicaSetMetadata] = jsonFormat1(ReplicaSetMetadata.apply)
   implicit val podReplicaSetFormat: RootJsonFormat[ReplicaSet] = jsonFormat1(ReplicaSet.apply)
 }
+
+private[kubernetes] object KubernetesJsonSupport {
+  val defaultRevisionAnnotations: HasRevisionAnnotations = new HasRevisionAnnotations {
+    val revisionAnnotations = Seq("deployment.kubernetes.io/revision")
+  }
```
johanandren, line 130: No need to hardcode the default if we put it in reference.conf as a default?

leviramsey: That's basically there so there's no hard dependency (e.g. for tests) on having Config

johanandren: Hardcoding a value that is duplicated across config and code seems objectively worse than depending on config in test to me.

rolling-update-kubernetes/src/main/resources/reference.conf
```diff
@@ -34,6 +34,10 @@ akka.rollingupdate.kubernetes {
     pod-name = ""
     pod-name = ${?KUBERNETES_POD_NAME}
 
+    # Annotations to check to determine the revision.  The default is suitable for "vanilla"
+	# Kubernetes Deployments, but other CI/CD systems may set a different annotation.
```
sebastian-alfers, line 38: Is this indentation a leftover? And does it use a tab, instead of space currently?

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesJsonSupport.scala
```diff
@@ -86,7 +91,30 @@ trait KubernetesJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
   implicit val podMetadataFormat: JsonFormat[PodMetadata] = jsonFormat1(PodMetadata.apply)
   implicit val podFormat: RootJsonFormat[Pod] = jsonFormat1(Pod.apply)
 
-  implicit val replicaAnnotationFormat: JsonFormat[ReplicaAnnotation] = jsonFormat1(ReplicaAnnotation.apply)
+  implicit val replicaAnnotationFormat: RootJsonFormat[ReplicaAnnotation] =
+    new RootJsonFormat[ReplicaAnnotation] {
+      // Not sure if we ever write this out, but if we do, this will let us write out exactly what we took in
```
sebastian-alfers, line 96: Shouldn't we add the "writes" part only then if really need?

sebastian-alfers: Then we can even further simplify the case class `ReplicaAnnotation`.

leviramsey: It's a `JsonFormat` so needs a `writes`.  If we're going to implement it, we might as well implement it right

johanandren (approved):

PR #1309, opened 2024-08-08
fix: init revisionAnnotation in KubernetesJsonSupport in native spec

sebastian-alfers:
Follow up to https://github.com/akka/akka-management/pull/1303


patriknw (approved):
LGTM

PR #1322, opened 2024-09-25
Only request running pods from Kubernetes API

leviramsey:
It seems like we'd be most interested in the pods which are currently running (cc: @girdharshubham ).  Without this filter, especially in environments where rollouts are frequent (e.g. dev/test environments), there's a chance that getting so many pods could cause HTTP responses which akka http rejects.

Can hide this behind a config setting if needed, but I'm not sure why one would want to discover pods which used to be there.

### DIFF
--- a/discovery-azure-api/src/main/scala/akka/discovery/azureapi/rbac/aks/AzureRbacAksServiceDiscovery.scala
+++ b/discovery-azure-api/src/main/scala/akka/discovery/azureapi/rbac/aks/AzureRbacAksServiceDiscovery.scala
@@ -185,7 +185,7 @@ final class AzureRbacAksServiceDiscovery(implicit system: ExtendedActorSystem) e
     for {
       ks <- kubernetesSetup
       token <- fetchAccessToken.map(_.getToken)
-      request <- podRequest(token, ks.namespace, selector)
+      request <- podRequest(token, ks.namespace, selector, true)
       pods <- pods(ks.ctx, request, resolveTimeout)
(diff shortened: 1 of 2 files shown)


leviramsey:
I assume the link checker failure is expected, post #1319

johanandren:
Not sure, but, could this cause problems with ready check on a cold cluster start?

leviramsey:
> Not sure, but, could this cause problems with ready check on a cold cluster start?

From a conversation with @girdharshubham, it appears that running does not take readiness into account:

```
kubectl --context=arn:aws:eks:us-east-2:xyz:cluster/dev-us-east-2 get pods -n some-uuid --field-selector=status.phase==Running 
NAME                                                READY   STATUS             RESTARTS           AGE
some-name-644cc5669-pwn62   2/3     CrashLoopBackOff   1550 (4m26s ago)   5d9h
some-name-cf679b76b-2hm46   2/3     CrashLoopBackOff   1576 (2m51s ago)   5d11h
some-name-cf679b76b-p8crf   2/3     CrashLoopBackOff   1570 (112s ago)    5d11h
some-name-cf679b76b-zr772   2/3     CrashLoopBackOff   1580 (3m43s ago)   5d11h
```

johanandren (approved):
LGTM, seems safe

leviramsey:
Note that this does not remove the filtering of the returned results for running pods on our side.  It may make sense to leave that in as a "defense in depth".

PR #1369, opened 2025-02-06
feat [discovery-kubernetes-api]: ability to consider readiness

leviramsey:
The `discovery-kubernetes-api` Discovery plugin is suitable for cluster bootstrap as it discovers unready pods: this in turn allows pods to not be ready until they have joined the cluster.  However, this makes it less suitable for discovering external services to interact with, as the discovered endpoints may not be ready to be interacted with.

This change adds an `ExternalKubernetesApiServiceDiscovery` plugin which only discovers ready pods.  It's implemented separately from the regular plugin so that both could conceivably be used in the same application (though the plugins would share the same configuration for k8s).


johanandren:
Before doing any code review: is there any clear benefit from using the K8 API for "regular" service discovery instead of DNS like services would normally do? It's not obvious to me (while the cluster bootstrap use case for K8 API is pretty obvious that it is useful).

leviramsey:
> Before doing any code review: is there any clear benefit from using the K8 API for "regular" service discovery instead of DNS like services would normally do?

The main benefit is niche: discovery via DNS is going to be constrained by DNS server TTL (CoreDNS defaults to 5s) so has some likelihood of returning a stale result.  Since DNS is likely global to a given k8s cluster (not sure if its possible to give different DNS configs to different pods) this allows configuring k8s DNS for most (less-critical) cases and using API for discovery in the cases where something more up-to-date is desired.

johanandren (commented):
Impl looks good but we should discuss more in team if we should add this or not

patriknw (commented):
looking good

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/JsonFormat.scala
```diff
@@ -18,7 +18,8 @@ import spray.json._
   implicit val containerFormat: JsonFormat[Container] = jsonFormat2(Container.apply)
   implicit val podSpecFormat: JsonFormat[PodSpec] = jsonFormat1(PodSpec.apply)
   implicit val containerStatusFormat: JsonFormat[ContainerStatus] = jsonFormat2(ContainerStatus.apply)
-  implicit val podStatusFormat: JsonFormat[PodStatus] = jsonFormat3(PodStatus.apply)
+  implicit val podConditionFormat: JsonFormat[PodCondition] = jsonFormat2(PodCondition.apply)
```
leviramsey, line 21: Part of me kind of would like to turn the conditions into a `Map` and do some magic in the spray format, but

* I'm not totally up on my spray-fu
* the convention here is to mirror the JSON structure directly

discovery-kubernetes-api/src/main/resources/reference.conf
```diff
@@ -41,4 +41,9 @@ akka.discovery {
     # When set, validate the container is not in 'waiting' state
     container-name = ""
   }
+
+  # All configuration for this discovery method, apart from the class, is shared with kubernetes-api
+  kubernetes-api-external = {
```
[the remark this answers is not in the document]
leviramsey: `kubernetes-api-for-client` strikes me as the best balance of conveying intent and succinctness.  I'm putting the description of the two further up in `reference.conf` with the all-caps recommendation to not use `for-client` to bootstrap and documenting explicitly how to use `kubernetes-api` for bootstrap and `kubernetes-api-for-client` as the default discovery method otherwise.

discovery-kubernetes-api/src/main/mima-filters/1.6.0.backwards.excludes
```diff
@@ -0,0 +1,11 @@
+ProblemFilters.exclude[DirectAbstractMethodProblem]("akka.discovery.ServiceDiscovery.lookup")
```
leviramsey, line 1: Not actually sure why this one fires... tbh: the method is still there (in `BaseKubernetesApiServiceDiscovery`)?

johanandren: Also, that class is not in this project so quite unexpected.

johanandren (approved):

PR #1385, opened 2025-04-29
chore: refactor the auth settings

johanandren:
Just noticed they were a bit messy so separated the concerns a bit.


patriknw (approved):
LGTM

PR #1413, opened 2025-08-21
fix: reload tokens in kubernetes lease

### LINKED ISSUE #1412: Kubernetes Lease fails Kubernetes API calls after first ServiceAccount token refresh
### Versions used 

Akka version: 2.8.7

Akka Management version: 1.4.1

Kubernetes version: 1.32.5

### Expected Behavior

Node should be able to acquire the Kubernetes lease resource.

### Actual Behavior

Node failed to access Kubernetes lease resource with Kubernetes API returning HTTP 401 Unauthorized status.

Kubernetes Lease appears to fetch the ServiceAccount token only when the `KubernetesApi` is first initialized: https://github.com/akka/akka-management/blob/4b884424a52817caee10256675b437e0e08da693/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala#L91

The token is not refreshed after the `LeaseActor` has started:
https://github.com/akka/akka-management/blob/4b884424a52817caee10256675b437e0e08da693/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala

According to [Kubernetes ServiceAccount documentation](https://kubernetes.io/docs/concepts/security/service-accounts/):
> In v1.22 and later, Kubernetes gets a short-lived, automatically rotating token

The current ServiceAccount token is valid when used from the pod by running curl command against the lease resources, meaning it has the required role:
```
$ curl -sSk -H "Authorization: Bearer $(cat /run/secrets/kubernetes.io/serviceaccount/token)" "https://kubernetes.default.svc/apis/akka.io/v1/namespaces/retracted-namespace/leases/retracted-lease"
```

The Lease actor should refresh the token to work after first token rotation, when running on Kubernetes >= 1.22.

### Relevant logs

```
2025-07-20T15:37:56.876 (UTC) WARN  [a.c.lease.kubernetes.LeaseActor] [a://retracted-cluster@1.2.3.4:25520/system/kubernetesLease2] - Failure communicating with the API server for owner retracted-cluster@1.2.3.4:25520 lease retracted-lease: [Unauthorized to communicate with Kubernetes API server. See https://doc.akka.io/docs/akka-management/current/kubernetes-lease.html#role-based-access-control for setting up access control. Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
```

### Reproducible Test Case

Related tickets:
- #1040

jroper:
Fixes #1412

This implements token reload for kubernetes lease coordination. The tokens are loaded as required, and cached for 10 minutes.

10 minutes is selected because kubelet token rotation is done when a token, which by default expires in 1 hour, has reached 80% of its time to live, that is, after 48 minutes.

### DIFF
--- a/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/Settings.scala
+++ b/discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/Settings.scala
@@ -5,10 +5,11 @@
 package akka.discovery.kubernetes
 
 import java.util.Optional
-
 import akka.actor._
 import com.typesafe.config.Config
 
+import scala.concurrent.duration.FiniteDuration
+import scala.jdk.DurationConverters._
 import scala.jdk.OptionConverters._
 
 final class Settings(system: ExtendedActorSystem) extends Extension {
@@ -39,6 +40,9 @@ final class Settings(system: ExtendedActorSystem) extends Extension {
 
   val apiServicePortEnvName: String =
     kubernetesApi.getString("api-service-port-env-name")
+
+  val apiTokenTtl: FiniteDuration =
+    kubernetesApi.getDuration("api-token-reload-interval").toScala
 
   val podNamespacePath: String =
     kubernetesApi.getString("pod-namespace-path")
--- a/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
+++ b/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
@@ -87,13 +87,20 @@ class KubernetesLease private[akka] (system: ExtendedActorSystem, leaseTaken: At
 
   private val k8sSettings = KubernetesSettings(settings.leaseConfig, settings.timeoutSettings)
   private val leaseName = makeDNS1039Compatible(settings.leaseName, k8sSettings.allowLeaseNameTruncation)
+  private object Dispatchers {
+    implicit val blocking: ExecutionContext = system.dispatchers.lookup(DefaultBlockingDispatcherId)
+  }
+
+  private def k8sApiToken(): Future[String] = {
+    import Dispatchers.blocking
+    Future {
+      readConfigVarFromFilesystem(k8sSettings.apiTokenPath, "api-token").getOrElse("")
+    }
+  }
 
   private val k8sApi: Future[KubernetesApi] = {
-    implicit val blockingDispatcher: ExecutionContext = system.dispatchers.lookup(DefaultBlockingDispatcherId)
+    import Dispatchers.blocking
     for {
-      apiToken: String <- Future {
-        readConfigVarFromFilesystem(k8sSettings.apiTokenPath, "api-token").getOrElse("")
-      }
       namespace: String <- Future {
         k8sSettings.namespace
           .orElse(readConfigVarFromFilesystem(k8sSettings.namespacePath, "namespace"))
@@ -101,7 +108,7 @@ class KubernetesLease private[akka] (system: ExtendedActorSystem, leaseTaken: At
       }
       httpsContext <- Future(clientHttpsConnectionContext())
     } yield {
-      new KubernetesApiImpl(system, k8sSettings, namespace, apiToken, httpsContext)
+      new KubernetesApiImpl(system, k8sSettings, namespace, k8sApiToken, httpsContext)
     }
   }
   private implicit val timeout: Timeout = Timeout(settings.timeoutSettings.operationTimeout)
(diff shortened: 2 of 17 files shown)


lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
```diff
@@ -87,21 +87,28 @@ class KubernetesLease private[akka] (system: ExtendedActorSystem, leaseTaken: At
...
+    }
+  }
 
   private val k8sApi: Future[KubernetesApi] = {
-    implicit val blockingDispatcher: ExecutionContext = system.dispatchers.lookup(DefaultBlockingDispatcherId)
+    import Dispatchers.blocking
     for {
-      apiToken: String <- Future {
-        readConfigVarFromFilesystem(k8sSettings.apiTokenPath, "api-token").getOrElse("")
-      }
```
patriknw, line 96: Do we have the same problem in rolling-update-kubernetes and discovery-kubernetes-api?
Search for "api-token" to find those.

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/KubernetesApiSpec.scala
```diff
@@ -53,7 +56,7 @@ class KubernetesApiSpec
 
   implicit val patience: PatienceConfig = PatienceConfig(testKitSettings.DefaultTimeout.duration)
 
-  val underTest = new KubernetesApiImpl(system, settings, "lease", "token", None)
+  val underTest = new KubernetesApiImpl(system, settings, "lease", () => Future.successful("token"), None)
```
patriknw, line 59: would it be possible to add a test with short token reload interval and verify that the `() => Future(token)` is called?

patriknw (commented):
looking good

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/KubernetesLease.scala
jroper: Yes we do. There's a lot of repeat code here, should we factor it out into a common Kubernetes utility library?

patriknw: That was intentionally split up like that since the modules don't have dependencies to each other. Adding another module for this purpose might be worth it, or not.

jroper: Done.

patriknw (commented):
looking good

johanandren (commented):
Logic looks good, some minor feedback, and doesn't currently compile compile.

discovery-kubernetes-api/src/main/scala/akka/discovery/kubernetes/Settings.scala
```diff
@@ -40,6 +41,9 @@ final class Settings(system: ExtendedActorSystem) extends Extension {
   val apiServicePortEnvName: String =
     kubernetesApi.getString("api-service-port-env-name")
 
+  val apiTokenTtl: FiniteDuration =
+    kubernetesApi.getDuration("api-token-reload-interval").toScala
```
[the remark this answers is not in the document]
patriknw: we have similar things in 3 modules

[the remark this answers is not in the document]
jroper: It needs to go in both, I missed discovery.

rolling-update-kubernetes/src/main/scala/akka/rollingupdate/kubernetes/KubernetesSettings.scala
```diff
@@ -64,14 +64,16 @@ private[kubernetes] class KubernetesSettings(
...
     val namespace: Option[String],
     val namespacePath: String,
     val podName: String,
     val secure: Boolean,
     val apiServiceRequestTimeout: FiniteDuration,
     val customResourceSettings: CustomResourceSettings,
     val revisionAnnotation: String,
-    val bodyReadTimeout: FiniteDuration = 1.second
+    val bodyReadTimeout: FiniteDuration = 1.second,
+    val insecureTokens: Boolean = false
```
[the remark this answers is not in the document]
jroper: Done, also in kubernetes-lease.

patriknw (approved):
LGTM, thanks!

PR #1440, opened 2026-04-03
fix: Enable a bad lease to recover if there is a timeout

JustinPihony:
## Retry heartbeat on transient API failure (configurable)

When the Kubernetes API server experiences a transient failure during a lease heartbeat, the previous behaviour was to immediately release the lease and invoke the lease-lost callback — even if the lease timeout was nowhere near expiry. This caused unnecessary leadership changes and cluster disruptions from brief network hiccups or API server restarts.

### What changed

A new config option `heartbeat-fail-fast-on-error` (default `true`, preserving existing behaviour) controls what happens when a heartbeat `Failure` is received while in the `Granted` state:

- **`true` (default):** existing fail-fast behaviour — any error immediately releases the lease.
- **`false`:** the heartbeat is retried at the normal `heartbeat-interval` cadence, as long as the lease has not yet timed out (i.e. `now < lastSuccessfulHeartbeat + heartbeat-timeout - 2 * heartbeat-interval`). Once the timeout window is exhausted, the lease is released as before.

```hocon
akka.coordination.lease.kubernetes {
  # Set to false to retry heartbeats on transient API errors rather than
  # immediately releasing the lease.
  heartbeat-fail-fast-on-error = false
}
```

### Implementation notes

- `GrantedVersion` gains a `lastHeartbeatTime: Long` field (set on construction and updated on each successful write response) so the retry guard can compare against the lease timeout without an extra API call.
- On a successful heartbeat response, `lastHeartbeatTime` is refreshed alongside `version`.
- The existing `hasLeaseTimedOut` predicate is reused for the retry guard, giving consistent semantics with the TTL logic elsewhere in the actor.

### Tests

- Existing heartbeat-failure tests are unchanged in behaviour (they run against the default fail-fast mode).
- Two new tests cover the retry path (`RetryTest` fixture sets `heartbeat-fail-fast-on-error = false`):
  - verifies `granted` stays `true` and a new heartbeat is issued after a transient failure
  - verifies the lease is released when fail-fast is explicitly enabled
- `KubernetesSettingsSpec` asserts the default value and that it can be overridden.


lease-kubernetes/src/main/resources/reference.conf
```diff
@@ -52,6 +52,12 @@ akka.coordination.lease.kubernetes {
     # of different lease resources.
     allow-lease-name-truncation = off
 
+    # When set to true (the default), any API server failure during heartbeat immediately
+    # releases the lease. When set to false, a transient failure will instead be retried
+    # as long as the lease timeout has not been reached (i.e. the timeout is "far enough
+    # off"), preventing a brief network hiccup from dropping the lease unnecessarily.
+    heartbeat-fail-fast-on-error = true
```
JustinPihony, line 59: It might be fine to change the default behavior on this?

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
```diff
@@ -270,6 +270,33 @@ class LeaseActorSpec
...
+      // Heartbeat fails transiently - actor should retry rather than release
+      updateProbe.expectMsg((ownerName, currentVersion))
+      updateProbe.reply(Failure(new LeaseException("Transient API failure")))
+
+      // A new heartbeat should be sent with the same version (K8s state didn't change)
+      updateProbe.expectMsg(leaseSettings.timeoutSettings.heartbeatInterval * 3, (ownerName, currentVersion))
+      granted.get() shouldEqual true
+    }
+
+    "heartbeat failure should release lease when heartbeat-fail-fast-on-error is enabled" in new Test {
```
johanandren: This test has the default `heartbeatFailFastOnError = true` so I think this exact case is already covered by existing test `heartbeat fail should set granted to false`

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
```diff
@@ -270,6 +270,33 @@ class LeaseActorSpec
...
+      granted.get() shouldEqual true
+
+      // Heartbeat fails transiently - actor should retry rather than release
+      updateProbe.expectMsg((ownerName, currentVersion))
+      updateProbe.reply(Failure(new LeaseException("Transient API failure")))
+
+      // A new heartbeat should be sent with the same version (K8s state didn't change)
+      updateProbe.expectMsg(leaseSettings.timeoutSettings.heartbeatInterval * 3, (ownerName, currentVersion))
+      granted.get() shouldEqual true
+    }
```
johanandren, line 293: I think this will be flaky, 25ms * 3 means at most 75ms before that expect fails, best to override leaseSettings with a longer heartbeat timeout in this case.

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
```diff
@@ -477,4 +506,8 @@ class LeaseActorSpec
     }
 
   }
+
+  trait RetryTest extends Test {
+    override def heartbeatFailFastOnError: Boolean = false
+  }
```
johanandren: Only used for a single test case, so just do that override inline in the test instead of this trait.

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -67,8 +71,13 @@ private[akka] object LeaseActor {
...
   case class InvalidRequest(reason: String) extends Response with DeadLetterSuppression
 
-  def props(k8sApi: KubernetesApi, settings: LeaseSettings, leaseName: String, granted: AtomicBoolean): Props = {
-    Props(new LeaseActor(k8sApi, settings, leaseName, granted))
+  def props(
+      k8sApi: KubernetesApi,
+      settings: LeaseSettings,
+      leaseName: String,
+      granted: AtomicBoolean,
+      heartbeatFailFastOnError: Boolean = true): Props = {
```
johanandren: Skip the default parameter, it is only here for the test anyway.

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -50,7 +50,11 @@ private[akka] object LeaseActor {
       operationStartTime: Long = System.nanoTime())
       extends Data
       with ReplyRequired
-  case class GrantedVersion(version: String, leaseLostCallback: Option[Throwable] => Unit) extends Data
+  case class GrantedVersion(
+      version: String,
+      leaseLostCallback: Option[Throwable] => Unit,
+      lastHeartbeatTime: Long = System.currentTimeMillis())
```
johanandren: `currentTimeMillis` moves with wall clock, so what happens when the system clock is adjusted and it jumps backwards or forwards? Use `System.nanoTime` instead, that is guaranteed to be monotonic.

The existing code uses timestamp from k8 so that will need to stay wall clock. But this new timestamp is only to figure out time since last heart beat which is completely local to the actor.

lease-kubernetes/src/main/mima-filters/1.6.4.backwards.excludes
johanandren, line 1: Latest release is 1.6.4, so this should be in 1.6.4.backwards.excludes

lease-kubernetes/src/main/resources/reference.conf
```diff
@@ -52,6 +52,12 @@ akka.coordination.lease.kubernetes {
     # of different lease resources.
     allow-lease-name-truncation = off
 
+    # When set to true (the default), any API server failure during heartbeat immediately
+    # releases the lease. When set to false, a transient failure will instead be retried
+    # as long as the lease timeout has not been reached (i.e. the timeout is "far enough
+    # off"), preventing a brief network hiccup from dropping the lease unnecessarily.
```
johanandren: Let's try to make it more correct, and concise, something like:

```suggestion
    # When set to true (the default), any API server failure during heartbeat immediately
    # releases the lease. When set to false, a failure will instead be retried
    # until `heartbeat-timeout - 2 * heartbeat-interval` has elapsed since the last successful heartbeat before releasing the lease if still failing.
```

docs/src/main/paradox/kubernetes-lease.md
```diff
@@ -155,4 +155,8 @@ Q. What happens if the node that holds the lease crashes?
 
 A. Each lease has a Time To Live (TTL) that is set `akka.coordination.lease.kubernetes.heartbeat-timeout` which defaults to 120s. A lease holder updates the lease every `1/10` of the timeout to keep the lease. If the TTL passes without
    the lease being updated another node is allowed to take the lease. For ultimate safety this timeout can be set very high but then an operator would need to come and clear the lease if a lease owner crashes with the lease taken.
+
+Q. What happens if the Kubernetes API server has a transient failure during a heartbeat?
+
+A. By default (`heartbeat-fail-fast-on-error = true`) the lease is immediately released and the lease-lost callback is invoked, even if the lease timeout is far from expiry. This can cause unnecessary leadership changes during brief network hiccups or API server restarts. Setting `heartbeat-fail-fast-on-error = false` changes this behaviour so that the heartbeat is retried at the normal interval as long as the lease has not yet timed out, preventing transient failures from needlessly dropping the lease.
```
johanandren: This should also be careful to mention that the feature introduces some pretty clear drawbacks:

With the defaults (timeout 120s, interval 12s) that's ~96 seconds of silently retrying before the application is told it lost the lease. If the API server is genuinely down (not transiently), every consumer of the lease is operating on stale "I still hold it" information for that whole window.

The K8s lease TTL is enforced by other nodes reading the resource and deciding "the owner hasn't refreshed in time, I can take it." If the heartbeat writes are failing because the connection to the API server is broken — but other nodes can still reach it it could end up with a scenario where multiple nodes has the lease (for example because of GC pause, clock skew). With the lease timeout there are safety margin against this.

It also kind of inverts the the default "if in doubt, drop it" which is there for safety: it optimizes for availability over safety, which is generally not what you want for SBR, cluster singleton or sharding where the lease is hooked into Akka internals.

lease-kubernetes/src/main/resources/reference.conf
johanandren: Was clear from my other review comments yesterday, but, to be extra clear, no. Let's stay with the safe as default.

docs/src/main/paradox/kubernetes-lease.md
```diff
@@ -158,5 +158,9 @@ A. Each lease has a Time To Live (TTL) that is set `akka.coordination.lease.kube
...
+
+   **Important trade-offs when using `false`:** With the default settings (timeout 120s, interval 12s) the retry window is ~96 seconds. During that window the local node continues to believe it holds the lease while writes to the API server are failing. If other nodes can still reach the API server they may observe the lease as expired and take it, resulting in multiple nodes simultaneously believing they hold the lease. The default `true` setting embodies a deliberate "if in doubt, drop it" safety principle that is especially important for Split Brain Resolver, Cluster Singleton, and Cluster Sharding, where holding a stale lease can cause split-brain scenarios. @@@ warning
+   Setting `heartbeat-fail-fast-on-error = false` trades that safety margin for availability and should only be used if the implications are well understood.
+   @@@
```
johanandren, line 165: I think this warning block will need to be on its own line and maybe with an empty row in-between to render correctly.

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -197,14 +197,17 @@ private[akka] class LeaseActor(
...
     case Event(WriteResponse(Left(lr @ _)), GrantedVersion(_, leaseLost, _)) =>
       log.warning("Conflict during heartbeat to lease {}. Lease assumed to be released.", lr)
       granted.set(false)
       executeLeaseLockCallback(leaseLost, None)
       goto(Idle).using(ReadRequired)
     case Event(Failure(t), gv @ GrantedVersion(_, leaseLost, lastHeartbeatTime)) =>
-      if (!heartbeatFailFastOnError && !hasLeaseTimedOut(lastHeartbeatTime)) {
+      val elapsed = System.nanoTime() - lastHeartbeatTime
+      val retryWindowNanos =
+        (settings.timeoutSettings.heartbeatTimeout - 2 * settings.timeoutSettings.heartbeatInterval).toNanos
```
johanandren, line 209: This doesn't change so can be calculated up front and kept in an actor field.

johanandren (commented):
Two more small things then I think it is good to go

patriknw:
I'm reviewing this, don't merge yet

docs/src/main/paradox/kubernetes-lease.md
```diff
@@ -155,4 +155,12 @@ Q. What happens if the node that holds the lease crashes?
 
 A. Each lease has a Time To Live (TTL) that is set `akka.coordination.lease.kubernetes.heartbeat-timeout` which defaults to 120s. A lease holder updates the lease every `1/10` of the timeout to keep the lease. If the TTL passes without
    the lease being updated another node is allowed to take the lease. For ultimate safety this timeout can be set very high but then an operator would need to come and clear the lease if a lease owner crashes with the lease taken.
+
+Q. What happens if the Kubernetes API server has a transient failure during a heartbeat?
```
patriknw, line 159: I added some more docs in https://github.com/akka/akka-management/pull/1441

lease-kubernetes/src/main/resources/reference.conf
patriknw: I don't think we should make this optional, we should always use the retries.

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -187,19 +197,29 @@ private[akka] class LeaseActor(
...
-      log.warning("Failure during heartbeat to lease: [{}]. Lease assumed to be released.", t.getMessage)
-      granted.set(false)
-      executeLeaseLockCallback(leaseLost, Some(t))
-      goto(Idle).using(ReadRequired)
-    case Event(Release(), GrantedVersion(version, leaseLost)) =>
+    case Event(Failure(t), gv @ GrantedVersion(_, leaseLost, lastHeartbeatTime)) =>
+      val elapsed = System.nanoTime() - lastHeartbeatTime
+      val retryWindowNanos =
+        (settings.timeoutSettings.heartbeatTimeout - 2 * settings.timeoutSettings.heartbeatInterval).toNanos
+      if (!heartbeatFailFastOnError && elapsed < retryWindowNanos) {
```
patriknw, line 210: slightly different in https://github.com/akka/akka-management/pull/1441

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
```diff
@@ -270,6 +270,28 @@ class LeaseActorSpec
...
+
+      // Heartbeat fails transiently - actor should retry rather than release
+      updateProbe.expectMsg((ownerName, currentVersion))
+      updateProbe.reply(Failure(new LeaseException("Transient API failure")))
+
+      // A new heartbeat should be sent with the same version (K8s state didn't change)
+      updateProbe.expectMsg(leaseSettings.timeoutSettings.heartbeatInterval * 3, (ownerName, currentVersion))
+      granted.get() shouldEqual true
+    }
+
```
patriknw, line 294: more tests in https://github.com/akka/akka-management/pull/1441

patriknw (commented):
I did a separate implementation in https://github.com/akka/akka-management/pull/1441

JustinPihony:
Closing in favor of #1441

PR #1441, opened 2026-04-08
fix: retry lease heartbeat if timeout far enough off

patriknw:
Alternative to https://github.com/akka/akka-management/pull/1440

### DIFF
--- a/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
+++ b/lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
@@ -50,7 +50,12 @@ private[akka] object LeaseActor {
       operationStartTime: Long = System.nanoTime())
       extends Data
       with ReplyRequired
-  case class GrantedVersion(version: String, leaseLostCallback: Option[Throwable] => Unit) extends Data
+  case class GrantedVersion(
+      version: String,
+      leaseLostCallback: Option[Throwable] => Unit,
+      lastHeartbeatTime: Long = System.nanoTime(),
+      retryCount: Int = 0)
+      extends Data
 
   sealed trait Command
   case class Acquire(leaseLostCallback: Option[Throwable] => Unit = ConstantFun.scalaAnyToUnit) extends Command
@@ -60,6 +65,7 @@ private[akka] object LeaseActor {
   private case class ReadResponse(response: LeaseResource) extends Command
   private case class WriteResponse(response: Either[LeaseResource, LeaseResource]) extends Command
   private case object Heartbeat extends Command
+  private case class HeartbeatRetry(retryCount: Int) extends Command
 
   sealed trait Response
   case object LeaseAcquired extends Response
@@ -177,7 +183,7 @@ private[akka] class LeaseActor(
   }
 
   when(Granted) {
-    case Event(Heartbeat, GrantedVersion(version, _)) =>
+    case Event(Heartbeat, GrantedVersion(version, _, _, _)) =>
       log.debug("Heartbeat: updating lease time. Version {}", version)
       pipe(k8sApi.updateLeaseResource(leaseName, ownerName, version).map(WriteResponse.apply)).to(self)
       stay()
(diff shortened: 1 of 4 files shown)


johanandren:
Always on still makes me cautious, because inverting the the default "if in doubt, drop it". Availability over safety is generally not what we want for SBR, cluster singleton or sharding where the lease is hooked into Akka internals.

Another aspect I commented on the original PR: With the defaults (timeout 120s, interval 12s) that's ~96 seconds of silently retrying before the application is told it lost the lease. If the API server is genuinely down (not transiently), every consumer of the lease is operating on stale "I still hold it" information for that whole window. Isn't it often better to act early?

patriknw:
Drop after one single failed api call is not very robust, since it's rather likely that some calls will fail. It's a heavy process to re-assign the lease to someone else, at least when using for singleton and sharding. We are trying to add retries but not for too long.

If the API server is genuinely down, another Akka node will not be able to acquire the lease either, and resulting in longer total unavailability.

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -280,6 +326,24 @@ private[akka] class LeaseActor(
...
+
+  private def hasTimeLeftForHeartbeatRetry(lastHeartbeatTime: Long): Boolean = {
+    val elapsed = (System.nanoTime() - lastHeartbeatTime).nanos
+    // Other nodes consider the lease expired at: heartbeatTimeout - 2*heartbeatInterval.
+    // We must give up before that, subtracting:
+    // - operationTimeout: time needed to complete the retry API call itself
+    // - heartbeatInterval: additional buffer for clock skew between nodes
+    val otherNodeTakeover = settings.timeoutSettings.heartbeatTimeout - (2 * settings.timeoutSettings.heartbeatInterval)
+    val safetyMargin = settings.timeoutSettings.operationTimeout + settings.timeoutSettings.heartbeatInterval
+    elapsed < otherNodeTakeover - safetyMargin
```
patriknw, line 344: if we want to increase the safety margin more we could increase the margin here, making less room for retries

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
```diff
@@ -280,6 +326,24 @@ private[akka] class LeaseActor(
     goto(Granting).using(OperationInProgress(reply, version, leaseLost))
   }
 
+  private def heartbeatRetryDelay(retryCount: Int): FiniteDuration = {
+    val interval = settings.timeoutSettings.heartbeatInterval
+    // Exponential backoff: interval/4, interval/2, interval, interval, ...
+    val delay = interval / 4 * (1L << math.min(retryCount - 1, 2))
```
patriknw, line 332: I was considering if it should retry even more frequently, but that would also increase the risk of loading the API server even more, which might be the original reason for the request timeout

lease-kubernetes/src/main/scala/akka/coordination/lease/kubernetes/LeaseActor.scala
johanandren: That would make me feel a bit less worried about having it always on.

patriknw: Here is how the parameters play out if we increase the safetyMargin to `op-timeout + 2 * interval`

**Parameters:**

- **interval** — `heartbeat-interval`: how often the lease holder writes a heartbeat to the API server
- **timeout** — `heartbeat-timeout`: how long a lease can go without update before another node may take it
- **op-timeout** — `lease-operation-timeout`: time allowed for a single API call
- **otherNodeTakeover** — `timeout - 2*interval`: elapsed time after which other nodes consider the lease expired and may acquire it
- **safetyMargin** — how much earlier the lease holder gives up retrying, to ensure it stops before other nodes can take over
- **Retry deadline** — `otherNodeTakeover - safetyMargin`: maximum elapsed time since last successful heartbeat during which retries are attempted
- **Gap** — equals safetyMargin; the time buffer between when the holder gives up and when other nodes can take over
- **~Retries** — approximate number of retries within the deadline, using exponential backoff delays of `interval/4`, `interval/2`, `interval`, `interval`, ...

**Alternative A: `safetyMargin = op-timeout + interval`** (current)

| Config | Ratio | takeover | margin | Retry deadline | Gap | ~Retries |
|---------------------------------------|-------|-------------------|--------------|----------------|-----|----------|
| 12s / 120s / 5s (default) | 1:10 | 96s | 17s | 79s | 17s | ~7 |
| 24s / 120s / 10s | 1:5 | 72s | 34s | 38s | 34s | ~2 |
| 30s / 120s / 10s | 1:4 | 60s | 40s | 20s | 40s | ~1 |

**Alternative B: `safetyMargin = op-timeout + 2 * interval`** (larger margin)

| Config | Ratio | takeover | margin | Retry deadline | Gap | ~Retries |
|---------------------------------------|-------|-------------------|--------------|----------------|-----|----------|
| 12s / 120s / 5s (default) | 1:10 | 96s | 29s | 67s | 29s | ~6 |
| 24s / 120s / 10s | 1:5 | 72s | 58s | 14s | 58s | ~1 |
| 30s / 120s / 10s | 1:4 | 60s | 70s | 0s | N/A | 0 |

Alternative B gives a much larger gap at the cost of fewer retries. At ratio 1:4 it eliminates retries entirely — no retry can fit within the safety margin.

johanandren: Ok, maybe it is safe enough as is with current then.

johanandren (approved):
LGTM

PR #1444, opened 2026-04-23
feat: Retries for lease acquire

johanandren:
Follow-up to #1441: applies the same retry-with-backoff pattern to the lease acquisition path (PendingRead / Granting), so transient kube-API failures during Acquire no longer immediately fail the caller.

  - Budget is the caller's operationTimeout (vs. the heartbeat timeout used in #1441); retry scheduled only if elapsed + delay < operationTimeout / 2, else the original failure is returned.
  - Exponential backoff from operationTimeout / 20, capped at / 5.
  - acquireStartTime is threaded from PendingRead into Granting so the budget spans both phases.

### DIFF
--- a/docs/src/main/paradox/kubernetes-lease.md
+++ b/docs/src/main/paradox/kubernetes-lease.md
@@ -183,7 +183,13 @@ Note that version conflicts during heartbeat (meaning another node has modified 
 
 #### Acquire failure retry
 
-Transient failures during acquire (network timeout, API server unavailability) are retried within the caller's `lease-operation-timeout` budget, spanning both the initial read and the update. A retry is only scheduled while the elapsed time plus the next delay is under half the budget, leaving the rest for the retried call. Backoff starts at `lease-operation-timeout / 20`, doubles each attempt, and is capped at `lease-operation-timeout / 5`.
+When an acquire fails due to a transient error (network timeout, API server unavailability) during the initial read or the update, the lease actor retries within the caller's `lease-operation-timeout` budget rather than failing immediately.
+
+The retry time budget is calculated with safety margins:
+
+* The caller's ask times out at `lease-operation-timeout`, covering both the initial read and the update.
+* A retry is only scheduled while the elapsed time plus the next delay is under half the budget, leaving the rest for the retried call to complete.
+* Retries use exponential backoff: starting at `lease-operation-timeout / 4`, doubling each attempt (`op-timeout/4`, `op-timeout/2`, `op-timeout`), and capped at `lease-operation-timeout`. With the half-budget rule this typically allows a single retry, keeping load on a struggling API server bounded.
 
 If an update was applied server-side but the client missed the response, the retry's update is answered with a conflict naming us as owner — this is treated as a successful acquire. Genuine conflicts (a different owner) are not retried.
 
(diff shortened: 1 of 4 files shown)


lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
```diff
@@ -113,6 +113,22 @@ class LeaseActorSpec
...
+      leaseProbe.expectMsg(leaseName)
+      leaseProbe.reply(LeaseResource(None, currentVersion, System.currentTimeMillis()))
+      updateProbe.expectMsg((ownerName, currentVersion))
+      updateProbe.reply(Failure(k8sApiFailure))
+      // retry sees the lease is already owned by us
+      updateProbe.expectMsg((ownerName, currentVersion))
+      incrementVersion()
+      updateProbe.reply(Left(LeaseResource(Some(ownerName), currentVersion, System.currentTimeMillis())))
+      senderProbe.expectMsg(LeaseAcquired)
+    }
```
patriknw, line 130: perhaps add another similar test for the scenario where it was acquired by another owner, combined with the retry mechanism?

docs/src/main/paradox/kubernetes-lease.md
```diff
@@ -181,6 +181,12 @@ The retry time budget is calculated with safety margins:
 
 Note that version conflicts during heartbeat (meaning another node has modified the lease resource) are *not* retried — they indicate a genuine lease loss.
 
+#### Acquire failure retry
+
+Transient failures during acquire (network timeout, API server unavailability) are retried within the caller's `lease-operation-timeout` budget, spanning both the initial read and the update. A retry is only scheduled while the elapsed time plus the next delay is under half the budget, leaving the rest for the retried call. Backoff starts at `lease-operation-timeout / 20`, doubles each attempt, and is capped at `lease-operation-timeout / 5`.
```
patriknw: No risk that we are retrying too frequently, and thereby increasing the load on the api server, resulting in more timeouts? Not same kind of retry, but compare with other PR https://github.com/akka/akka-management/pull/1441/changes#r3051197500

patriknw (commented):
looking good

docs/src/main/paradox/kubernetes-lease.md
johanandren: Tricky balance between making room for some number of retries and not hammering on a broken service. The timeout / 2 here means 2-3 retries, going with the `timeout / 4` from #1441 means 1-2 retries, with default settings. I'll align with that.

johanandren: Aligned, and made the backoff time calculation shared.

lease-kubernetes/src/test/scala/akka/coordination/lease/kubernetes/LeaseActorSpec.scala
johanandren: Added [fc6e291](https://github.com/akka/akka-management/pull/1444/commits/fc6e291a2687119068e463b6787a7a5b28b0cc1f)

patriknw (approved):
LGTM

PR #1449, opened 2026-05-06
build: Don't email about build failures from PRs

johanandren:
We only want nightlies and fails in main sent there. PR failures are seen right away by PR author and reviewers anyway

Also, pull email on fail into a shared action to repeat a bit less stuff


patriknw:
thanks

